PR #7, opened 2024-06-02
Added a new toolset to HolmesGPT 

adithyapaib:
Added the following Toolsets 
1) Add a toolset to check what Jenkins jobs are running at the moment
2) Add a toolset to query Jaeger traces


aantn:
Hey, thank you for contributing!

There are two issues we need to resolve before merging:

1. For Jaeger, querying by trace_id generally will not be useful, as HolmesGPT has no way of knowing the trace ID upfront and wont be able to craft an appropriate API call. It may be more useful here to take one of the following approaches

(a) Query Jaeger's query dependency graph (see [here](https://www.jaegertracing.io/docs/1.57/apis/#service-dependencies-graph-internal)) so that you can ask questions like `what services are speaking to myservice`

(b) Allow querying a list of all Jaeger services and then traces for that service - [see here for example](https://github.com/Ashmita152/jaeger-datasets/blob/master/bookinfo/extract.py)

2. For Jenkins, the added tool makes sense (you can ask about faiing jenkins jobs using it), but I want to make some changes to Holmes before merging so that can give `{{ jenkins_url }}` and similar parameters special treatment as secrets that the user defines, and not regular function parameters injected by the LLM. This part is on our side.

aantn:
Hey, I'm closing this for now but please feel free to re-open after addressing the comments above!

PR #11, opened 2024-06-03
Env var-based configuration for the API server

RobertSzefler:


holmes/config.py
```diff
@@ -53,6 +56,41 @@ class ConfigFile(RobustaBaseConfig):
     custom_runbooks: List[FilePath] = []
     custom_toolsets: List[FilePath] = []
 
+    @classmethod
+    def load_from_env(cls):
+        # XXX this is currently only used in server.py. holmes.py uses
+        # load_from_file only.
+        llm_str = os.getenv("HOLMES_LLM", "OPENAI")
+        try:
```
aantn: Is it possible to write this a little more generically so that we don't need to give specific variables special treatment?

E.g. load everything with

```
return cls(
            llm=os.getenv("LLM"),
            model=os.getenv("MODEL", "gpt-4o"),
            api_key=os.getenv("API_KEY", None),
            ...
)
```

and let the runner handle the logic of putting variables the right place? E.g. writing the right value to API_KEY according to OpenAI or Azure. 

I am suggesting this because I imagine that the runner will need to be familiar with all those options anyway, so we are not going to be able to reduce the runner's complexity by adding complexity here. So lets keep it simple here and have the runner do what it will have to do anyway.

Regarding the LLM types etc I think you can let pydantic handle all the validation if taking this approach.

aantn: Sorry, by runner I mean the helm chart that wraps this or whoever sets those as env vars

RobertSzefler: done

holmes/config.py
```diff
@@ -53,6 +56,32 @@ class ConfigFile(RobustaBaseConfig):
     custom_runbooks: List[FilePath] = []
     custom_toolsets: List[FilePath] = []
 
+    @classmethod
+    def load_from_env(cls):
+        kwargs = {"llm": LLMType(os.getenv("HOLMES_LLM", "OPENAI").lower())}
+        for field_name, env_key in [
```
aantn: keep it simple! why not just convert to uppercase for all the fields and do that?

RobertSzefler: yeah, good idea

PR #12, opened 2024-06-04
Add tool call info in API responses if requested

RobertSzefler:


server.py
```diff
@@ -55,14 +57,26 @@ def investigate_issue(request: InvestigateRequest):
...
-    return {
-        "analysis": ai.investigate(
-            issue,
-            # TODO prompt should probably be configurable?
-            prompt=load_prompt("builtin://generic_investigation.jinja2"),
-            console=console,
-        ).result
+    investigation = ai.investigate(
+        issue,
+        # TODO prompt should probably be configurable?
```
aantn, line 88: Yes, I think it is worth supporting an override (optional though)
@arikalon1 do you agree?

aantn (approved):
LGTM. Left one comment not directly related to this PR. You can merge regardless of that comment.

server.py
arikalon1: yeah, I agree
We should allow an optional override
@RobertSzefler can you please add it?

RobertSzefler: added

PR #16, opened 2024-06-06
Remove unused field in LLMResult

aantn:


RobertSzefler (approved):
+1

btw there's quite a bit of unused imports in that file and other ones

aantn:
Good point, worth removing another time.

PR #17, opened 2024-06-06
added helm chart for running holmes as a deployment in the cluster

arikalon1:
pulling robusta store configuration, from the existing secret, if exists fix some small bugs


helm/holmes/templates/holmesgpt-service-account.yaml
```diff
@@ -0,0 +1,181 @@
...
+      - configmaps
+      - daemonsets
+      - deployments
+      - events
+      - namespaces
+      - persistentvolumes
+      - persistentvolumeclaims
+      - pods
+      - pods/status
+      - pods/exec
```
aantn, line 23: I would probably add a flag and make this optional (it's a higher permission than the rest and we can do without it for the default toolsets anyway)

holmes/core/tools.py
```diff
@@ -157,6 +157,7 @@ def check_prerequisites(self):
             except subprocess.CalledProcessError as e:
                 self._enabled = False
                 self._disabled_reason = f"prereq check failed w/ errorcode {e.returncode}"
+                logging.exception(f"Toolset {self.name} : Failed to run prereq command {prereq}")
```
aantn, line 160: Could you add a log for this elsewhere? We're deliberately not logging it here to keep the CLI output clean.

holmes/core/supabase_dal.py
```diff
@@ -1,52 +1,108 @@
...
+            config = RobustaConfig(**yaml_content)
+            for conf in config.sinks_config:
+                if "robusta_sink" in conf.keys():
+                    token = conf["robusta_sink"].get("token")
+                    return RobustaToken(**json.loads(base64.b64decode(token)))
+
+        return None
+
+    def __init_config(self) -> bool:
+        robusta_token = self.__load_robusta_config()
```
aantn, line 65: Could you document the behaviour somewhere - either in the helm values or in this file - that we load from robusta token on disk first if possible, and if not fallback to environment variables if given

helm/holmes/values.yaml
```diff
@@ -0,0 +1,35 @@
+
+# Optional openai key and endpoint to create openai queries.
+openaiKey: ""
+openaiEndpoint: ""
+openaiType: ""
```
aantn, line 5: What does openaiType refer to? Same question about version?

helm/holmes/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+openaiVersion: ""
+openaiDeployment: ""
+aiModel: ""
+
+certificate: "" # base64 encoded
```
aantn, line 10: What is this used for?

helm/holmes/templates/holmes.yaml
```diff
@@ -0,0 +1,102 @@
...
+                key: openaiKey
+          - name: OPENAI_ENDPOINT
+            value: {{ .Values.openaiEndpoint }}
+          - name: OPENAI_TYPE
+            value: {{ .Values.openaiType }}
+          - name: OPENAI_VERSION
+            value: {{ .Values.openaiVersion }}
+          - name: OPENAI_DEPLOYMENT
+            value: {{ .Values.openaiDeployment }}
+          - name: AI_MODEL
```
aantn, line 55: I think the names you're using here don't all match up to the names read in Config.load_from_env. 
E.g. AI_MODEL should just be MODEL

helm/holmes/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: holmes
+description: HolmesGPT Helm chart for Kubernetes
+type: application
+
+# these are set to the right value by .github/workflows/release.yaml
```
aantn, line 6: We should do that!

aantn (approved):
Looks good, left some comments.

holmes/core/tools.py
arikalon1: fixed

holmes/core/supabase_dal.py
arikalon1: I'll put a comment here

helm/holmes/values.yaml
arikalon1: for self hosted environments, some users use self signed certificates.
They can add it here, to trust their self signed certs

helm/holmes/Chart.yaml
arikalon1: yeah, I know
We will do it soon

helm/holmes/values.yaml
arikalon1: removing for now, leaving only the openaikey

helm/holmes/templates/holmes.yaml
arikalon1: removing

PR #34, opened 2024-06-10
add opsgenie and pagerduty sources

aantn:
Also:

1. rename cli flags for updating sources to consistently be named `--update`
2. clean up code a little


holmes/plugins/sources/pagerduty/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+            raise
+
+# Run with:
+# poetry run python3 -m holmes.plugins.sources.pagerduty <api-key> <user-email>
+if __name__ == "__main__":
+    import sys
+    pd_source = PagerDutySource(api_key=sys.argv[1], user_email=sys.argv[2])
+    issues = pd_source.fetch_issues()
+    for issue in issues:
+        pd_source.write_back_result(issue.id, LLMResult(result="This is a test"))
```
Avi-Robusta, line 94: was "This is a test" left here on purpose?

aantn: Yes, it's just for the test where we want to test the PagerDutySource but not the LLM.

Avi-Robusta (approved):

PR #35, opened 2024-06-10
Updated Readme

pavangudiwada:


aantn (approved):
LGTM. While we are removing README redundancies, can you make one more improvement and merge the "Customising config" section into "Reading setting from a config file" (under the Advanced Usage)

Also, please make sure you have the CLA signed so we can merge.

PR #36, opened 2024-06-10
Increase compatibility of linux binaries

aantn:
@Avi-Robusta note: we should do the same for KRR too.

### DIFF
--- a/.github/workflows/build-on-release.yaml
+++ b/.github/workflows/build-on-release.yaml
@@ -136,7 +136,7 @@ jobs:
           name: holmes-ubuntu-20.04-${{ github.ref_name }}
       - name: Calculate hash
         id: calc-hash
-        run: echo "::set-output name=LINUX_BUILD_HASH::$(sha256sum holmes-$-ubuntu-20.04-${{ github.ref_name }}.zip | awk '{print $1}')"
+        run: echo "::set-output name=LINUX_BUILD_HASH::$(sha256sum holmes-ubuntu-20.04-${{ github.ref_name }}.zip | awk '{print $1}')"
 
   # TODO: update homebrew formula
   update-formula:
--- a/.github/workflows/build-on-release.yaml
+++ b/.github/workflows/build-on-release.yaml
@@ -8,7 +8,10 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-on-release.yaml
```diff
@@ -152,7 +155,7 @@ jobs:
           TAG_NAME=${{ github.ref_name }}
           awk 'NR==6{$0="        url \"https://github.com/robusta-dev/holmesgpt/releases/download/'"$TAG_NAME"'/holmes-macos-latest-'"$TAG_NAME"'.zip\""}1' ./Formula/holmesgpt.rb > temp && mv temp ./Formula/holmesgpt.rb
           awk 'NR==7{$0="        sha256 \"'$MAC_BUILD_HASH'\""}1' ./Formula/holmesgpt.rb > temp && mv temp ./Formula/holmesgpt.rb
-          awk 'NR==9{$0="        url \"https://github.com/robusta-dev/holmesgpt/releases/download/'"$TAG_NAME"'/holmes-ubuntu-latest-'"$TAG_NAME"'.zip\""}1' ./Formula/holmesgpt.rb > temp && mv temp ./Formula/holmesgpt.rb
+          awk 'NR==9{$0="        url \"https://github.com/robusta-dev/holmesgpt/releases/download/'"$TAG_NAME"'/holmes-ubuntu-20.04-'"$TAG_NAME"'.zip\""}1' ./Formula/holmesgpt.rb > temp && mv temp ./Formula/holmesgpt.rb
           awk 'NR==10{$0="        sha256 \"'$LINUX_BUILD_HASH'\""}1' ./Formula/holmesgpt.rb > temp && mv temp ./Formula/holmesgpt.rb
```
Avi-Robusta, line 159: I dont think this is related but it looks like there are issues with calculating the hash for ubuntu, it looks like its from before this too, might want to verify brew install works on ubuntu
Screen Shot 2024-06-10 at 16 43 06

Avi-Robusta (approved):

.github/workflows/build-on-release.yaml
aantn: Very nice catch. Fixed it.

PR #37, opened 2024-06-10
added option to dump complete raw data to json

Avi-Robusta:
`poetry run python3 holmes.py ask "What is the issue with payment-processing-worker" -raw-json-file="out.json"`

### DIFF
--- a/holmes.py
+++ b/holmes.py
@@ -12,7 +12,7 @@ from rich.console import Console
 from rich.logging import RichHandler
 from rich.markdown import Markdown
 from rich.rule import Rule
-
+from holmes.utils.file_utils import write_json_file
 from holmes.config import Config, LLMType
 from holmes.plugins.destinations import DestinationType
 from holmes.plugins.prompts import load_prompt
@@ -100,9 +100,9 @@ opt_slack_channel: Optional[str] = typer.Option(
     "--slack-channel",
     help="Slack channel if --destination=slack (experimental). E.g. #devops",
 )
-opt_dump_raw_output: Optional[str] = typer.Option(
+opt_json_output_file: Optional[str] = typer.Option(
     None,
-    "-raw-json-file",
+    "--json-output-file",
     help="Save the complete output in json format in to a file",
 )
 
@@ -134,7 +134,7 @@ def ask(
         "--show-tool-output",
         help="Advanced. Show the output of each tool that was called",
     ),
-    dump_raw_output: Optional[str] = opt_dump_raw_output
+    json_output_file: Optional[str] = opt_json_output_file
 ):
     """
     Ask any question and answer using available tools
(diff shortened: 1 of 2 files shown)


holmes.py
```diff
@@ -100,6 +100,11 @@ def init_logging(verbose = False):
     "--slack-channel",
     help="Slack channel if --destination=slack (experimental). E.g. #devops",
 )
+opt_dump_raw_output: Optional[str] = typer.Option(
+    None,
+    "-raw-json-file",
```
aantn: Can we name it --json-output-file please.
If you name the cli argument `json_output_file` then I don't think you'll have to explicitly specify the cli flag.

holmes.py
```diff
@@ -148,6 +154,9 @@ def ask(
     console.print("[bold yellow]User:[/bold yellow] " + prompt)
     response = ai.call(system_prompt, prompt)
     text_result = Markdown(response.result)
+    if dump_raw_output:
+        with open(dump_raw_output , 'w' , encoding='utf-8') as f:
+            json.dump(json.loads(response.model_dump_json()), f, ensure_ascii=False, indent=4)
```
aantn: Do we need the 3-way roundtrip? Is there a reason not to write `response.model_dump_json()` to file?

holmes.py
```diff
@@ -234,11 +244,13 @@ def alertmanager(
         console.print(
             f"[bold yellow]Analyzing all {len(issues)} issues. (Use --alertname to filter.)[/bold yellow] [red]Press Ctrl+C to stop.[/red]"
         )
+    results = []
     for i, issue in enumerate(issues):
         console.print(
             f"[bold yellow]Analyzing issue {i+1}/{len(issues)}: {issue.name}...[/bold yellow]"
         )
         result = ai.investigate(issue, system_prompt, console)
+        results.append({"issue": json.loads(issue.json()), "result": json.loads(result.json())})
```
aantn: Here too, I don't think the roundtrip is necessary. What are you trying to achive with it?

holmes.py
```diff
@@ -250,6 +262,9 @@ def alertmanager(
         elif destination == DestinationType.SLACK:
             slack.send_issue(issue, result)
 
+    if dump_raw_output:
```
aantn: Can you create a shared function to dump the json that both commands use? (And please add it to the other commands as well for consistency.)

I suspect there is an idiomatic way to dump pydantic models to file that will work and be re-usable in all the calls.

aantn (approved):
Left some comments.

holmes.py
Avi-Robusta: model_dump_json makes it a json string, used model_dump

holmes.py
Avi-Robusta: done

aantn (approved):

PR #44, opened 2024-06-16
Fix python dependencies and test them on each commit/PR

aantn:
# Background
We accidentally shipped broken binaries for the last version which were missing a python dependency. To fix that, this PR:

1. Removes requirements.txt and always relies on poetry so that the two cannot be out of sync
2. Build and tests binaries on every commit and PR

# Testing
I pushed a "bad commit" to this branch which imported a package not in pyproject.toml. I verified that the tests failed. Then I verified that once the commit was reverted, this branch builds properly and all tests pass.


Sheeproid (approved):
Nice work. LGTM

PR #45, opened 2024-06-16
added filter on alert by label

Avi-Robusta:
Example run
```
kubectl apply -f https://raw.githubusercontent.com/robusta-dev/kubernetes-demos/main/pending_pods/pending_pod_node_selector.yaml

# wait till alert fires...

poetry run python3 /Users/avirobusta/git/holmesgpt/holmes.py investigate alertmanager --alertmanager-url=http://localhost:9093/ --alert-label-key=pod --alert-label-value=user-profile-import

Configuration file exists at /Users/avirobusta/Library/Preferences/pypoetry, reusing this directory.

Consider moving TOML configuration files to /Users/avirobusta/Library/Application Support/pypoetry, as support for the legacy directory will be removed in an upcoming release.
Analyzing all 1 issues. (Use --alertmanager-alertname to filter.) Press Ctrl+C to stop.
Analyzing issue 1/1: KubePodNotReady...
No runbooks found for this issue. Using default behaviour. (Add runbooks to guide the investigation.)
Running `kubectl describe Pod user-profile-import -n default`                                                                                                                                                                           tools.py:89
────────────────────────────────────────────────────────
AI: Node Affinity Mismatch                                                                                                                                                                                                                             

Resource: Pod user-profile-import                                                                                                                                                                                                                  

Details: Pod user-profile-import is pending because 0/3 nodes match its node affinity/selector criteria.                                                                                                                                           
────────────────────────────────────────────────────────

```

### DIFF
--- a/holmes.py
+++ b/holmes.py
@@ -217,6 +217,7 @@ def alertmanager(
         alertmanager_username=alertmanager_username,
         alertmanager_password=alertmanager_password,
         alertmanager_alertname=alertmanager_alertname,
+        alertmanager_label=alertmanager_label,
         slack_token=slack_token,
         slack_channel=slack_channel,
         custom_toolsets=custom_toolsets,
@@ -233,16 +234,6 @@ def alertmanager(
 
     try:
         issues = source.fetch_issues()
-        if alertmanager_label:
-            label_parts = alertmanager_label.split('=')
-            if len(label_parts) != 2:
-                logging.error(f"The label {alertmanager_label} is of the wrong format use '--alertmanager-label key=value'")
-                return
-            alert_label_key, alert_label_value = label_parts
-            issues = [issue for issue in issues if issue.raw.get("labels", {}).get(alert_label_key, None) == alert_label_value]
-            if not issues:
-                logging.error(f"No valid alerts with the label {alert_label_key} and value {alert_label_value}")
-                return
     except Exception as e:
         logging.error(f"Failed to fetch issues from alertmanager", exc_info=e)
         return
--- a/holmes.py
+++ b/holmes.py
@@ -198,11 +198,8 @@ def alertmanager(
     system_prompt: Optional[str] = typer.Option(
         "builtin://generic_investigation.jinja2", help=system_prompt_help
     ),
-    alert_label_key: Optional[str] = typer.Option(
-        None, help="A key for filtering alerts with specific label key"
-    ),
-    alert_label_value: Optional[str] = typer.Option(
-        None, help="A value for filtering alerts with specific label value"
+    alertmanager_label: Optional[str] = typer.Option(
+        None, help="For filtering alerts with specific labelm must be of format key=value"
     ),
 ):
     """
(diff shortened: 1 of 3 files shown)


holmes.py
```diff
@@ -230,6 +233,16 @@ def alertmanager(
 
     try:
         issues = source.fetch_issues()
+        if alertmanager_label:
+            label_parts = alertmanager_label.split('=')
+            if len(label_parts) != 2:
+                logging.error(f"The label {alertmanager_label} is of the wrong format use '--alertmanager-label key=value'")
+                return
+            alert_label_key, alert_label_value = label_parts
+            issues = [issue for issue in issues if issue.raw and issue.raw.get("labels", {}).get(alert_label_key, None) == alert_label_value]
```
aantn: Is it possible for `issue.raw` to not be defined? If it isn't possible, don't check for it.

aantn (approved):
Left one comment. LGTM.

holmes.py
Avi-Robusta: it could in theory be None correct? and then None.get might throw an exception

aantn: It can't be None. (See the AlertManagerSource code.)

holmes.py
```diff
@@ -198,6 +198,9 @@ def alertmanager(
     system_prompt: Optional[str] = typer.Option(
         "builtin://generic_investigation.jinja2", help=system_prompt_help
     ),
+    alertmanager_label: Optional[str] = typer.Option(
+        None, help="For filtering alerts with specific labelm must be of format key=value"
```
aantn: Typo in labelm

holmes/plugins/sources/prometheus/plugin.py
```diff
@@ -72,6 +74,22 @@ def fetch_issues(self) -> List[Issue]:
             )
             for alert in alerts
         ]
+        return self.label_filter_issues(issues)
```
aantn: Is there a reason we filter on alert name above but on alert labels below?

Lets be consistent unless there is a very good reason not to be.

aantn (approved):
Final comments!

holmes/plugins/sources/prometheus/plugin.py
```diff
@@ -72,6 +74,22 @@ def fetch_issues(self) -> List[Issue]:
...
+            return issues
+
+        label_parts = self.label.split('=')
+        if len(label_parts) != 2:
+            raise Exception(f"The label {self.label} is of the wrong format use '--alertmanager-label key=value'")
+
+        alert_label_key, alert_label_value = label_parts
+        filtered_issues = [issue for issue in issues if issue.raw.get("labels", {}).get(alert_label_key, None) == alert_label_value]
+        if not filtered_issues:
+            raise Exception(f"No valid alerts with the label {alert_label_key} and value {alert_label_value}")
```
aantn: Please make the behaviour consistent with alertname filtering. (If there are no matching alerts we return an empty list instead of throwing an exception.)

holmes/plugins/sources/prometheus/plugin.py
```diff
@@ -73,6 +79,21 @@ def fetch_issues(self) -> List[Issue]:
...
+        if not self.label:
+            return issues
+
+        label_parts = self.label.split('=')
+        if len(label_parts) != 2:
+            raise Exception(f"The label {self.label} is of the wrong format use '--alertmanager-label key=value'")
+
+        alert_label_key, alert_label_value = label_parts
+        filtered_issues = [issue for issue in issues if issue.labels.get(alert_label_key, None) == alert_label_value]
+        if not filtered_issues:
```
aantn: unecessary

aantn: without this check you still return []

PR #55, opened 2024-06-20
Update README.md

xxradar:


aantn:
Already fixed in #54!

You were first, but I didn't see your PR until after I pushed a fix myself. Thank you!

PR #56, opened 2024-06-22
Add `holmes generate alertmanager-tests` command

aantn:
and add flags to `holmes investigate alertmanager` to use that data instead of connecting to a live alertmanager


holmes.py
```diff
@@ -267,6 +280,37 @@ def alertmanager(
     if json_output_file:
         write_json_file(json_output_file, results)
 
+
+@generate_app.command("alertmanager-tests")
+def generate_alertmanager_tests(
+    alertmanager_url: Optional[str] = typer.Option(None, help="AlertManager url"),
```
Avi-Robusta: correct me if im wrong but it looks like for this command alertmanager_url is not optional

aantn: It's still optional here, because it could be defined in the config file.

Avi-Robusta:
for tests
`poetry run python3 ./holmes.py investigate alertmanager --alertmanager-file ...
`when no url is specified you will get a pydantic exception 
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Issue                                                                                                                                                                                                     
source_instance_id                                 
```    
since it uses the url and its none in the alertmanager source
I would change the line in `AlertManagerSource` for `source_instance_id` to 

```
        return [
            Issue(
                id=alert.unique_id,
                name=alert.name,
                source_type="prometheus",
                source_instance_id= self.filepath if self.filepath else self.url,
```

Avi-Robusta (commented):
LGTM,
2 changes requested

holmes.py
Avi-Robusta: than specifically in this function I would check before the line
`    source = config.create_alertmanager_source()
`to make the error output shorter for users
This is what I see:
Screen Shot 2024-06-23 at 11 34 03

Avi-Robusta (approved):

holmes.py
Avi-Robusta: also a different issue here, want to make sure you dont miss it
https://github.com/robusta-dev/holmesgpt/pull/56#issuecomment-2184900236

aantn: I think it's ok here. The last line of output clearly says `--alertmanager-url must be specified`.
If someone misses it or is confused I'll rework it later.

aantn:
> for tests `poetry run python3 ./holmes.py investigate alertmanager --alertmanager-file ... `when no url is specified you will get a pydantic exception
> 
> ```
> pydantic_core._pydantic_core.ValidationError: 1 validation error for Issue                                                                                                                                                                                                     
> source_instance_id                                 
> ```
> 
> since it uses the url and its none in the alertmanager source I would change the line in `AlertManagerSource` for `source_instance_id` to
> 
> ```
>         return [
>             Issue(
>                 id=alert.unique_id,
>                 name=alert.name,
>                 source_type="prometheus",
>                 source_instance_id= self.filepath if self.filepath else self.url,
> ```

Good catch, will fix.

Avi-Robusta (approved):

PR #58, opened 2024-06-22
Fix version command when it runs outside of a git repo

aantn:
It gave unwanted output due to STDERR not being captured


arikalon1 (approved):
LGTM

PR #59, opened 2024-06-22
Improve prompt to fix failing tests

aantn:
# Motivation
Without this change, the ask command occasionally gives "lazy output" like stating that a pod is pending but not investigating why.

# Testing
I ran our benchmarks on this branch to verify that this change fixes the problem and does not cause regressions for other test scenarios. I tested our scenarios with --repeat=5. All tests passed according to the LLM evaluated-results. I then reviewed all the test output manually to verify that there were no undesired consequences of this change in LLM output.

The results look better to me as a result of this change, and I can confirm that is reduces LLM-laziness. (The LLM was not lazy in a single test scenario after this change.)


arikalon1 (approved):
LGTM

PR #61, opened 2024-06-23
Read default config file from ~/.holmes/config.yaml not current directory

aantn:
This makes holmes behave like other command line tools (e.g. kubectl) and more like users expect. This came up in a recent user feedback call.


aantn:
@pavangudiwada FYI

arikalon1 (approved):
LGTM

PR #62, opened 2024-06-23
Update config.py

aantn:


aantn:
Not relevant, I had a bug in the code and fixed this separately.

PR #65, opened 2024-06-24
Add support for many llms via litellm

aantn:
Note: this is currently backwards incompatible as it removes old flags for azure. We need to decide if we are adding these back for backwards compatibility, or if we will stick with the litellm way of doing things instead with environment variables only.


arikalon1 (approved):
LGTM

Lets merge it, and then do the required updates on master, since this branch is old

PR #66, opened 2024-06-24
update holmes chart

arikalon1:
fix slack upload - change to new api

### DIFF
--- a/server.py
+++ b/server.py
@@ -12,7 +12,8 @@ import logging
 import uvicorn
 import colorlog
 
-from typing import List, Union
+from holmes.core.tool_calling_llm import ToolCallResult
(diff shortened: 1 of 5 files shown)


server.py
```diff
@@ -35,8 +36,8 @@ class InvestigateRequest(BaseModel):
     title: str
     description: str
     subject: dict
-    context: List[InvestigateContext]
```
Avi-Robusta, line 38: if we dont use InvestigateContext anymore we should get rid of the class from above

arikalon1: nice catch, fixing

Avi-Robusta (approved):
LGTM,
One comment

PR #75, opened 2024-06-28
Minor improvements to allow testing `investigate` commands more easily

aantn:
Right now this PR only touches the jira command, but similar PRs will follow for other commands later.


arikalon1 (approved):
LGTM

PR #76, opened 2024-06-30
Fix holmes version in more scenarios

aantn:
Now holmes version will work correctly for git archives. This is important because it lets us record correct versions of holmes when testing holmes with commands like

```
pipx run --spec https://github.com/robusta-dev/holmesgpt/archive/refs/heads/more-fixes-to-version-command.zip holmes version
```


arikalon1 (approved):
LGTM

PR #79, opened 2024-07-02
make it easier to test alertmanager from our test framework

aantn:


holmes/main.py
```diff
@@ -262,6 +268,10 @@ def alertmanager(
     except Exception as e:
         logging.error(f"Failed to fetch issues from alertmanager", exc_info=e)
         return
+    
+    if alertmanager_limit is not None:
+        console.print(f"[bold yellow]Limiting to {alertmanager_limit}/{len(issues)} issues.[/bold yellow]")
+        issues = issues[:alertmanager_limit]
```
Avi-Robusta, line 274: is this an issue if the limit is bigger than the length of the array?

holmes/plugins/sources/prometheus/plugin.py
```diff
@@ -57,6 +59,10 @@ def __fetch_issues_from_api(self) -> List[PrometheusAlert]:
             "silenced": "false",
             "inhibited": "false",
         }
+        if self.label_filter:
+            params["filter"] = self.label_filter
```
Avi-Robusta, line 63: how does this work with a list? do you have an example? is it 'and' or 'or' filtering?

holmes/main.py
```diff
@@ -194,8 +194,9 @@ def alertmanager(
         None,
         help="Investigate all alerts with this name (can be regex that matches multiple alerts). If not given, defaults to all firing alerts",
     ),
-    alertmanager_label: Optional[str] = typer.Option(
-        None, help="For filtering alerts with a specific label. Must be of format key=value"
+    alertmanager_label: Optional[List[str]] = typer.Option(
+        [],
+        help="For filtering alerts with a specific label. Must be of format key=value. If --alertmanager-label is passed multiple times, alerts must match ALL labels"
```
Avi-Robusta, line 199: not sure what this means
`If --alertmanager-label is passed multiple times`

Avi-Robusta (approved):
looks good to me, a few things I wasnt clear about,
one potential issue i left in a comment you should check

holmes/main.py
aantn: No, it's fine. 

```python
>>> [0,1,2][:10]
[0, 1, 2]
```

holmes/plugins/sources/prometheus/plugin.py
aantn: It's AND. You filter on both, which is what we want here.

holmes/main.py
aantn: You can run `holmes investigate alertmanager --alertmanager-label x=y --alertmanager-label a=b`
Any ideas on how to phrase better?

aantn:
> looks good to me, a few things I wasnt clear about, one potential issue i left in a comment you should check

Thank you!

PR #80, opened 2024-07-03
Add pipx installation instructions

aantn:
I have been using pipx myself for recent installs and have had an excellent experience. It works very well.


arikalon1 (approved):
LGTM

PR #87, opened 2024-07-18
Added Prometheus Operator RBAC for holmes to service accout template

itisallgood:
Added Prometheus Operator RBAC for holmes to service accout template 

https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/rbac.md#prometheus-operator-rbac


RoiGlinik (approved):
Looks good

PR #88, opened 2024-07-22
[MAIN-1860] Added post processing option for chatty ai's

Avi-Robusta:
Example command:
```
poetry run python3 holmes.py ask "Investigate the issues in detail with the server-2 pods?" --model=bedrock/anthropic.claude-3-5-sonnet-20240620-v1:0
```

Example result
```
AI: Issues with server-2 pods:                                                                                                                                                                                                                                                    

 1 Multiple pending pods in default and stg namespaces.                                                                                                                                                                                                                       
 2 Pod server-2-59c95cbdcb-ntrh4 in default namespace is in Pending state.                                                                                                                                                                                                    
 3 Scheduling error: "0/4 nodes are available: persistentvolumeclaim "non-existant-claim" not found."                                                                                                                                                                         
 4 Pod attempting to use non-existent PersistentVolumeClaim "non-existant-claim".                                                                                                                                                                                             
 5 Container trying to mount volume at "/usr/share/nginx/html" using the missing PVC.                                                                                                                                                                                         
 6 Pod using "nginx:latest" image.   
```

Example  of chatty pre post processing result
```
Based on the investigation, there are several issues with the server-2 pods. Let's break down the problems:                                                                                                                                                                   
                                                                                                                                                                                                                                                                              
1. Multiple pending pods:                                                                                                                                                                                                                                                     
   There are three pending server-2 pods across two namespaces (default and stg). We'll focus on the one in the default namespace for which we got detailed information.                                                                                                      
                                                                                                                                                                                                                                                                              
2. Pod Status:                                                                                                                                                                                                                                                                
   The pod server-2-59c95cbdcb-ntrh4 in the default namespace is in a Pending state, which means it's not running.                                                                                                                                                            
                                                                                                                                                                                                                                                                              
3. Scheduling Issue:                                                                                                                                                                                                                                                          
   The main problem preventing the pod from running is a scheduling issue. The error message states: "0/4 nodes are available: persistentvolumeclaim "non-existant-claim" not found."                                                                                         
                                                                                                                                                                                                                                                                              
4. Volume Configuration:                                                                                                                                                                                                                                                      
   The pod is trying to use a PersistentVolumeClaim named "non-existant-claim", which doesn't exist in the cluster. This is preventing the pod from being scheduled.                                                                                                          
                                                                                                                                                                                                                                                                              
5. Volume Mount:                                                                                                                                                                                                                                                              
   The container is trying to mount a volume at "/usr/share/nginx/html" using the non-existent PVC.                                                                                                                                                                           
                                                                                                                                                                                                                                                                              
6. Container Image:                                                                                                                                                                                                                                                           
   The pod is using the "nginx:latest" image.                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                              
To resolve these issues:                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                              
1. Create the missing PersistentVolumeClaim:                                                                                                                                                                                                                                  
   You need to create a PersistentVolumeClaim named "non-existant-claim" in the default namespace. The name suggests this might be a misconfiguration, so you may want to review if this is the intended name or if it should be changed.                                     
                                                                                                                                                                                                                                                                              
2. Check PersistentVolume availability:                                                                                                                                                                                                                                       
   Ensure that there's a PersistentVolume available that can satisfy the requirements of the PersistentVolumeClaim.                                                                                                                                                           
                                                                                                                                                                                                                                                                              
3. Review the Deployment or StatefulSet configuration:                                                                                                                                                                                                                        
   The pod is controlled by a ReplicaSet, which is likely managed by a Deployment. You should review the Deployment configuration to ensure it's correctly specifying the volume claims and mounts.                                                                           
                                                                                                                                                                                                                                                                              
4. Check other namespaces:                                                                                                                                                                                                                                                    
   Since there are similar pods in the stg namespace, you might want to review the configuration there as well to ensure consistency across environments.                                                                                                                     
                                                                                                                                                                                                                                                                              
5. Consider using a specific image tag:                                                                                                                                                                                                                                       
   Instead of using "nginx:latest", it's generally a good practice to use a specific version tag for better control and predictability.                                                                                                                                       
                                                                                                                                                                                                                                                                              
To get more information and potentially resolve the issue, you might want to:                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                              
1. Check the details of the Deployment controlling these pods.                                                                                                                                                                                                                
2. Verify if the PersistentVolumeClaim exists and its status.                                                                                                                                                                                                                 
3. Review the available PersistentVolumes in the cluster.                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                              
Would you like me to investigate any of these areas further?                      
```

Notes:
The builtin prompt for post processing `generic_post_processing.jinja2` still can be improved for optimum summaries.

To enable it in cluster add this to your values file:
`enablePostProcessing: true`


holmes/core/tool_calling_llm.py
```diff
@@ -120,14 +122,26 @@ def call(self, system_prompt, user_prompt) -> LLMResult:
 
             tools_to_call = getattr(response_message, "tool_calls", None)
             if not tools_to_call:
+                # claude is a chatty model so we automatically post process it
+                if post_process_prompt or "claude" in self.model:
```
arikalon1: I would just use the env var

arikalon1 (approved):
LGTM

Left one comment

holmes/core/tool_calling_llm.py
arikalon1: fixing

PR #90, opened 2024-07-23
updating python packages

Avi-Robusta:
setting url lib to 1.26.19 because of
CVE-2024-37891
boto3 and botocore dont support newer versions of urllib3


arikalon1 (approved):
LGTM

PR #91, opened 2024-07-23
Init lc.

eggitayi:


aantn:
Hey, there are a bunch of different changes in here. Can you ellaborate a little more on the motivation and we'll figure out a way to get this merged or supported in a different manner? 

We'd ideally like to avoid adding much langchain specific code, but we have ideas how to accomplish the same thing with LiteLLM and other approaches. (E.g. turn holmesgpt into a library that you can load from your code.)

eggitayi:
I just want to use a free llm and langchain react agent to re-complish this project work.
Juset a test, no need to merge.

aantn:
Got it, thanks. If you want to get something merged in the future please lmk.

PR #92, opened 2024-07-23
update the helm chart to suppot working with litellm env vars

arikalon1:


README.md
```diff
@@ -252,7 +252,7 @@ To work with Azure AI, you need an [Azure OpenAI resource](https://learn.microso
 
 * AZURE_API_VERSION - e.g. 2024-02-15-preview
 * AZURE_API_BASE - e.g. https://my-org.openai.azure.com/
-* AZURE_OPENAI_API_KEY (optional) - equivalent to the `--api-key` cli argument
+* AZURE_API_KEY (optional) - equivalent to the `--api-key` cli argument
```
Avi-Robusta: in main.py I would update the help since it still says:

```
opt_api_key: Optional[str] = typer.Option(
    None,
    help="API key to use for the LLM (if not given, uses environment variables OPENAI_API_KEY or AZURE_OPENAI_API_KEY)",
)
```

Avi-Robusta (approved):
LGTM

README.md
```diff
@@ -305,6 +305,95 @@ In particular, note that [vLLM does not yet support function calling](https://gi
 
 </details>
 
+### Install using helm - Run HolmesGPT in your cluster
+
+You can install HolmesGPT in your Kubernetes cluster, and use its rest API to investigate incidents. 
+This works best with the [Robusta-Holmes integration ↗](https://docs.robusta.dev/master/configuration/ai-analysis.html).
+
+In this mode, all the parameters should be passed to the HolmesGPT deployment, using environment variables.
```
aantn, line 231: I think this could be confusing for users. You're implying here that you are giving instructions for installing holmes with the Robusta integration, but actually users who follow the below instructions will end up with Holmes running in the cluster alone

aantn (approved):
Left a comment on the docs

README.md
arikalon1: yes, good point

README.md
arikalon1: fixed

PR #95, opened 2024-07-26
Allow reading model param from config file

aantn:
Note to whoever reviews this: we are removing the default value for the model from *main.py* but we are **not** removing the default value. The default is still defined in *config.py* as it should be. The unecessary redefining of the default in main.py caused the bug.
 
Thank you to @vfarcic for reporting.


arikalon1 (approved):
LGTM

PR #96, opened 2024-07-28
[MAIN-1849] added dockerhub build to gitaction

Avi-Robusta:


.github/workflows/build-docker-images.yaml
```diff
@@ -58,6 +58,22 @@ jobs:
         build-args: |
           BUILDKIT_INLINE_CACHE=1
 
+    - name: Login to Docker Hub
```
Sheeproid, line 61: check if you can unify this with Build and push Docker images

Sheeproid (approved):

PR #102, opened 2024-08-05
allow not echoing user request

aantn:
e.g. run `poetry run holmes ask "what is wrong with my cluster?" --no-echo` and the user's question will not be echoed back to them. This is useful for scripting scenarios.

### DIFF
--- a/holmes/main.py
+++ b/holmes/main.py
@@ -101,8 +101,8 @@ opt_verbose: Optional[bool] = typer.Option(
 )
 opt_echo_request: bool = typer.Option(
     True,
-    "--echo-request/--no-echo-request",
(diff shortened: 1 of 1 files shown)


holmes/main.py
```diff
@@ -99,6 +99,11 @@ def init_logging(verbose = False):
     "-v",
     help="Verbose output",
 )
+opt_echo_request: bool = typer.Option(
+    True,
```
Avi-Robusta, line 103: Im not sure i personally would keep this as true even though this was the default behavior until now

Avi-Robusta (approved):

holmes/main.py
aantn: merging for now and open to changing later! thanks.

PR #106, opened 2024-08-06
Main 1907 workload health endpoint

RoiGlinik:


holmes/core/supabase_dal.py
```diff
@@ -113,9 +114,81 @@ def get_issue_data(self, issue_id: str) -> Optional[Dict]:
...
             .table(EVIDENCE_TABLE)
-            .select(f"*")
+            .select("*")
             .filter("issue_id", "eq", issue_id)
             .execute()
         )
-        issue_data["evidence"] = evidence.data
-        return issue_data
\ No newline at end of file
+        enrichment_blacklist = {"text_file", "alert_labels", "ai_analysis", "holmes"}
```
arikalon1: Why are you also removing the alert labels? and text files?

holmes/core/supabase_dal.py
```diff
@@ -113,9 +114,81 @@ def get_issue_data(self, issue_id: str) -> Optional[Dict]:
...
+
+            res = (
+                self.client
+                .table(EVIDENCE_TABLE)
+                .select("data, enrichment_type")
+                .in_("issue_id", unique_issues)
+                .execute()
+            )
+
+            enrichment_blacklist = {"text_file", "alert_labels", "ai_analysis", "holmes"}
```
arikalon1: same

arikalon1 (approved):
Looks good, left a minor comment

holmes/core/supabase_dal.py
RoiGlinik: text files are zipped and require some extra work. also the logs are taken in the investigation so it seems redundant.
alert labels just not sure how useful they are

PR #120, opened 2024-08-28
holmes deployment add tolerations

julianobarbosa:
Added tolerations to deployments

### DIFF
--- a/helm/holmes/templates/holmes.yaml
+++ b/helm/holmes/templates/holmes.yaml
@@ -49,9 +49,10 @@ spec:
           preStop:
             exec:
               command: ["bash", "-c", "kill -SIGINT 1"]
-       {{- if .Values.tolerations }}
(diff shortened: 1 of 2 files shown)


helm/holmes/templates/holmes.yaml
```diff
@@ -49,6 +49,10 @@ spec:
           preStop:
             exec:
               command: ["bash", "-c", "kill -SIGINT 1"]
+      {{- if .Values.tolerations }}
```
arikalon1, line 52: I think you added the `tolerations` inside the container, but they should be one level up, in the `spec`
The `volumeMounts` are part of the container

arikalon1 (commented):
thanks for the PR @julianobarbosa

Left one comment
Please also sign the CLA, so I'll be able to merge the PR

PR #131, opened 2024-09-10
Handle large inputs and LLM context windows better for ask

itisallgood:


holmes/core/tool_calling_llm.py
```diff
@@ -243,7 +251,21 @@ def _post_processing_call(self, prompt, investigation, user_prompt: Optional[str
             logging.exception("Failed to run post processing", exc_info=True)
             return investigation
 
-           
+    def truncate_messages_to_fit_context(self, messages: list, max_context_size: int, maximum_output_token: int) -> list:
+        messages_except_tools = [message for message in messages if message["role"] != "tool"]
+        message_size_without_tools = self.count_tokens_for_message(messages_except_tools)
+
+        tool_call_messages = [message for message in messages if message["role"] == "tool"]
+        if not tool_call_messages or message_size_without_tools >= max_context_size:
```
arikalon1: I would add parentheses to make the logic clearer. We have `not`, `or`, and `>=`. Parentheses will make the evaluation precedence clear.

Can you explain the logic?
If we have no tool calls we return the `messages` as is? 
Or, if we do have tool calls, but `messages` is bigger than max context size, we also return it as is?

Wouldn't we get an error in both cases, because we exceeded the allowed context size?

arikalon1 (commented):
nice work, left one comment

holmes/core/tool_calling_llm.py
itisallgood: I see what you mean. I think I should change name of the function from `truncate_messages_to_fit_context` to `truncate_tool_messages` or similar, and remove those checks. 

We use this function only to truncate tool messages for `ask` holmes because we actually from other messages have only 
```python
messages = [
                {
                    "role": "system",
                    "content": system_prompt,
                },
                {
                    "role": "user",
                    "content": user_prompt,
                },
            ]
``` 
everything else is tools. Since these are the only messages and they're essential, we don't truncate them. If their `system_prompt +  user_prompt` size exceeds the context window, then we should just let it raise an error as clearly something is wrong in this case

PR #139, opened 2024-09-22
Reuse prompts

aantn:
Use jinja2 template includes so we don't repeat prompts everywhere.

Tests

- [x] Test holmes ask
- [ ] Test holmes in server mode and integrated with runner - test all 3 prompts used that way (investigate alert, ask follow up question, and check for microservice issue)
- [ ] Test a holmes investigate command


arikalon1 (approved):
LGTM

PR #150, opened 2024-10-06
NOT-READY - oci tool calling working

Avi-Robusta:


README.md
```diff
@@ -341,6 +341,31 @@ For OpenAI, only the ``model`` and ``api-key`` should be provided
 
 </details>
 
+<details>
+<summary>OCI (Oracle Cloud Infrastructure) in Kubernetes</summary>
```
Avi-Robusta, line 345: update in cluster docs for all params from ~/.oci/config file

holmes/core/aux_llms/oci_llms.py
```diff
@@ -0,0 +1,194 @@
...
+class Choice(BaseModel):
+    message: Message
+
+class LiteLLMResponse(BaseModel):
+    choices: List[Choice]
+
+class ModelSupportException(Exception):
+    pass
+
+class OCILLM:
```
Avi-Robusta, line 22: maybe make some sort of interface base class so we have something generic for every time we need to expand litellm?
Like all the functions we need accessable are:
supports_llm function
the maximum size functions 
the chat/completion function

RoiGlinik:
@Avi-Robusta  is this relevant or can be closed ?

PR #162, opened 2024-10-21
MAIN-2345 Support new conversation functionality

itisallgood:


holmes/core/models.py
```diff
@@ -67,6 +67,7 @@ class HolmesConversationIssueContext(BaseModel):
 
 class ConversationType(str, Enum):
     ISSUE = "issue"
+    CHAT = "chat"
```
arikalon1: I think this is not in use

holmes/core/models.py
```diff
@@ -79,6 +80,18 @@ class ConversationRequest(BaseModel):
     include_tool_call_results: bool = False
 
 
+class HolmesConversationIssueContext(BaseModel):
```
arikalon1, line 85: I think it's not in use

holmes/core/tool_calling_llm.py
```diff
@@ -126,9 +123,12 @@ def get_maximum_output_token(self) -> int:
         except Exception as e:
             logging.warning(f"Couldn't find model's name {model_name} in litellm's model list, fallback to 4096 tokens for max_output_tokens")
             return 4096
-
-    def call(self, system_prompt, user_prompt, post_process_prompt: Optional[str] = None, response_format: dict = None) -> LLMResult:
-        messages = [
+    
+    def call(self, system_prompt: Optional[str] = None, user_prompt: Optional[str] = None, post_process_prompt: Optional[str] = None, response_format: dict = None,
```
arikalon1: I think this is confusing 
system_prompt, user_prompt, and messages cannot be all optional

I think it would be clearer if you separate it to 2 functions, which hold the common code in a third function

looks like you have `messages_call` and `prompt_call`

holmes/core/conversations.py
```diff
@@ -0,0 +1,306 @@
...
+
+    template_context = {
+        "investigation": investigation_analysis,
+        "tools_called_for_investigation": truncated_investigation_result_tool_calls,
+        "issue": issue_chat_request.issue_type,
+    }
+    system_prompt_with_truncated_tools = load_and_render_prompt(
+        template_path, template_context
+    )
+    conversation_history[0]["content"] = system_prompt_with_truncated_tools
```
arikalon1, line 293: can you explain this?

arikalon1 (approved):
nice work

left a few comments

arikalon1:
@itisallgood if this is not relevant, please close it

holmes/core/conversations.py
itisallgood: We always expect conversation_history to be in the format we returned it back, and we expect the first item to be this 
{"role": "system", "content": "rendered_prompt"}

arikalon1: ok, thx

PR #165, opened 2024-10-25
feat: add support for passing runbook URLs to issue investigator

nherment:
This PR adds the ability to explicitly direct holmesgpt to specific runbook web pages.
In addition, the webpage fetch no longer use w3m but uses a combination of playwright and pandoc to generate markdown.


holmes/core/supabase_dal.py
```diff
@@ -173,9 +174,18 @@ def get_resource_instructions(self, type: str, name: str) -> List[str]:
             .execute()
         )
         if res.data:
-            return res.data[0].get("runbook").get("instructions")
+            instructions = res.data[0].get("runbook").get("instructions")
+            documents_data = res.data[0].get("runbook").get("documents")
+            documents = []
+            for document_data in documents_data:
```
arikalon1: Will this fail if there are no documents? (documents_data will be None, so not iterable)?

server.py
```diff
@@ -84,9 +84,10 @@ def investigate_issues(investigate_request: InvestigateRequest):
             investigate_request.context.get("robusta_issue_id")
         )
 
-        instructions = dal.get_resource_instructions(
-            "alert", investigate_request.context.get("issue_type")
-        )
+        # resource_instructions = dal.get_resource_instructions(
```
arikalon1: did you remove this on purpose, or just forgot to handle?
I think this remove fetching the alert instructions

Dockerfile
```diff
@@ -100,5 +100,7 @@ RUN git config --global core.symlinks false
 # Remove setuptools-65.5.1 installed from python:3.11-slim base image as fix for CVE-2024-6345 until image will be updated
 RUN rm -rf /usr/local/lib/python3.11/site-packages/setuptools-65.5.1.dist-info
 
+RUN python -m playwright install --with-deps chromium
```
arikalon1, line 103: why do we need to install it both in `pyproject.toml` and here?

Dockerfile
```diff
@@ -71,9 +70,10 @@ COPY . /app
 # We're installing here libexpat1, to upgrade the package to include a fix to 3 high CVEs. CVE-2024-45491,CVE-2024-45490,CVE-2024-45492
 RUN apt-get update \
     && apt-get install -y \
-       git \
-       apt-transport-https \
-       gnupg2 \
+    git \
+    apt-transport-https \
+    gnupg2 \
+    pandoc \
```
arikalon1, line 76: `pypandoc` needs `pandoc` on the image to work?

arikalon1 (approved):
very nice work

Left a few minor comments

server.py
nherment: That is a mistake and the next line after the commented ones should be removed.

Dockerfile
nherment: Yes. It is possible to download it from the code though if we prefer to do that.

Cf. https://pypi.org/project/pypandoc/ => `Installing pandoc via pypandoc`:

```py
pip install pypandoc
from pypandoc.pandoc_download import download_pandoc
# see the documentation how to customize the installation path
# but be aware that you then need to include it in the `PATH`
download_pandoc()
```

Dockerfile
nherment: This here is installing the required browsers for playwright to run. This is not done automatically when the python `playwright` dependency is installed.

Similarly to pypandoc, I believe we could do this at runtime but thought it'd be cleaner like this.

holmes/core/supabase_dal.py
nherment: Yes good catch, this will fail. Fixed

PR #166, opened 2024-10-29
Avoid listing tools used during investigations

nherment:
This change leads openAI gpt-4o to not return the Tools used directly in the text response.

This has been tested for the following models:
- gpt-4o
- gpt-3.5-turbo
- o1-mini 
- claude-3-opus-20240229
- claude-3-5-sonnet-20241022


aantn:
@nherment is it possible to test this automatically using our internal AI evals framework? (e.g. to add a condition for the Prometheus alert tests which verifies that for each test there is no "Tools" section in the output.)

That would give us confidence that the result is consistent not just across models but also across different alerts.

nherment:
> @nherment is it possible to test this automatically using our internal AI evals framework? (e.g. to add a condition for the Prometheus alert tests which verifies that for each test there is no "Tools" section in the output.)

This is done.

aantn (approved):
Looks good

PR #174, opened 2024-11-01
Support custom LLM implementation

nherment:
This adds the ability to inject your own LLM into holmesgps when you invoke it as a library.


holmes/core/tool_calling_llm.py
```diff
@@ -147,26 +89,21 @@ def call(self, system_prompt, user_prompt, post_process_prompt: Optional[str] =
...
             logging.debug(f"sending messages {messages}")
             try:
-                full_response = litellm.completion(
-                    model=self.model,
-                    api_key=self.api_key,
+                full_response = self.llm.completion(
                     messages=messages,
                     tools=tools,
                     tool_choice=tool_choice,
-                    base_url=self.base_url,
```
arikalon1, line 184: why did we remove the base_url, temperature and drop_params?

arikalon1 (commented):
nice work

I think we missed something.
This [method](https://github.com/robusta-dev/holmesgpt/blob/830bcec379ad1b659f85879ef1d972fdcae5cd41/holmes/config.py#L143), still returns `ToolCallingLLM`

I think it causes an issue [here](https://github.com/robusta-dev/holmesgpt/blob/830bcec379ad1b659f85879ef1d972fdcae5cd41/server.py#L203) because it doesn't contain the required methods anymore

WDYT?

holmes/core/tool_calling_llm.py
nherment: Fixed temp & drop_params. base_url is implementated in the `DefaultLLM`

nherment:
>I think we missed something.
> This [method](https://github.com/robusta-dev/holmesgpt/blob/830bcec379ad1b659f85879ef1d972fdcae5cd41/holmes/config.py#L143), still returns ToolCallingLLM

> I think it causes an issue [here](https://github.com/robusta-dev/holmesgpt/blob/830bcec379ad1b659f85879ef1d972fdcae5cd41/server.py#L203) because it doesn't contain the required methods anymore

The code has been updated to reach into the llm object of the toolCalling instance:
image

arikalon1 (approved):
nice work

PR #176, opened 2024-11-04
Reduce image size by removing pandoc and using firefox instead of chromium

nherment:
This PR:
- Removes pandoc as an external dependency
- Uses firefox instead of chromium as its install size is smaller by ~200MB
- Disables the test that fetches a webpage through an investigation b/c it is unstable and partially depends on cluster state
- Add an integration test for the URL fetcher.


arikalon1 (approved):
LGTM

PR #178, opened 2024-11-05
bug fixes:

arikalon1:
Fix bug in jwt retry - patch was not called
Fix bug if a non existing tool was called
Fix bug if tool arguments is not a valid json
Add MODEL_TYPE env var, to support azure token counting


nherment (approved):
Looks good.

PR #179, opened 2024-11-05
support removing bedrock prefix for model name token

arikalon1:


holmes/core/tool_calling_llm.py
```diff
@@ -110,6 +110,8 @@ def _strip_model_prefix(self) -> str:
         model_name = self.model
         if model_name.startswith('openai/'):
             model_name = model_name[len('openai/'):]  # Strip the 'openai/' prefix
+        elif model_name.startswith('bedrock/'):
+            model_name = model_name[len('bedrock/'):]  # Strip the 'openai/' prefix
```
nherment: incorrect comment on that line.

nherment (approved):
Minor comment.

holmes/core/tool_calling_llm.py
arikalon1: fixing, thanks

PR #185, opened 2024-11-07
Add tags to ask holmes

nherment:


holmes/core/models.py
```diff
@@ -82,16 +82,11 @@ class ConversationRequest(BaseModel):
     include_tool_call_results: bool = False
 
 
-class HolmesConversationIssueContext(BaseModel):
-    investigation_result: IssueInvestigationResult
-    issue_type: str
```
nherment, line 87: If you look a little bit above, this class is a duplicate so I removed it.

helm/holmes/values.yaml
```diff
@@ -1,12 +1,10 @@
-
-
 certificate: "" # base64 encoded
 logLevel: INFO
 
 additionalEnvVars: []
 imagePullSecrets: []
 
-allowedToolsets: "kubernetes/core"
+allowedToolsets: "*/core"
```
arikalon1: Does this loads by default the `Internet` and `Findings` toolsets?

helm/holmes/values.yaml
```diff
@@ -34,4 +32,4 @@ resources:
     cpu: 50m
     memory: 512Mi
   limits:
-    memory: 512Mi
\ No newline at end of file
+    memory: 512Mi
```
arikalon1: here as well, we changed the default to 1024

holmes/main.py
```diff
@@ -1,4 +1,6 @@
 import os
+from holmes.core.conversations import build_chat_messages
```
arikalon1: please move the imports after line 12

On the first line of the process, we load custom certificates, if the user configured any
This is required to trust self signed certs

In the past, doing imports before the custom cert was loaded, caused an http client to be initialized without the cert, and failing

So to prevent it, we start importing only after loading the custom certs

holmes/main.py
```diff
@@ -781,6 +783,18 @@ def opsgenie(
                 f"[bold]Not updating alert {issue.url}. Use the --update option to do so.[/bold]"
             )
 
+def chat(chat_request:ChatRequest, ai:ToolCallingLLM) -> ChatResponse:
```
arikalon1: This is used only on the server (reset API) and test
Any reason not to put it on the server.py?

holmes/plugins/toolsets/findings.py
```diff
@@ -0,0 +1,66 @@
+import os
+from typing import Any, Optional, cast
+from typing_extensions import Dict
+from holmes.core.supabase_dal import SupabaseDal
+from holmes.core.tools import Tool, ToolParameter, Toolset, ToolsetCommandPrerequisite
+from pydantic import SecretStr
+#!/usr/bin/env python
```
arikalon1: is this needed?

holmes/plugins/toolsets/findings.py
```diff
@@ -0,0 +1,66 @@
...
+import re
+import sys
+import logging
+from holmes.plugins.sources.prometheus.plugin import AlertManagerSource
+
+PARAM_FINDING_ID = "id"
+
+class FetchRobustaFinding(Tool):
+
+    _dal:SupabaseDal = SupabaseDal()
```
arikalon1: I think it would be best to inject the supabase dal, instead of creating a new one
This dal wrapped a supabase client, with async threads for refreshing tokens, etc
I think it's better to use one instance of it  (the one we initialize on the server.py)

holmes/plugins/toolsets/findings.py
```diff
@@ -0,0 +1,66 @@
...
+from typing_extensions import Dict
+from holmes.core.supabase_dal import SupabaseDal
+from holmes.core.tools import Tool, ToolParameter, Toolset, ToolsetCommandPrerequisite
+from pydantic import SecretStr
+#!/usr/bin/env python
+
+import re
+import sys
+import logging
+from holmes.plugins.sources.prometheus.plugin import AlertManagerSource
```
arikalon1: this one, and some above looks like unused imports

holmes/plugins/toolsets/findings.py
```diff
@@ -0,0 +1,66 @@
...
+                PARAM_FINDING_ID: ToolParameter(
+                    description="The id of the finding to fetch",
+                    type="string",
+                    required=True,
+                )
+            },
+        )
+
+    def _fetch_finding(self, finding_id:str) -> Optional[Dict]:
+        if os.environ.get("ROBUSTA_AI"):
```
arikalon1: I think lines 35-36 are not used
The ROBUSTA_AI env var, indicates which LLM is used. (Robusta's llm, or azure, openai etc)
It doesn't say if we're connected to the saas

I think we shouldn't do any check here, just do a call to the dal
And, to add a pre-req check to the tool, that verify we have a dal object
If we do, it's initialized, and we can connect to supabase

WDYT?

holmes/utils/tags.py
```diff
@@ -0,0 +1,72 @@
+
+
+from typing_extensions import Dict, List
+import re
+import json
+from copy import deepcopy
+
+def format_tag(tag:Dict[str, str]):
```
arikalon1: can you please add return type?
Also, I think it would help to add a comment, with a few examples of input tags, and the output string for it
I have some prior knowledge, so I have some idea of what this means
But the code it self, is not self explanatory.

holmes/utils/tags.py
```diff
@@ -0,0 +1,72 @@
...
+def format_messages_tags(messages:List[Dict[str, str]]) -> List[Dict[str, str]]:
+    """
+        Parses the user messages for tags and format these.
+        This method returns a shallow copy of the messages list with the exception of the messages that have been parsed.
+    """
+    formatted_messages = []
+
+    for message in messages:
+        if message.get("role") == "user":
+            formatted_str = format_message_tags(message.get("content"))
```
arikalon1: I would name the inner function a little different, just for easier readability
`format_message_tags` is very similar to `format_messages_tags`. At first glance I thought this is recursion :)

server.py
```diff
@@ -1,5 +1,6 @@
 import os
 from holmes.utils.cert_utils import add_custom_certificate
+from holmes.utils.robusta import load_robusta_api_key
```
arikalon1: same here, I would move the import after loading the custom cert

arikalon1 (approved):
nice work!

Left a few comments

Some other thoughts:
When investigating a workload (App), we fetch from the saas alerts that were firing on the apps, and also instructions ([here](https://github.com/robusta-dev/holmesgpt/blob/6a16ed76ea82fb1918ed673a033fe0bcad77b171/server.py#L126)).
If we want the same behavior, maybe we also need a tool for those
We also did some demo, for fetching workload related changes([here](https://github.com/robusta-dev/holmesgpt/blob/433d49c54cf7d0df3b1d487fd3bb43c19074ad62/holmes/core/supabase_dal.py#L158)), which I assume we'll want to add at some point. Maybe this is also a tool?

For investigating alerts, we also pre-fetch runbooks ([here](https://github.com/robusta-dev/holmesgpt/blob/6a16ed76ea82fb1918ed673a033fe0bcad77b171/server.py#L90))

helm/holmes/values.yaml
nherment: Yes. Added a unit test for the function matching the pattern against toolsets.

helm/holmes/values.yaml
nherment: I'll merge master in.

holmes/main.py
nherment: Ok, noted and moved.

holmes/main.py
nherment: The intention is to call the same method as the server.py from the test.

The goal is to improve testability of server.py as currently I don't think I can import its methods from the tests. 

High level I can think of two ways to do this (not necessarily exclusive):
- One approach is to create holmes primitives used by both server and CLI where each of the CLI and server are very thin. This is already done to some extent through the tool_calling_llm module but there is still a lot of logic in both the CLI and server so my gut feeling is that the level of abstraction is not high enough.
- Another approach is to move server.py down into the Holmes packages.


I would agree though that putting this chat method in main.py is not an improvement. I'll move it, possibly in its own module.

holmes/plugins/toolsets/findings.py
nherment: No. copied over from internet.py (it was called through YAML as an initial implementation). Removed there as well.

holmes/plugins/toolsets/findings.py
nherment: I'll pass the changes through a linter to remove all unused imports.

holmes/plugins/toolsets/findings.py
nherment: ok I'll look into how to pass it down to the toolset(s)

holmes/plugins/toolsets/findings.py
nherment: Yes agree

holmes/utils/tags.py
nherment: added

holmes/utils/tags.py
nherment: renamed to `stringify_tag`, `format_tags_in_string` and `parse_message_tags`

server.py
nherment: done

helm/holmes/values.yaml
nherment: Ended up changing that back to `allowedToolsets: "kubernetes/core,findings,internet"`

holmes/main.py
nherment: I ended up reverting that change

PR #192, opened 2024-11-08
fix: incorrect prerequisite for webpage: use python -m to invoke playwright exec and specify browser

nherment:


holmes/core/tools.py
```diff
@@ -173,8 +173,8 @@ def check_prerequisites(self):
                         return
                 except subprocess.CalledProcessError as e:
                     self._enabled = False
-                    self._disabled_reason = f"prereq check failed with errorcode {e.returncode}"
-                    logging.debug(f"YAMLToolset {self.name} : Failed to run prereq command {prereq}; {str(e)}")
+                    self._disabled_reason = f"prereq check failed with errorcode {e.returncode}: {str(e)}"
```
nherment, line 176: @arikalon1 added this for easier troubleshooting in the future. I'm not sure if it was left out on purpose in which case I can revert.

arikalon1: I'm not sure if it was on purpose, but it looks fine to me

arikalon1 (approved):
LGTM

PR #193, opened 2024-11-08
Only load builtin toolsets when the config is initialised

nherment:
This PR ensures builtin toolsets are loaded only at initialization time.
This makes sure that the prerequisite checks for each toolset is only run at startup and not for every API call to the server.

If we want this behaviour to also apply to the custom toolsets, then this PR is not enough.


arikalon1 (approved):
LGTM

PR #199, opened 2024-11-12
Improve k8s function definitions

moshemorad:


holmes/plugins/toolsets/kubernetes.yaml
```diff
@@ -5,17 +5,27 @@ toolsets:
 
   tools:
   - name: "kubectl_describe"
-    description: "Run kubectl describe on a Kubernetes resource"
+    description: > 
+      Run kubectl describe <kind> <name> -n <namespace>, 
+      call this when users ask for description, 
+      for example when a user asks
+        - 'desctibe pod xyz-123'
```
aantn: typo in describe

holmes/plugins/toolsets/kubernetes.yaml
```diff
@@ -5,17 +5,27 @@ toolsets:
...
   tools:
   - name: "kubectl_describe"
-    description: "Run kubectl describe on a Kubernetes resource"
+    description: > 
+      Run kubectl describe <kind> <name> -n <namespace>, 
+      call this when users ask for description, 
+      for example when a user asks
+        - 'desctibe pod xyz-123' 
+        - 'show service xyz-123 in namespace my-ns'
+        - 'show information about deployment my-deployment"
```
aantn: will that really work? because we don't know the namespace. you probably want to use kubectl_find_resource in that case

moshemorad: I think it is very dependent in what you are actually want to see, for me when i say show me information the find_resource alone won't be enough. But yes there is some issues with not providing the namespace.

I think that this is something that we need to consider further, i will remove it for now.

aantn: You can add it back! Just add the namespace to your example.

aantn (approved):

RoiGlinik (approved):

PR #204, opened 2024-11-14
Add llm evaluations

nherment:
This PR has mostly 3 changes:
- Improvement of the prompt to prevent the LLM mentioning separately `kubectl logs` and `kubectl logs --previous`
- Introduction of a test suite for investigations, including the ability to mock the DB access
- Integration with brantrust,dev although that code is currently commented out as I need to talk to them as I hit the freemium limits.

It's a big PR and most of it is the refactor of the existing mock mechanism to work well with the DAL.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,6 +70,8 @@ RUN python -m playwright install firefox --with-deps
 # We're installing here libexpat1, to upgrade the package to include a fix to 3 high CVEs. CVE-2024-45491,CVE-2024-45490,CVE-2024-45492
 RUN apt-get update \
     && apt-get install -y \
+    curl \
+    jq \
     git \
     apt-transport-https \
     gnupg2 \
--- a/holmes/core/llm.py
+++ b/holmes/core/llm.py
@@ -125,14 +125,7 @@ class DefaultLLM(LLM):
             drop_params=drop_params
         )
 
-
-
         if isinstance(result, ModelResponse):
-            response = result.choices[0]
-            response_message = response.message
-            # when asked to run tools, we expect no response other than the request to run tools unless bedrock
-            if response_message.content and ('bedrock' not in self.model and logging.DEBUG != logging.root.level):
-                logging.warning(f"got unexpected response when tools were given: {response_message.content}")
             return result
         else:
             raise Exception(f"Unexpected type returned by the LLM {type(result)}")
--- a/holmes/core/tool_calling_llm.py
+++ b/holmes/core/tool_calling_llm.py
@@ -117,7 +117,7 @@ class ToolCallingLLM:
                     messages, max_context_size, maximum_output_token
                 )
 
-            logging.debug(f"sending messages {messages}")
+            logging.debug(f"sending messages={messages}\n\ntools={tools}")
             try:
                 full_response = self.llm.completion(
                     messages=parse_messages_tags(messages),
@@ -127,7 +127,7 @@ class ToolCallingLLM:
                     response_format=response_format,
                     drop_params=True,
                 )
-                logging.debug(f"got response {full_response}")
+                logging.debug(f"got response {full_response.to_json()}")
             # catch a known error that occurs with Azure and replace the error message with something more obvious to the user
             except BadRequestError as e:
                 if (
--- a/holmes/main.py
+++ b/holmes/main.py
@@ -51,7 +51,7 @@ app.add_typer(generate_app, name="generate")
 
 class Verbosity(Enum):
     NORMAL = 0
-    LOG_QUERIES = 1
+    LOG_QUERIES = 1  # TODO: currently unused
     VERBOSE = 2
     VERY_VERBOSE = 3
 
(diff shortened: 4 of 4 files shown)


tests/llm/utils/system.py
```diff
@@ -0,0 +1,31 @@
...
+import os
+import pwd
+from datetime import datetime
+from typing import Dict
+from pathlib import Path
+
+
+def get_active_branch_name():
+
+    head_dir = Path(".") / ".git" / "HEAD"
```
moshemorad: You can move all the relevant parts to the init
e.g. 
```
Path(".",".git","HEAD")
```

tests/llm/utils/mock_toolset.py
```diff
@@ -61,6 +61,7 @@ def _auto_generate_mock_file(self, params:Dict):
         with open(mock_file_path, 'w') as f:
             f.write(mock_metadata_json + '\n')
             f.write(output)
+            f.close()
```
moshemorad: Why you need it? when the context manager is exited it make sure that the file is closed.

tests/llm/utils/mock_dal.py
```diff
@@ -0,0 +1,78 @@
...
+        else:
+            data = super().get_issue_data(issue_id)
+            file_path = self._get_mock_file_path("issue_data")
+
+            with open(file_path, 'w') as f:
+                f.write(json.dumps(data or {}, indent=2))
+                f.close()
+
+            logging.warning(f"A mock file was generated for you at {file_path} with the contentof dal.get_issue_data({issue_id})")
+            if self._dal_passthrough:
```
moshemorad, line 35: Just to make sure i understand, that passthourgh check is here after the call so we will be able to autogenerate the data? 
Maybe we should have dedicated task for generating relevant data instead so it will simplify the code itself?

tests/llm/utils/mock_dal.py
```diff
@@ -0,0 +1,78 @@
...
+                return data
+            else:
+                raise Exception(f"dal.get_issue_data({issue_id}) was invoked and is not mocked. A mock file was generated for you at {file_path}. Remove the '{AUTO_GENERATED_FILE_SUFFIX}' suffix to enable that file")
+
+    def get_resource_instructions(self, type: str, name: Optional[str]) -> Optional[ResourceInstructions]:
+        if self._resource_instructions is not None:
+            return self._resource_instructions
+        else:
+            print(f"**** self._resource_instructions={str(self._resource_instructions)}")
```
moshemorad: In that case the resource_instructions is always None, right? any reason to indeed print it?
Maybe we should move it to other line?

Also consider use logger instead of print that way we can configure the level in test easily.

tests/llm/utils/mock_dal.py
```diff
@@ -0,0 +1,78 @@
...
+            logging.warning(f"A mock file was generated for you at {file_path} with the contentof dal.get_resource_instructions({type}, {name})")
+            if self._dal_passthrough:
+                return data
+            else:
+                raise Exception(f"dal.get_resource_instructions({type}, {name}) was invoked and is not mocked. A mock file was generated for you at {file_path}. Remove the '{AUTO_GENERATED_FILE_SUFFIX}' suffix to enable that file")
+
+    def _get_mock_file_path(self, entity_type:str):
+        return f"{self._test_case_folder}/{entity_type}.json{AUTO_GENERATED_FILE_SUFFIX}"
+
+pydantic_resource_instructions = TypeAdapter(ResourceInstructions)
```
moshemorad, line 60: Why do you need TypeAdapter in that case? it is already a pydantic model. Im missing something?

tests/llm/utils/mock_utils.py
```diff
@@ -0,0 +1,205 @@
...
+    contextual_precision: float = 0.5
+    contextual_recall: float = 0
+    contextual_relevancy: float = 0
+
+class Message(BaseModel):
+    message: str
+
+T = TypeVar('T')
+
+class HolmesTestCase(BaseModel, Generic[T]):
```
moshemorad: Can we remove the Generic? i don't see it is used anywhere..

tests/llm/utils/mock_dal.py
```diff
@@ -0,0 +1,78 @@
...
+
+def load_mock_dal(test_case_folder:Path, dal_passthrough:bool = False):
+    issue_data_mock_path = test_case_folder.joinpath(Path("issue_data.json"))
+    issue_data = None
```
moshemorad: Reuse functions from mock_utils like load_issue_data etc..

tests/llm/utils/mock_dal.py
nherment: > that passthourgh check is here after the call so we will be able to autogenerate the data?

Yes

>we should have dedicated task for generating relevant data instead so it will simplify the code itself?

Yes, I agree. That also links with the current ai-tests methodology that relies on a cluster that has these issues. We should either fall back on "live" behaviour with the ability to generate mocks, or have simple mocks with simpler logic.

tests/llm/utils/mock_dal.py
nherment: removed

tests/llm/utils/mock_dal.py
nherment: Doesn't that save me from deserialising the json myself? I can change that if you think it's better.

tests/llm/utils/mock_dal.py
nherment: done

tests/llm/utils/mock_toolset.py
nherment: removed

arikalon1 (approved):
LGTM

PR #210, opened 2024-11-21
feat: suport entra id

nherment:


holmes/core/llm.py
```diff
@@ -41,6 +41,18 @@ def completion(self, messages: List[Dict[str, Any]], tools: Optional[List[Tool]]
         pass
 
 
+def get_lite_llm_config(api_key:Optional[str], base_url:Optional[str]) -> Dict[str, Any]:
+
+    if os.environ.get("AZURE_API_BASE"):
```
arikalon1, line 63: will this work with the way we guide users to use azure openai?
https://docs.robusta.dev/master/configuration/ai-analysis.html#choosing-and-configuring-an-ai-provider

nherment: it should, This just avoids setting the api_key and base_url explicitly but litellm can  read these though env vars, which is how we tell users to configure holmes.

nherment:
Closing this. It should be revisited once litellm allows directly using entra id with an AD token

PR #211, opened 2024-11-22
Improve logging

aantn:


holmes/core/tool_calling_llm.py
```diff
@@ -118,7 +118,7 @@ def call(
                     messages, max_context_size, maximum_output_token
                 )
 
-            logging.debug(f"sending messages {messages}")
+            logging.debug(f"sending messages={messages}\n\ntools={tools}")
```
nherment, line 121: At some point we may want to move to structured logs to make analysis, search etc. simpler.

Logs like this spanning multiple lines can be mangled in prod environment. This being Holmes and debug level is likely fine.

nherment (approved):
LGTM

holmes/core/tool_calling_llm.py
aantn: Yeah, good call. I think it's OK for now because at debug level there's is a ton of unstructured spam from third party libraries, so this wont make things worse.

PR #214, opened 2024-11-27
Llm eval improvements

nherment:
- improvements to LLM evals
  - Change the mock behaviour, now controlled through `generate_mocks=True` in test_case.yaml
  - Lower default faithfulness success threshold from .5 to .3
  - Added LLM test readme at `tests/llm/README.md`
  - Added ability to run live LLM tests (`RUN_LIVE=1`)
  - An exception is no longer thrown if a tool is not mocked. All tools are passthrough.
- Import of promptfoo's ask_holmes easy and medium test cases


tests/llm/utils/langfuse.py
```diff
@@ -0,0 +1,98 @@
+
```
nherment, line 1: This is not used but I suggest we keep it until we make a ruling on braintrust

pyproject.toml
```diff
@@ -35,7 +35,7 @@ boto3 = "^1.34.145"
 setuptools = "^72.1.0"
 aiohttp = "^3.10.2"
 cachetools = "^5.5.0"
-playwright = "^1.48.0"
+playwright = "1.48.0"
```
moshemorad, line 38: Why we pinned it?

tests/llm/fixtures/test_ask_holmes/07_high_latency/helm/Dockerfile
moshemorad, line 1: Thinking here loud, do we really want to duplicate those and not use the ones from the kubernetes demo repo?

moshemorad (changes requested):

pyproject.toml
nherment: There was some issue with the newly released playwright wheel. It's fixed now and that change was reverted.

tests/llm/fixtures/test_ask_holmes/07_high_latency/helm/Dockerfile
nherment: I think we want to avoid replicating indeed. Ideally we have everything in one place though but there is usefulness to the k8s demo repo beyond Holmes.

moshemorad: So why aren't we just use them from the demo repo?

nherment: This one `07_high_latency` is modified and not exactly the same as the k8s demo: it does not actually require DB connection. Others like `21_job_fail_curl_no_svc_account` have been ported as is from promptfoo but in most cases where promptfoo was using the k8s demo code directly, the same k8s demo code is used.

nherment:
@moshemorad what chages are needed to greenlight this PR?

nherment:
The previous commit freezes playwright to 1.48.0 due to https://github.com/microsoft/playwright-python/issues/2679

arikalon1 (approved):

PR #216, opened 2024-11-29
test: llm eval, add correctness score, port alertmanager test cases

nherment:


holmes/main.py
```diff
@@ -368,6 +368,7 @@ def alertmanager(
         logging.error(f"Failed to fetch issues from alertmanager", exc_info=e)
         return
 
+    write_json_file(f"./issue_{alertmanager_label}.json", issues)
```
arikalon1: is this, and line 401 below is something we want for the cli?
What's the use of those files?
Or it's just code to generate tests data?

arikalon1 (approved):
nice work

left a small comment

holmes/main.py
nherment: It was to generate data. Removed.

arikalon1 (approved):

PR #217, opened 2024-11-30
 Toolset integrations

itisallgood:


holmes/core/supabase_dal.py
```diff
@@ -264,3 +267,53 @@ def get_workload_issues(self, resource: dict, since_hours: float) -> List[str]:
...
+            )
+        except Exception as error:
+            logging.error("Error happened during upserting holmes status", exc_info=True)
+
+        return None
+    
+    def sync_toolsets(self, toolsets) -> None:
+        provided_toolset_names = [toolset['toolset_name'] for toolset in toolsets]
+        try:
+            self.client.table("HolmesToolsStatus").upsert(
```
moshemorad: Use const

nherment:
@itisallgood 
Moving the prerequisite check for the toolsets when the server is started is a great improvement.
In the current code, both the `Internet` and `Findings` toolset do run the prerequisites when they are instantiated. This means that each of these toolset will still run the prerequisites for each request to Holmes server.

holmes/utils/holmes_sync_toolsets.py
```diff
@@ -0,0 +1,161 @@
...
+                **toolset.model_dump(exclude_none=True),
+                toolset_name=toolset.name,
+                cluster_id=CLUSTER_NAME,
+                account_id=dal.account_id,
+                status=toolset.get_status(),
+                error=toolset.get_error(),
+                updated_at=updated_at
+            ).model_dump(exclude_none=True)
+        )
+    print(db_toolsets)
```
nherment: Is that a leftover?

holmes/core/supabase_dal.py
```diff
@@ -262,5 +266,64 @@ def get_workload_issues(self, resource: dict, since_hours: float) -> List[str]:
...
+
+    def upsert_holmes_status(self, holmes_status_data: dict) -> None:
+        updated_at = datetime.now().isoformat()
+        try:
+            res = (
+                self.client
+                .table(HOLMES_STATUS_TABLE)
+                .upsert({
+                    "account_id": self.account_id,
+                    "updated_at": updated_at,
```
nherment, line 277: Not super familiar with the supabase client but can't we use postgresql to set the updated_at ?
Typically in SQL something like `updated_at = NOW()` would work well and would ensure a single source of truth for the time instead of relying on the clients.

holmes/core/supabase_dal.py
```diff
@@ -262,5 +266,64 @@ def get_workload_issues(self, resource: dict, since_hours: float) -> List[str]:
...
+            )
+        except Exception as error:
+            logging.error(f"Error happened during upserting holmes status: {error}", 
+                          exc_info=True)
+
+        return None
+    
+    def sync_toolsets(self, toolsets: list[dict]) -> None:
+        if not toolsets:
+            logging.error("No toolsets were provided for synchronization.")
```
nherment: Is that really an error or a warning or even an info level log?
I'd suggest keeping errors for things someone should actively investigate and work to fix. 
In this case, could having no toolset be intentional? If so then maybe info or warning level logs are enough.

nherment (approved):
Looks good! Some small comments.

holmes/core/supabase_dal.py
itisallgood: Agree

holmes/core/supabase_dal.py
itisallgood: @nherment  Unfortunately, it won't be solved that easily,  NOW() works only for INSERT, not UPDATE. So it would require from us to create trigger. I don't know what's better solution - trigger in our database or update relying on clients

nherment: `"updated_at": "NOW()",` should work as expected.

nherment: I suspect supabase just copies the column content into the SET part of a postgresql upsert.

```
INSERT INTO HOLMES_STATUS_TABLE (..., updated_at, ...) 
VALUES (..., NOW(), ...)
ON CONFLICT DO UPDATE
SET
  "updated_at"= NOW(),
```

helm/holmes/values.yaml
```diff
@@ -26,7 +26,8 @@ serviceAccount:
   imagePullSecrets: []
   annotations: {}
 
-toolsets: []
+enabledByDefaultToolsets: "kubernetes/core,kubernetes/logs,findings,internet"
```
arikalon1: why do we need this?

Can't we just write:
```
toolsets:
  kubernetes/core:
    enabled: true
  kubernetes/logs:
    enabled: true
  findings:
    enabled: true
  internet:
    enabled: true  
```

holmes/core/supabase_dal.py
```diff
@@ -262,5 +266,64 @@ def get_workload_issues(self, resource: dict, since_hours: float) -> List[str]:
...
+                'toolset_name', provided_toolset_names
+            ).execute()
+
+            logging.info("Toolsets synchronized successfully.")
+
+        except Exception as e:
+            logging.exception(f"An error occurred during toolset synchronization: {e}",
+                              exc_info=True)
+
+    def get_toolsets_for_holmes(self,):
```
arikalon1: why are we getting this from the db?

The enabled/disabled status is coming from helm
We're saving it to report the status to the saas, but to read it, we have it locally (locally is the source of truth)

holmes/utils/holmes_sync_toolsets.py
```diff
@@ -0,0 +1,155 @@
...
+def merge_and_override_bultin_toolsets_with_toolsets_config(
+    toolsets_loaded_from_config: list[ToolsetYamlFromConfig],
+    default_toolsets_by_name: dict[str, YAMLToolset],
+    enabled_by_default_toolsets: list[YAMLToolset],
+) -> dict[str, YAMLToolset]:
+    """
+    Merges and overrides default_toolsets_by_name with custom 
+    config from /etc/holmes/config/custom_toolset.yaml
+    """
+    toolsets_with_updated_statuses = {}
```
arikalon1: why do we need this code?
Can't we get that logic free from helm?
Since the toolsets are dictionaries, wouldn't helm solve the merge for us?

itisallgood: @arikalon1 This particular code is not related to helm logic. It loads both python and yaml toolsets from `holmes/plugins/toolsets` and updates them with helm config

nherment (dismissed):

helm/holmes/values.yaml
```diff
@@ -1,12 +1,12 @@
 certificate: "" # base64 encoded
 logLevel: INFO
 
+clusterName: ""
```
RoiGlinik: In what cases do we need this?

holmes/plugins/toolsets/__init__.py
```diff
@@ -9,24 +9,45 @@
...
+from typing import Dict
+from pydantic import BaseModel
+from typing import Optional
+import yaml
 
 THIS_DIR = os.path.abspath(os.path.dirname(__file__))
 
-class ListOfToolSets(BaseModel):
+
+class FallbackOldStructureToolsetsList(BaseModel):
```
RoiGlinik: i think we can delete it if its not used anymore

RoiGlinik (approved):

helm/holmes/values.yaml
itisallgood: You are right, we don't need it anymore

nherment (approved):

PR #220, opened 2024-12-06
ArgoCD toolset

nherment:


holmes/plugins/toolsets/argocd.yaml
```diff
@@ -0,0 +1,37 @@
+toolsets:
+  - name: "argocd/core"
+    prerequisites:
+      - command: "argocd version"
+
+    tools:
+      - name: "argocd_app_get"
+        description: "Retrieve information about an existing application, such as its status and configuration"
+        command: "argocd app get {{ app_name }} --show-operation -o yaml --auth-token {{ auth_token }}"
```
arikalon1: I think we don't want to auth token to be sent to the llm
Can we somehow pass the token via env var?
So, for example, in order to use the tool, the user need to inject the env var: ARGO_AUTH_TOKEN,
and the tool command will be:
```
command: "argocd app get {{ app_name }} --show-operation -o yaml --auth-token $ARGO_AUTH_TOKEN"
```
Can something like this work?

nherment: Turns out we don't event need to use `--auth-token` as argocd will read the env var `ARGO_AUTH_TOKEN` by itself. Reflected in the documentation.

holmes/plugins/toolsets/argocd.yaml
```diff
@@ -0,0 +1,42 @@
+toolsets:
+  argocd/core:
+    description: "Set of tools to get argocd metadata like list of apps, repositories, projects, etc."
+    docs_url: "https://argo-cd.readthedocs.io/en/stable/user-guide/commands/argocd/"
+    icon_url: "https://argo-cd.readthedocs.io/en/stable/assets/logo.png"
+    tags:
+      - core
+    prerequisites:
```
itisallgood, line 8: Should we add `ARGOCD_AUTH_TOKEN`env  as one of the prerequisites ?

itisallgood (approved):
lgtm

PR #221, opened 2024-12-06
OpenSearch toolset

nherment:


Avi-Robusta:
Screen Shot 2024-12-30 at 10 26 56
Screen Shot 2024-12-30 at 10 27 12

README.md
```diff
@@ -761,6 +772,33 @@ Configure Slack to send notifications to specific channels. Provide your Slack t
...
 </details>
 
+
+<details>
+<summary>OpenSearch Integration</summary>
+
+The OpenSearch toolset (`opensearch`) allows Holmes to consult an opensearch cluster for its health, settings and shards information.
+The toolset supports multiple opensearch or elasticsearch clusters that are configured by editing Holmes' configuration file (or in cluster to the configuration secret):
+
+```                                                                                                                           holmesgpt-DoRnpO3K-py3.11 (add_tool_elasticsearch|💩) 15:20
```
moshemorad: 💩 ?

README.md
```diff
@@ -761,6 +772,33 @@ Configure Slack to send notifications to specific channels. Provide your Slack t
...
+opensearch_clusters:
```
moshemorad, line 835: Didn't we said that this will only loaded from secret? no reason to have it here right?

helm/holmes/templates/holmes.yaml
```diff
@@ -55,6 +55,9 @@ spec:
             mountPath: /etc/robusta/config
           - name: custom-toolsets-configmap
             mountPath: /etc/holmes/config
+          - name: holmes-config-secret
+            mountPath: /root/.holmes/config.yaml
```
moshemorad: We should provide config.yaml both in the monthPath and subPath indeed?

README.md
```diff
@@ -645,6 +645,17 @@ You can view an example config file with all available settings [here](config.ex
 
 By default, without specifying `--config` the agent will try to read `~/.holmes/config.yaml`. When settings are present in both config file and cli, the cli option takes precedence.
 
+
+When running **HolmesGPT in a Kubernetes cluster**, you can securely pass configuration settings using a **Kubernetes Secret**. The secret will be mounted into the pod at `~/.holmes/config.yaml`:
```
moshemorad: @Avi-Robusta are you sure that this what we decided in the daily meeting? I though that we will enable loading only os secret with relevant info not the whole config. @arikalon1 WDYT?

holmes/plugins/toolsets/opensearch.py
```diff
@@ -0,0 +1,127 @@
...
+        return str(health)
+
+    def get_parameterized_one_liner(self, params) -> str:
+        return f"opensearch GetClusterSettings({params.get('host')})"
+
+class OpenSearchToolset(Toolset):
+    def __init__(self, clusters_configs:List[Dict]):
+        clients: List[OpenSearchClient] = []
+        for config in clusters_configs:
+            logging.info(f"Setting up OpenSearch client: {str(config)}")
```
moshemorad: Please don't print the config, it includes secrets and we don't want them to shown in logs etc...

holmes/config.py
```diff
@@ -108,6 +113,12 @@ def load_from_env(cls):
             if val is not None:
                 kwargs[field_name] = val
             kwargs["cluster_name"] = Config.__get_cluster_name()
+            try:
```
moshemorad: So we ignore all the env if file was provided as secret?

holmes/plugins/toolsets/opensearch.py
```diff
@@ -0,0 +1,127 @@
...
+from typing import Any, Dict, List, Optional
+from holmes.core.tools import StaticPrerequisite, Tool, ToolParameter, Toolset
+from opensearchpy import OpenSearch
+
+class OpenSearchClient:
+    def __init__(self, **kwargs):
+        
+        # Handle http_auth explicitly
+        http_auth = kwargs.pop("http_auth")
+        if isinstance(http_auth, dict):
```
moshemorad: Should we add it to the docs?

holmes/plugins/toolsets/opensearch.py
```diff
@@ -0,0 +1,127 @@
...
+            print(client.client.cluster.health())
+            clients.append(client)
+
+        super().__init__(
+            name = "opensearch",
+            description="Provide cluster metadata information like health, shards, settings.",
+            docs_url="https://opensearch.org/docs/latest/clients/python-low-level/",
+            icon_url="https://upload.wikimedia.org/wikipedia/commons/9/91/Opensearch_Logo.svg",
+            prerequisites = [
+                StaticPrerequisite(enabled=len(clients) > 0, disabled_reason="No opensearch client was configured")
```
moshemorad: We don't want to check the client status?

holmes/plugins/toolsets/opensearch.py
```diff
@@ -0,0 +1,127 @@
...
+            ],
+            tools = [
+                ListShards(clients),
+                GetClusterSettings(clients),
+                GetClusterHealth(clients),
+            ],
+            tags=["core",]
```
moshemorad:
```suggestion
            tags=[ToolsetTag.CORE,]
```

holmes/plugins/toolsets/__init__.py
```diff
@@ -3,12 +3,16 @@
...
+from opensearchpy.helpers.signer import Dict
+
 from holmes.core.supabase_dal import SupabaseDal
 from holmes.plugins.toolsets.findings import FindingsToolset
 from holmes.plugins.toolsets.internet import InternetToolset
 from pydantic import BaseModel
 
 from holmes.core.tools import Toolset, YAMLToolset
+from holmes.plugins.toolsets.opensearch import OpenSearchToolset
+from holmes.utils.pydantic_utils import load_model_from_file
```
moshemorad: Unused

README.md
Avi-Robusta: lol super weird, probably from rebase

README.md
Avi-Robusta: locally running holmes it is in the configuration file (in terminal), in cluster its from secret

holmes/config.py
Avi-Robusta: no, they override those params in the config file and are passed to here (they are the kwargs)
```                   
 return cls.load_from_file(config_file=DEFAULT_CONFIG_LOCATION, **kwargs)
```

holmes/plugins/toolsets/opensearch.py
Avi-Robusta: made it safer, only add healthy clients to list

holmes/plugins/toolsets/opensearch.py
Avi-Robusta: added example in docs

holmes/config.py
moshemorad: Oh missed it, thanks!

holmes/plugins/toolsets/opensearch.py
```diff
@@ -0,0 +1,131 @@
...
+        return f"opensearch GetClusterSettings({params.get('host')})"
+
+class OpenSearchToolset(Toolset):
+    def __init__(self, clusters_configs:List[Dict]):
+        clients: List[OpenSearchClient] = []
+        for config in clusters_configs:
+            try:
+                logging.info(f"Setting up OpenSearch client")
+                client = OpenSearchClient(**config)
+                client.health = client.client.cluster.health()
```
moshemorad: Why we assign it to client.health var? we want indeed add new var to the client?

holmes/plugins/toolsets/opensearch.py
```diff
@@ -0,0 +1,131 @@
...
+    def __init__(self, clusters_configs:List[Dict]):
+        clients: List[OpenSearchClient] = []
+        for config in clusters_configs:
+            try:
+                logging.info(f"Setting up OpenSearch client")
+                client = OpenSearchClient(**config)
+                client.health = client.client.cluster.health() 
+                if client.health:
+                    clients.append(client)
+            except:
```
moshemorad: Please use except Exception otherwise it might also catch exception that it shouldn't like system exceptions.

README.md
```diff
@@ -761,6 +772,39 @@ Configure Slack to send notifications to specific channels. Provide your Slack t
...
+The OpenSearch toolset (`opensearch`) allows Holmes to consult an opensearch cluster for its health, settings and shards information.
+The toolset supports multiple opensearch or elasticsearch clusters that are configured by editing Holmes' configuration file (or in cluster to the configuration secret):
+
+```                                                                                 
+opensearch_clusters:
+  - hosts:
+      - https://my_elasticsearch.us-central1.gcp.cloud.es.io:443
+    headers:
+      Authorization: "ApiKey <your_API_key>"
+# or
```
moshemorad, line 840: The # are intended?

moshemorad (changes requested):
Few small things.

README.md
Avi-Robusta: yea, incase someone copies everything, to keep codeblock still functional

README.md
Avi-Robusta: I thought we wanted the whole config to be the secret, 
doesnt make sense to only do the opensearch, since we need this for other parts of the config that cant be loaded via env var like grafana/loki , thats why we did this instead

moshemorad (dismissed):

holmes/config.py
```diff
@@ -34,13 +34,44 @@
...
+from holmes.utils.holmes_sync_toolsets import load_custom_toolsets_config, merge_and_override_bultin_toolsets_with_toolsets_config, extract_config_from_custom_toolsets_config
 from holmes.core.tools import YAMLToolset
 from holmes.common.env_vars import ROBUSTA_CONFIG_PATH
 from holmes.utils.definitions import RobustaConfig
+import re
 
 DEFAULT_CONFIG_LOCATION = os.path.expanduser("/Users/avirobusta/git/holmesgpt/config.yaml")
 
+def get_env_replacement(value: str) -> Optional[str]:
+    env_values = re.findall(r"{{[ ]*env\.(.*)[ ]*}}", value)
```
moshemorad: `(.*)` this means we allow whitespace to be captured this is indeed what we want?

holmes/config.py
```diff
@@ -34,13 +34,44 @@
...
+            env_var_value = get_env_replacement(value)
+            if env_var_value:
+                values[key] = env_var_value
+        elif isinstance(value, SecretStr):
+            env_var_value = get_env_replacement(value.get_secret_value())
+            if env_var_value:
+                values[key] = SecretStr(env_var_value)
+        elif isinstance(value, dict):
+            env_var_value = replace_env_vars_values(value)
+            if env_var_value:
```
moshemorad: Is this really needed? Aren't we do the swap in place and the dict is the same dict referance?

holmes/config.py
```diff
@@ -34,13 +34,44 @@
...
 from holmes.utils.definitions import RobustaConfig
+import re
 
 DEFAULT_CONFIG_LOCATION = os.path.expanduser("/Users/avirobusta/git/holmesgpt/config.yaml")
 
+def get_env_replacement(value: str) -> Optional[str]:
+    env_values = re.findall(r"{{[ ]*env\.(.*)[ ]*}}", value)
+    if env_values:
+        env_var_value = os.environ.get(env_values[0].strip(), None)
+        if not env_var_value:
```
moshemorad: This will fail on cases like empty string. I think it is better to check if environ contain the key or not WDYT?

holmes/config.py
```diff
@@ -108,7 +144,14 @@ def load_from_env(cls):
             if val is not None:
                 kwargs[field_name] = val
             kwargs["cluster_name"] = Config.__get_cluster_name()
-        return cls(**kwargs)
+        configs = extract_config_from_custom_toolsets_config()
+        combined = {}
+        for item in configs:
+            if isinstance(item, dict):
+                for key, value in item.items():
+                    combined[key] = value
```
moshemorad: What will happen if some tools will have the same key? it will override each other?

I think that the whole idea to move the config of tool to the tool level is to make them standalone and not continue to use the "global" config object. When we initialize the tool we can check if config provided and move it into the initialization. This will also help us to keep tools organize and understand what each tool expect to receive when configured.

Let's separate there relevant tool configs and provide them to the init when tool is loaded instead of loading it to the config and provide them to all tools.

WDYT?

moshemorad (changes requested):

holmes/config.py
Avi-Robusta: I believe there’s some misunderstanding about the goal here. We are not aiming to refactor the entire configuration system. Our goal is to continue supporting the existing configuration from the Helm chart. 

What you are proposing would break backward compatibility and disrupt Holmes' ability to run locally. This is because your approach removes the tools’ configurations from the global config, which they currently rely on.

> **What will happen if some tools have the same key? Will they override each other?**

The behavior will remain the same as it is today: if someone writes a new tool using the same key, it will override the configuration of another tool. This is why it’s essential to ensure each tool’s configuration is unique. Otherwise, conflicts between tools can occur, causing unintended overrides.

If you want to refactor the tools config on a global scale than I recommend opening a new task for it since this one has already become bloated enough

moshemorad (approved):

holmes/config.py
```diff
@@ -28,12 +32,45 @@
...
-from holmes.utils.holmes_sync_toolsets import load_custom_toolsets_config, merge_and_override_bultin_toolsets_with_toolsets_config
+from pydantic import ValidationError
+from holmes.utils.holmes_sync_toolsets import load_custom_toolsets_config, merge_and_override_bultin_toolsets_with_toolsets_config, extract_config_from_custom_toolsets_config
+from holmes.core.tools import YAMLToolset
 from holmes.common.env_vars import ROBUSTA_CONFIG_PATH
 from holmes.utils.definitions import RobustaConfig
+import re
 
-DEFAULT_CONFIG_LOCATION = os.path.expanduser("~/.holmes/config.yaml")
+DEFAULT_CONFIG_LOCATION = os.path.expanduser("/Users/avirobusta/git/holmesgpt/config.yaml")
```
arikalon1: @Avi-Robusta please make sure not to commit this

Avi-Robusta: good call

helm/holmes/templates/holmes.yaml
```diff
@@ -58,6 +58,8 @@ spec:
             mountPath: /etc/robusta/config
           - name: custom-toolsets-configmap
             mountPath: /etc/holmes/config
+          - name: holmes-config-secret
+            mountPath: /root/.holmes/
```
arikalon1: any reason we're mounting here to `/root/..` ?
is it going to be an issue on hardened clusters?

arikalon1 (approved):
looks good,
left one small comment

arikalon1 (approved):
nice work

itisallgood (approved):

arikalon1 (approved):

PR #229, opened 2024-12-13
Add kafka tool

nherment:
Add kafka tool to Holmes.
Supported authentication is only username/password or anonymous.

This PR also makes some unrelated tweaks to the code in master and adds mocks to the `ask_holmes:21_job_fail_curl_no_svc_account` test case so that it passed without the need to run it live


holmes/plugins/toolsets/kafka.py
```diff
@@ -0,0 +1,262 @@
...
+                    description="The ID of the consumer group to describe",
+                    type="string",
+                    required=True,
+                )
+            },
+        )
+        self._admin_client = admin_client
+
+    def invoke(self, params: Dict) -> str:
+        print(params)
```
moshemorad: print remove/change to log all the prints.

tests/test_tool_kafka.py
```diff
@@ -0,0 +1,155 @@
...
+def docker_compose(admin_client):
+    try:
+        subprocess.Popen(
+            "docker compose up -d".split(),
+            cwd=FIXTURE_FOLDER,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+
+        if not wait_for_containers(FIXTURE_FOLDER):
```
moshemorad: You can use --wait in the docker compose up command. it will only return after all containers report healthy
https://docs.docker.com/reference/cli/docker/compose/up/

moshemorad (changes requested):

PR #234, opened 2024-12-16
Global instructions

itisallgood:


holmes/core/tool_calling_llm.py
```diff
@@ -53,6 +53,10 @@ class ResourceInstructionDocument(BaseModel):
     url: str
 
 
+class Intructions(BaseModel):
+    instructions: List[str] = []
```
moshemorad, line 57: Why is that list of str? why not only single str?

holmes/core/conversations.py
```diff
@@ -9,6 +9,7 @@
 )
 from holmes.plugins.prompts import load_and_render_prompt
 from holmes.core.tool_calling_llm import ToolCallingLLM
+from holmes.core.tool_calling_llm import ResourceInstructionDocument, ResourceInstructions, Intructions
```
moshemorad: Unused?

holmes/plugins/prompts/generic_investigation.jinja2
```diff
@@ -6,6 +6,19 @@ Provide an terse analysis of the following {{ issue.source_type }} alert/issue a
 
 If the user provides you with extra instructions in a triple quotes section, ALWAYS perform their instructions and then perform your investigation.
 
+
+Global Instructions
```
moshemorad, line 10: Please move it to dedicated jinja template and include in all relevant places.
Also are we sure we want to add it to every call even if global instruction wasn't configured?

holmes/core/supabase_dal.py
```diff
@@ -194,6 +194,25 @@ def get_resource_instructions(self, type: str, name: Optional[str]) -> Optional[
...
+            .eq("account_id", self.account_id)  
+            .eq("subject_type", "Account")
+            .execute()
+        )
+
+            if res.data:
+                instructions = res.data[0].get("runbook").get("instructions")
+                return Intructions(instructions=instructions)
+        except Exception:
+            logging.exception("Failed to fetch global instructions", exc_info=True)
```
moshemorad, line 212: exc_info=True is redundant. This is the difference between exception end error in logger call.

holmes/core/conversations.py
```diff
@@ -177,8 +179,12 @@ def build_issue_chat_messages(issue_chat_request: IssueChatRequest, ai: ToolCall
     tools_for_investigation = issue_chat_request.investigation_result.tools
 
     if not conversation_history or len(conversation_history) == 0:
+        if global_instructions and global_instructions.instructions and len(global_instructions.instructions[0]) > 0:
```
moshemorad: This code block is repeated in multiple places, please extract and reuse.

holmes/core/conversations.py
```diff
@@ -249,6 +255,9 @@ def build_issue_chat_messages(issue_chat_request: IssueChatRequest, ai: ToolCall
             },
         ]
 
+    if global_instructions and global_instructions.instructions and len(global_instructions.instructions[0]) > 0:
```
moshemorad: Im not 100% precent sure since not familiar enough with the code but this flow is part of the conversation history right? So it mean that the history should already include the globa instructions no? why we are resending this?

moshemorad (changes requested):
Awesome job. Few comments, let me know if something isn't clear.

holmes/core/tool_calling_llm.py
itisallgood: The frontend inserts the runbooks to the db in this format, I decided not to change it and leave it as it is to avoid confusion

holmes/plugins/prompts/generic_investigation.jinja2
itisallgood: I would rather not move it cause it varies a bit from with system prompts and we might want to change them for more specific cases (e.g. in the workload prompt). So I would rather not move it to dedicated template for now

holmes/core/supabase_dal.py
itisallgood: got it, thanks

holmes/core/conversations.py
itisallgood: We have global instructions saved in the database and fetched every time because user might want to change it dynamically during the chats, and continue to chat after. The most effective method is to pass them through user_prompt every time, and define rules for using them in system_prompt

holmes/core/conversations.py
itisallgood: updated the PR

holmes/core/tool_calling_llm.py
```diff
@@ -53,6 +53,10 @@ class ResourceInstructionDocument(BaseModel):
     url: str
 
 
+class Intructions(BaseModel):
```
moshemorad: Typo

holmes/core/conversations.py
moshemorad: Not sure I understand, why we move them in some cases as user prompt and in other cases as system prompt? why it isn't always system prompt?

holmes/plugins/prompts/generic_investigation.jinja2
moshemorad: I think it is better to do it other way around. Let's extract and reuse. If there will be situation that something will need some change we can consider then how to separate them.

holmes/core/tool_calling_llm.py
itisallgood: Fixed

moshemorad (approved):

holmes/core/conversations.py
```diff
@@ -340,5 +353,4 @@ def build_chat_messages(
         ai, conversation_history_without_tools, number_of_tools
     )
     truncate_tool_messages(conversation_history, tool_size)
-
     return conversation_history
```
moshemorad, line 356: The conversation history is including now the global instructions this means that this what the user will see in the history chat?

PR #238, opened 2024-12-19
[MAIN-2708] Gemini support

Avi-Robusta:
Tested on

- Robusta
- bedrock/anthropic.claude-3-5-sonnet-20240620-v1:0
- openai/gpt-4o
- vertex_ai/gemini-2.0-flash-exp
- gemini/gemini-pro


holmes/core/tools.py
```diff
@@ -78,7 +78,7 @@ class Tool(ABC, BaseModel):
     )
     additional_instructions: Optional[str] = None
 
-    def get_openai_format(self):
+    def get_openai_format(self, model: Optional[str]):
```
itisallgood: @Avi-Robusta maybe it will make sense to make a separate method like `get_gemini_format` and do check for `if "gemini"` in `def call`?

holmes/core/tool_calling_llm.py
```diff
@@ -100,7 +100,7 @@ def call(
     ) -> LLMResult:
 
         tool_calls = []
-        tools = self.tool_executor.get_all_tools_openai_format()
+        tools = self.tool_executor.get_all_tools_openai_format(self.llm.model)
```
itisallgood: maybe it makes sense to write a more generalized method to check model and if model gemini use  `get_gemini_format`, otherwise use `get_openapi_format` because we might have much more cases like this one in the future ?

arikalon1 (approved):
LGTM

PR #240, opened 2024-12-21
ROB-717 stream

RoiGlinik:


holmes/core/tool_calling_llm.py
```diff
@@ -28,7 +29,7 @@
 from holmes.core.issue import Issue
 from holmes.core.runbooks import RunbookManager
 from holmes.core.tools import ToolExecutor
-
+from litellm import stream_chunk_builder
```
moshemorad: Not in use?

holmes/core/tool_calling_llm.py
```diff
@@ -357,6 +365,97 @@ def truncate_messages_to_fit_context(
...
+            if (total_tokens + maximum_output_token) > max_context_size:
+                logging.warning("Token limit exceeded. Truncating tool responses.")
+                messages = self.truncate_messages_to_fit_context(
+                    messages, max_context_size, maximum_output_token
+                )
+                perf_timing.measure("truncate_messages_to_fit_context")
+
+            logging.debug(f"sending messages={messages}\n\ntools={tools}")
+            try:
+                full_response = self.llm.completion(
```
moshemorad, line 408: We don't need to also check here the incorrect_tool_call like we do in the sync call?

holmes/core/tool_calling_llm.py
```diff
@@ -357,6 +365,97 @@ def truncate_messages_to_fit_context(
...
+            with concurrent.futures.ThreadPoolExecutor(max_workers=16) as executor:
+                futures = []
+                for t in tools_to_call:
+                    futures.append(executor.submit(self._invoke_tool, t))
+                    yield json.dumps({"type": "start_tool_calling", "details": {"tool_name": t.function.name, "id": t.id}})
+
+                for future in concurrent.futures.as_completed(futures):
+                    tool_call_result: ToolCallResult = future.result()
+                    tool_calls.append(tool_call_result)
+                    messages.append(tool_call_result.as_dict())
```
moshemorad: Please save as_dict output to var and reuse instead of calling it multiple times.

holmes/core/llm.py
```diff
@@ -5,6 +5,8 @@
 from litellm.types.utils import ModelResponse
 import sentry_sdk
 
+from pydantic.types import SecretStr
```
moshemorad: Unused?

holmes/core/tool_calling_llm.py
```diff
@@ -357,6 +365,97 @@ def truncate_messages_to_fit_context(
                 message["content"] = message["content"][:tool_size]
         return messages
 
+    def call_stream(
```
moshemorad, line 369: Please add tests.

holmes/core/llm.py
```diff
@@ -170,6 +173,7 @@ def completion(
         response_format: Optional[Union[dict, Type[BaseModel]]] = None,
         temperature: Optional[float] = None,
         drop_params: Optional[bool] = None,
+		stream: Optional[bool] = None
```
moshemorad: We are currently not using this at all right? can we please remove it then?

holmes/core/investigation.py
```diff
@@ -49,3 +57,92 @@ def investigate_issues(
         tool_calls=investigation.tool_calls or [],
         instructions=investigation.instructions,
     )
+
+
+def get_investigation_context(investigate_request: InvestigateRequest, dal: SupabaseDal, config: Config):
```
moshemorad: Please consider to move it into `IssueInvestigator` Looks like we are duplicatin a lot of the code twice and it is maintanance hell. every change in issue investigate we will need to remember to fix twice as long as we have both implementations.

holmes/core/tool_calling_llm.py
```diff
@@ -357,6 +365,97 @@ def truncate_messages_to_fit_context(
...
+        self,
+        system_prompt: str,
+        user_prompt: Optional[str] = None,
+        response_format: Optional[Union[dict, Type[BaseModel]]] = None,
+        sections: Optional[InputSectionsDataType] = None,
+    ):
+        messages = [
```
moshemorad, line 376: Since most of this code is duplicate, there isn't a way to refactor and add only streaming support to the original function? or refactor some parts of it and reuse? Maybe change the original implementation to also be generator, when request stream it yield all relevant info and when not only yields the result. WDYT?

holmes/core/investigation.py
RoiGlinik: We discussed this, its a tech debt and ill work on refactoring this function over the server and main so it will be more flexible

holmes/core/tool_calling_llm.py
RoiGlinik: https://github.com/BerriAI/litellm/issues/8241 bug is already merged on our correct version so ill leave it out in this function.

added a ticket to remove from the other function next sprint

holmes/core/tool_calling_llm.py
RoiGlinik: added as a technical debt for next sprint to arrange these function .

PR #241, opened 2024-12-23
When holmesgpt is missing kubernetes permissions, suggest fix to user

itisallgood:


holmes/plugins/prompts/_general_instructions.jinja2
```diff
@@ -29,10 +29,43 @@ If investigating Kubernetes problems:
 ** check the application aspects through the logs (kubectl_logs and kubectl_previous_logs) and other relevant tools
 ** look for misconfigured ingresses/services etc
 
+Handling Permission Errors
+If during the investigation you encounter a permissions error (e.g., `Error from server (Forbidden):`), **ALWAYS** follow these steps to ensure a thorough resolution:
+1. **Analyze the Error Message**
+  - Identify the missing resource, API group, and verbs from the error details.
+  - Never stop at reporting the error—proceed with an in-depth investigation.
```
nherment: Is there a missing line feed?

```
- Never stop at reporting the error
- proceed with an in-depth investigation.
```

nherment (commented):
This looks good @itisallgood. Given it's a prompt change I think there should be some llm evals to at least validate the nominal behaviour

nherment (approved):

PR #242, opened 2024-12-26
[Helm] Refactoring Kubernetes labels

nlamirault:
- support for [Kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)
- support for additional labels


aantn:
Hi, thanks for opening the PR! Can you share more details on the motivation for this change? Does it just add new labels or does it change any existing labels? (We want to make sure we're not breaking anything for existing users.)

PR #250, opened 2025-01-18
Save stack traces for holmes errors

itisallgood:
* Added FastAPI style exception handling for Holmes server
* Removed old exception handling from server


server.py
```diff
@@ -88,124 +89,119 @@ async def lifespan(app: FastAPI):
 console = Console()
 
 
+@app.exception_handler(Exception)
+async def general_exception_handler(request: Request, exc: Exception):
+    logging.exception(f"Error during request to {request.url.path} endpoint:")
+    return JSONResponse(
+        status_code=500,
+        content={"message": str(exc), "traceback": traceback.format_exc()},
```
nherment, line 97: Isn't sharing the exception with the client a concern? Genuinely asking. Typically this is only acceptable during development because bubbling up exception stack traces can provide a lot of information about a system to an untrusted client.

In this case I think it MIGHT be the case but holmes has no authentication... Some info could still make it to the UI with users not necessarily allowed to see some details.

nherment (commented):
Looks good, just one comment on returning stack trace to the client.

PR #252, opened 2025-01-20
MAIN-2806: Enable-AI-Output-Customization

nherment:


holmes/plugins/toolsets/kubernetes_logs.yaml
```diff
@@ -35,4 +35,8 @@ toolsets:
 
       - name: "kubectl_logs_grep"
         description: "Search for a specific term in the logs of a single Kubernetes pod. Only provide a pod name, not a deployment or other resource."
-        command: "kubectl logs {{ name }} -n {{ namespace }} | grep {{ search_term }}"
+        command: "kubectl logs {{ pod_name }} -n {{ namespace }} | grep {{ search_term }}"
+
+      - name: "kubectl_logs_all_containers_grep"
```
moshemorad, line 40: Added new tool? Is that other requirement?

holmes/core/tool_calling_llm.py
```diff
@@ -160,13 +158,24 @@ def call(
             )
 
             tools_to_call = getattr(response_message, "tool_calls", None)
+            text_response = response_message.content
+            sections:Optional[Dict[str, str]] = None
+            if isinstance(text_response, str):
```
moshemorad, line 163: Can please explain the use case here?

moshemorad (approved):

holmes/plugins/toolsets/kubernetes_logs.yaml
nherment: No. When reading though and fixing evals I thought it made sense to add an equivalent grep capability to the tool that fetches logs for all containers.

holmes/core/tool_calling_llm.py
nherment: The LLM is only asked for structured output for investigations but not for ask holmes for example. The call() method is generic to all use cases and this code is to handle a structured output as a result. It then combines the structured output into the 'monolithic' text_response for backward compatibility reasons.

holmes/plugins/toolsets/kubernetes_logs.yaml
nherment: That new tool is also used by some evals. I 'think' it helps when a pod has multiple containers and an issue that the LLM needs to find is not in the default container.

PR #254, opened 2025-01-21
MAIN-2888: make browser more lightweight

nherment:
Revert browser to a lighter version.
Playwright was removed in favour of a simple HTTP GET. The existing tag cleanup with beautifulsoup was kept and improved with more tags. The tranformation step to markdown was unchanged.


holmes/plugins/toolsets/internet.py
```diff
@@ -1,79 +1,101 @@
...
-def scrape_with_playwright(url):
+import requests
 
-    with sync_playwright() as p:
-        try:
-            browser = p.firefox.launch()
-        except Exception as e:
-            logging.error(str(e))
-            return None, None
+# TODO: change and make it holmes
```
moshemorad, line 14: What is this TODO? i see it contains holmesgpt so not sure i fully understand what it means :)

moshemorad (approved):

holmes/plugins/toolsets/internet.py
nherment: This is a reminder to potentially change the user agent. Bots or automated systems should typically not use a browser user agent but something specific to them. I used a user agent b/c some websites don't return the html unless the user agent is a browser.

PR #262, opened 2025-01-30
fix: issue with sections handling messing workload health structured output

nherment:
The code to parse the sections for the custom RCA output was in the wrong place, causing it meddle with the structured output from the workload health checks.
This PR moves this code out of the tool_calling_llm.py into the investigation helper

### DIFF
--- a/holmes/core/investigation_structured_output.py
+++ b/holmes/core/investigation_structured_output.py
@@ -7,8 +7,7 @@ InputSectionsDataType = Dict[str, str]
 
 OutputSectionsDataType = Optional[Dict[str, Union[str, None]]]
 
-class SectionsData(RootModel):
-    root: OutputSectionsDataType
+SectionsData = RootModel[OutputSectionsDataType]
 
 DEFAULT_SECTIONS:InputSectionsDataType = {
     "Alert Explanation": "1-2 sentences explaining the alert itself - note don't say \"The alert indicates a warning event related to a Kubernetes pod doing blah\" rather just say \"The pod XYZ did blah\" because that is what the user actually cares about",
--- a/holmes/core/investigation_structured_output.py
+++ b/holmes/core/investigation_structured_output.py
@@ -48,7 +48,6 @@ def combine_sections(sections: Dict) -> str:
     content = ''
     for section_title, section_content in sections.items():
         if section_content:
-            # content = content + f'\n# {" ".join(section_title.split("_")).title()}\n{section_content}'
             content = content + f'\n# {section_title}\n{section_content}\n'
     return content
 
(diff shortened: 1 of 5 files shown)


holmes/core/tool_calling_llm.py
```diff
@@ -159,22 +156,12 @@ def call(
...
-            sections:Optional[Dict[str, str]] = None
-            if isinstance(text_response, str):
-                try:
-                    parsed_json = json.loads(text_response)
-                    text_response = parsed_json
-                except json.JSONDecodeError:
-                    pass
-            if not isinstance(text_response, str):
-                sections = text_response
-                text_response = combine_sections(sections)
```
nherment, line 171: This is the code responsible for messing things up

holmes/core/investigation_structured_output.py
```diff
@@ -34,12 +44,32 @@ def get_output_format_for_investigation(sections: Dict[str, str]) -> Dict[str, A
...
+            content = content + f'\n# {section_title}\n{section_content}\n'
+    return content
+
+
+def process_response_into_sections(response: Any) -> Tuple[str, OutputSectionsDataType]:
+    if isinstance(response, dict):
+        # No matter if the result is already structured, we want to go through the code below to validate the JSON
+        response = json.dumps(response)
+
+    if not isinstance(response, str):
```
moshemorad, line 59: Trying to understand, if it isn't str and not dict what it could be?

holmes/core/investigation_structured_output.py
```diff
@@ -1,6 +1,16 @@
...
+import json
 
-DEFAULT_SECTIONS = {
+from pydantic import  RootModel
+
+InputSectionsDataType = Dict[str, str]
+
+OutputSectionsDataType = Optional[Dict[str, Union[str, None]]]
+
+class SectionsData(RootModel):
```
moshemorad: Consider to use RootModel instead of having model with root, https://docs.pydantic.dev/2.10/concepts/models/#rootmodel-and-custom-root-types

WDYT?

holmes/core/investigation_structured_output.py
nherment: Could be a list, a boolean or a number. I don't think it'll happen and I don't think this is actually a good way of handling this scenario. However it's put there purely for defensive purposes.

holmes/core/investigation_structured_output.py
nherment: :heavy_check_mark:

holmes/core/investigation_structured_output.py
itisallgood: @nherment is this code resilient to errors and unexpected responses? did you ever get responses that produced exceptions?

moshemorad (approved):

PR #263, opened 2025-01-30
fix: dockerfile should download appropriate image for arm64

nherment:
Fix for this build issue: https://github.com/robusta-dev/holmesgpt/actions/runs/13054902903


arikalon1 (approved):
LGTM

PR #265, opened 2025-02-03
Structured output improvements

nherment:


holmes/core/investigation_structured_output.py
```diff
@@ -43,7 +44,7 @@ def get_output_format_for_investigation(
         "json_schema": {
             "name": "InvestigationResult",
             "schema": schema,
-            "strict": False,
+            "strict": True,
```
arikalon1: is there a downside of using `strict: True` ?
Is there a chance it will fail on some scenario? old LLMs maybe?

holmes/core/investigation_structured_output.py
```diff
@@ -67,6 +68,17 @@ def process_response_into_sections(response: Any) -> Tuple[str, OutputSectionsDa
...
         response = str(response)
 
+    if response.startswith("```json\n") and response.endswith("\n```"):
+        try:
+            parsed = json.loads(
+                response[8:-3]
+            )  # if this parses as json, set the response as that.
+            if isinstance(parsed, dict):
+                logging.warning("LLM did not return structured data")
+                response = response[8:-3]
```
arikalon1: do we want to set the response to the `parsed` or to the `response`?
Also, I think we might want to return, because in this case the json parsing on line 83 will always fail

arikalon1 (approved):
nice work
left a couple of comments

holmes/core/investigation_structured_output.py
nherment: This code is removing the leading ` ```json ` and trailing ` ``` ` from a string.
This then gets parsed again further below (current line is 156). I'm doing this double parsing because I think it "simplifies" reading and understanding it but I could be wrong. Happy to change if it's confusing.

holmes/core/tool_calling_llm.py
```diff
@@ -150,7 +151,17 @@ def call(
                 else:
                     raise
             response = full_response.choices[0]
+
             response_message = response.message
+            if i == 0 and response_message and not DISABLE_SYNTHETIC_STRUCTURED_OUTPUT:
```
nherment: This is the 'hack' to make structured output work around the [litellm 'bug'](https://github.com/BerriAI/litellm/issues/8241#issuecomment-2635720470).
It tries to see if the answer on the first llm call is a 'disguised' tool call and if so disables explicit structured output by setting `response_format = None`.

After that, it is expected for the LLM to return markdown which will then be parsed and split into sections.

holmes/core/investigation_structured_output.py
nherment: I set it back to False

holmes/core/investigation_structured_output.py
```diff
@@ -1,22 +1,29 @@
...
+import logging
+from typing import Any, Dict, Optional, Tuple
 import json
+import re
+import os
 
-from pydantic import RootModel
 
-InputSectionsDataType = Dict[str, str]
+REQUEST_STRUCTURED_OUTPUT_FROM_LLM = os.environ.get(
```
arikalon1: we have a method for loading booleans: `load_bool` 
might be better to use it, to be consistent

holmes/core/investigation_structured_output.py
```diff
@@ -58,7 +64,101 @@ def combine_sections(sections: Dict) -> str:
...
+    # Header Title
+    Content here
+    ```
+    =>
+    {
+      "Header Title": "Content here"
+    }
+    """
+    # Split the text into sections based on headers (# Section)
+    matches = re.split(r"\n(?=# )", markdown_content.strip())
```
moshemorad, line 115: Just raising edge case here, what if the response won't start with new line? this could happen?

holmes/core/investigation_structured_output.py
```diff
@@ -58,7 +64,101 @@ def combine_sections(sections: Dict) -> str:
...
+
+    if not matches[0].startswith("#"):
+        matches = matches[1:]
+    # Create a dictionary to store the sections
+    sections = {}
+
+    for match in matches:
+        if match.strip():
```
moshemorad: Please extract `match.strip()` to var and reuse, you also use it in `match.strip().split("\n", 1)`

holmes/core/investigation_structured_output.py
```diff
@@ -67,14 +167,105 @@ def process_response_into_sections(response: Any) -> Tuple[str, OutputSectionsDa
...
 
+    # In some cases, the LLM will not return a structured json but instead embed the JSON into a markdown code block
+    # This is not ideal and actually should not happen
+    if response.startswith("```json\n") and response.endswith("\n```"):
+        response = extract_within(response, 8, -3)
+
+    if response.startswith('"{') and response.endswith('}"'):
+        # Some Anthropic models embed the actual JSON dict inside a JSON string
+        # In that case it gets parsed once to get rid of the first level of marshalling
+        try:
```
moshemorad: If you want just to ignore it you can use: https://docs.python.org/3.5/library/contextlib.html#contextlib.suppress

holmes/core/investigation_structured_output.py
```diff
@@ -58,7 +64,101 @@ def combine_sections(sections: Dict) -> str:
...
+    if len(sections) > 0:
+        return sections
+    else:
+        return None
+
+
+def extract_within(content: str, from_idx: int, to_idx: int) -> str:
+    try:
+        parsed = json.loads(
+            content[from_idx:to_idx]
```
moshemorad: Please do the slicing once `content[from_idx:to_idx]` and reuse.

holmes/core/investigation_structured_output.py
```diff
@@ -67,14 +167,105 @@ def process_response_into_sections(response: Any) -> Tuple[str, OutputSectionsDa
...
+                content = json.loads(content)
+            if not isinstance(content, dict):
+                return False
+            for section_title in sections:
+                if section_title in content:
+                    return False
+            return True
+
+    except Exception:
+        pass
```
moshemorad: Why need the pass? just return False in the except clause no?

holmes/core/tool_calling_llm.py
```diff
@@ -150,7 +160,24 @@ def call(
...
+                    sections, dict_response.get("choices", [{}])[0]
+                )
+
+                if incorrect_tool_call:
+                    logging.warning(
+                        "Detected incorrect tool call. Structured output will be disabled. This can happen on models that do not support tool calling. For Azure AI, make sure the model name contains 'gpt-4o'. To disable this holmes behaviour, set REQUEST_STRUCTURED_OUTPUT_FROM_LLM to `false`."
+                    )
+                    # disable structured output going forward and and retry
+                    response_format = None
+                    max_steps = max_steps + 1
```
moshemorad, line 180: Changing max_steps after range was initialized with it won't add another step to the for loop. If im understadning correctly this is what you tried to achive here right?

tests/llm/fixtures/test_investigate/01_oom_kill/test_case.yaml
```diff
@@ -5,4 +5,4 @@ retrieval_context:
   - There are 5 pods in running state
   - 7 pods are not running as indicated by the STATUS column
 evaluation:
-  correctness: 1
+  correctness: 0
```
moshemorad, line 8: Why?

tests/llm/test_investigate.py
```diff
@@ -138,10 +139,14 @@ def test_investigate(experiment_name, test_case):
     print(f"\n** OUTPUT **\n{output}")
     print(f"\n** SCORES **\n{scores}")
 
-    assert result.sections
-    assert len(result.sections) >= len(DEFAULT_SECTIONS)
+    assert result.sections, "Missing sections"
+    assert (
+        len(result.sections) >= len(DEFAULT_SECTIONS)
+    ), f"Received {len(result.sections)} sections but expected {len(DEFAULT_SECTIONS)}. Received: {result.sections.keys()}"
```
moshemorad, line 145: Logging receivted len twice.

holmes/core/investigation_structured_output.py
nherment: In that case, matches[0] will have the first section. Added a test for it so that it does not get broken during a future code update.

holmes/core/investigation_structured_output.py
nherment: Thanks, using this across this file

holmes/core/investigation_structured_output.py
nherment: The pass here is to reduce the number of exit points for the function. This approach can make functions easier to read. However, using `with suppress()` now so it's no longer relevant.

holmes/core/tool_calling_llm.py
nherment: :facepalm: changed to a while loop. Thank you.

tests/llm/fixtures/test_investigate/01_oom_kill/test_case.yaml
nherment: It was left to `1` by mistake. Our system never gets this one right

tests/llm/test_investigate.py
nherment: Do you mean I'm logging the `len()` twice? I don't think so but maybe I'm misunderstanding.

moshemorad: You are right, my bad.

moshemorad (approved):

PR #269, opened 2025-02-06
Added except Exception clause for endpoints

itisallgood:


nherment (approved):
Looks good. This could also be done through a middleware for all APIs but this is clean and simple.

PR #275, opened 2025-02-11
Add sentry

moshemorad:


server.py
```diff
@@ -109,6 +122,11 @@ async def log_requests(request: Request, call_next):
             )
 
 
+@app.get("/sentry-debug")
```
itisallgood: Forgot to remove endpoint for testing

server.py
```diff
@@ -86,6 +90,15 @@ async def lifespan(app: FastAPI):
     yield
 
 
+if dal.enabled and ENABLE_TELEMETRY and SENTRY_DSN:
```
itisallgood: is it possible to have a server.py running without dal.enabled?

server.py
moshemorad: It is there so we will also be able to test after changes in future.

itisallgood (dismissed):

itisallgood (approved):

aantn (approved):

PR #277, opened 2025-02-11
Fixed loading of the toolsets for holmes cli

itisallgood:


holmes/config.py
```diff
@@ -185,13 +185,18 @@ def create_console_tool_executor(
...
-        # Enable all matching toolsets that have CORE or CLI tag
-        for toolset in matching_toolsets:
-            toolset.enabled = True
+        toolsets_loaded_from_config = self.load_custom_toolsets_config()
+        loaded_toolsets_from_env = None
+        if self.toolsets:
+            loaded_toolsets_from_env = self.load_toolsets_config(self.toolsets, "env")
+
+        # Enable all matching toolsets if no custom configs are provided!
+        if not toolsets_loaded_from_config and not loaded_toolsets_from_env:
```
moshemorad: Please revert the enable of default/builtin tools according to the allowed_toolsets.

holmes/config.py
```diff
@@ -431,6 +434,33 @@ def load_custom_toolsets_config(self) -> list[ToolsetYamlFromConfig]:
                 description: "Perform a curl request to example.com using variables"
                 command: "curl -X GET '{{api_endpoint}}?query={{ query_param }}' "
         """
+        loaded_toolsets = []
+        for custom_path in self.custom_toolsets:
+            if os.path.isfile(custom_path):
```
moshemorad: Please extract it to method and reuse also in loading of CUSTOM_TOOLSET_LOCATION

holmes/config.py
```diff
@@ -170,25 +170,30 @@ def create_console_tool_executor(
...
+        2. Custom toolsets from config files which can override built-in toolsets
+        3. Toolsets defined in self.toolsets which can override both built-in and custom toolsets config
         """
         default_toolsets = [
             toolset
             for toolset in load_builtin_toolsets(dal)
             if any(tag in (ToolsetTag.CORE, ToolsetTag.CLI) for tag in toolset.tags)
         ]
 
+        for toolset in default_toolsets:
```
moshemorad, line 188: this should be after the matching toolsets as it was before and only enable matching toolsets.

holmes/config.py
```diff
@@ -431,16 +436,72 @@ def load_custom_toolsets_config(self) -> list[ToolsetYamlFromConfig]:
                 description: "Perform a curl request to example.com using variables"
                 command: "curl -X GET '{{api_endpoint}}?query={{ query_param }}' "
         """
+        loaded_toolsets = []
+        for custom_path in self.custom_toolsets:
+            if os.path.isfile(custom_path):
```
moshemorad: Please add log warning if couldn't find the file.

holmes/config.py
```diff
@@ -431,16 +436,72 @@ def load_custom_toolsets_config(self) -> list[ToolsetYamlFromConfig]:
...
+        return self.parse_toolsets_file(CUSTOM_TOOLSET_LOCATION, raise_error=True)
+
+    def parse_toolsets_file(
+        self, path: str, raise_error: bool = True
+    ) -> Optional[ToolsetYamlFromConfig]:
+        try:
+            with open(path, "r", encoding="utf-8") as file:
+                parsed_yaml = yaml.safe_load(file)
+        except yaml.YAMLError as err:
+            logging.exception(f"Error parsing YAML from {path}: {err}")
```
moshemorad: Please change to warning, since from now on all logging.error will be sent to sentry and in that case it isn't something that we need to fix.

Same for all below.
WDYT?

moshemorad (changes requested):

holmes/config.py
itisallgood: Oh, alright, I will remember that in future. Thanks for pointing out

holmes/config.py
itisallgood: It doesn't make any difference as matching toolsets are basically default ones filtered by name

moshemorad: i guess you are talking on the comment above.
It does matter if the user doesn't enable all tools with '*' right?

itisallgood: yes, sure, sorry

holmes/config.py
itisallgood: It doesn't make any difference as matching toolsets are basically default ones filtered by name

holmes/config.py
itisallgood: Added warnings

holmes/config.py
moshemorad: It does matter, what will happen if the user won't set it to `*`?

itisallgood: it will just filter out by names and get the objects from default_toolsets list

PR #279, opened 2025-02-12
Latest fixes

moshemorad:


nherment (dismissed):

Dockerfile
```diff
@@ -55,6 +55,20 @@ RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
 RUN chmod 777 argocd
 RUN ./argocd --help
 
+# Set the architecture-specific aws-cli
```
itisallgood: Can u add as a comment explanation why we comment out here cli installation, but leave it out in docker.dev ? It will be helpful for others while working on Docker images

nherment:
Should we not just remove the aws code from the dockerfile instead of commenting it?

itisallgood (approved):

PR #295, opened 2025-02-20
ROB-699: prometheus + graph generating capability

nherment:


holmes/utils/cache.py
```diff
@@ -0,0 +1,84 @@
...
+    try:
+        decompressed = bz2.decompress(compressed_data)
+        json_str = decompressed.decode("utf-8")
+        data = json.loads(json_str)
+        return data
+    except Exception as e:
+        raise Exception(f"Decompression failed: {str(e)}")
+
+
+class TTLCache:
```
arikalon1, line 33: why are we implementing our own ttl cache, and not using some 3rd party package?

nherment:
Related docs PR: https://github.com/robusta-dev/robusta/pull/1733

holmes/plugins/toolsets/prometheus.py
```diff
@@ -0,0 +1,456 @@
...
+    cache: Optional[TTLCache],
+    metrics_labels_time_window_hrs: Union[int, None],
+    metric_name: str,
+) -> dict:
+    """This is a slow query. Takes 5+ seconds to run"""
+    cache_key = f"metrics_labels:{metric_name}"
+    if cache:
+        cached_result = cache.get(cache_key)
+        if cached_result:
+            logging.info("fetch_metrics_labels() result retrieved from cache")
```
arikalon1, line 88: is this really info?

holmes/plugins/toolsets/prometheus.py
```diff
@@ -0,0 +1,456 @@
...
+            return cached_result
+
+    series_url = urljoin(prometheus_url, "/api/v1/series")
+    params: dict = {
+        "match[]": f'{{__name__=~".*{metric_name}.*"}}',
+    }
+    # params: dict = {
```
arikalon1: can we remove this?

holmes/plugins/toolsets/prometheus.py
```diff
@@ -0,0 +1,456 @@
...
+        if not self._cache and self.toolset.config.metrics_labels_cache_duration_hrs:
+            self._cache = TTLCache(
+                self.toolset.config.metrics_labels_cache_duration_hrs * 3600
+            )
+        try:
+            prometheus_url = self.toolset.config.prometheus_url
+            metrics_labels_time_window_hrs = (
+                self.toolset.config.metrics_labels_time_window_hrs
+            )
+            if not prometheus_url:
```
arikalon1: already checked on 179

arikalon1 (dismissed):
nice work!

left a few minor comments

holmes/plugins/toolsets/prometheus.py
nherment: I would think so. What I'm thinking is that we do log the tool calls and such message could help users quickly understand if Holmes is missing some new metrics that were added recently.
I can move it to debug if you prefer.

holmes/plugins/toolsets/prometheus.py
nherment: yes good catch. thank youi

holmes/utils/cache.py
nherment: Because it's low maintenance to roll our own (as long as it stays simple) and lowers attack surface for supply chain attacks or CVEs related maintenance.
Happy to pull in a library if you prefer.

arikalon1: got it, thanks

holmes/plugins/toolsets/prometheus.py
arikalon1: got it, thanks

arikalon1 (approved):
nice work

PR #300, opened 2025-02-26
log every tool call - even python tools

aantn:
We're currently not logging python tool calls, which makes it difficult in the CLI to follow which tools were called. This fix is a little ugly, so feel free to take an alternative approach if you prefer.

I've also removed a typo from one of the log lines.

### DIFF
--- a/holmes/core/tools.py
+++ b/holmes/core/tools.py
@@ -111,6 +111,9 @@ class Tool(ABC, BaseModel):
 
         return result
 
+    def log(self, params: Dict):
+        logging.info(f"Running tool: {self.__class__.__name__} with params {params}")
+
     @abstractmethod
     def invoke(self, params: Dict) -> str:
         return ""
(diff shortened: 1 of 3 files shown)


holmes/core/tools.py
```diff
@@ -111,6 +111,9 @@ def get_openai_format(self):
 
         return result
 
+    def log(self, params: Dict):
+        logging.info(f"Running tool: {self.__class__.__name__} with params {params}")
```
nherment: I think we should at least log the name of the tool which is often different from the class name.
I will care for updating this PR. The plan is to have something similar to:

```python
class Tool(ABC):
    def invoke(self, params):
        logging.info(f"Running tool: {self.get_parameterized_one_liner(context)}")
        return self._invoke(params)

    @abstractmethod
    def _invoke(self, params):
        pass
```

aantn: Sounds good, thanks

nherment (approved):

PR #308, opened 2025-03-04
ROB-708: use loki as full fledged log provider

nherment:


holmes/core/conversations.py
```diff
@@ -200,6 +200,8 @@ def build_issue_chat_messages(
     investigation_analysis = issue_chat_request.investigation_result.result
     tools_for_investigation = issue_chat_request.investigation_result.tools
 
+    enabled_toolsets = [ts.name for ts in ai.tool_executor.enabled_toolsets]
```
moshemorad: consider to move it as property/function to the tool_executor instead of recreate this list everytime.

Also instead of list please change to set since we are using in operator on it, that way it won't need to scan the list.

holmes/plugins/prompts/_general_instructions.jinja2
```diff
@@ -18,18 +18,17 @@ If investigating Kubernetes problems:
...
 * when investigating a pod that crashed or application errors, always run kubectl_describe and fetch the logs
-* if a pod has multiple containers, make sure you fetch the logs for either all or relevant containers using one of the containers log functions like kubectl_logs_all_containers, kubectl_logs_all_containers_grep or any other.
 * do not give an answer like "The pod is pending" as that doesn't state why the pod is pending and how to fix it.
 * do not give an answer like "Pod's node affinity/selector doesn't match any available nodes" because that doesn't include data on WHICH label doesn't match
 * if investigating an issue on many pods, there is no need to check more than 3 individual pods in the same deployment. pick up to a representative 3 from each deployment if relevant
 * if the user says something isn't working, ALWAYS:
 ** use kubectl_describe on the owner workload + individual pods and look for any transient issues they might have been referring to
-** check the application aspects through the logs (kubectl_logs and kubectl_previous_logs) and other relevant tools
+** check the application aspects through the logs
```
moshemorad, line 27: Shouldn't we also move this line? since if none of the logs tools enabled it will only confuse the llm no?

holmes/plugins/toolsets/datetime.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def _invoke(self, params: Dict) -> str:
+        now = datetime.datetime.now(datetime.timezone.utc)
+        return f"The current UTC date and time are {now}. The current UTC timestamp in seconds is {int(now.timestamp())}."
+
+    def get_parameterized_one_liner(self, params) -> str:
+        return "fetched current time"
+
+
+class DatetimeToolset(Toolset):
```
moshemorad, line 23: Thinking here loud.
A. Do we want to make sure that this toolset always enabled and presented? if so let's open a tech debt ticket to make some tools immutalbe.
B. Are we fine to show it in the toolsets screen in the platform?

holmes/plugins/toolsets/grafana/common.py
```diff
@@ -19,20 +22,96 @@ def headers(api_key: str):
     }
 
 
+def is_int(string):
```
moshemorad, line 25: Consider using https://dateparser.readthedocs.io/en/latest/ I think it will support those cases.

holmes/plugins/toolsets/grafana/common.py
```diff
@@ -19,20 +22,96 @@ def headers(api_key: str):
...
         end_timestamp = int(time.time())
+
+    # If no start_timestamp provided, default to one hour before end
     if not start_timestamp:
-        start_timestamp = end_timestamp - ONE_HOUR_IN_SECONDS
-    if start_timestamp < 0:
-        start_timestamp = end_timestamp + start_timestamp
+        start_timestamp = -ONE_HOUR_IN_SECONDS
+
+    if start_timestamp and is_int(start_timestamp):
```
moshemorad: Please refactor the whole if else to a function and reuse.
def try_parse_datetime()

holmes/plugins/toolsets/grafana/common.py
```diff
@@ -19,20 +22,96 @@ def headers(api_key: str):
...
+        elif start_timestamp < 0:
+            start_timestamp = end_timestamp + start_timestamp
+        elif end_timestamp < 0:
+            # start/end are inverted. end_timestamp should be after start_timestamp
+            delta = end_timestamp
+            end_timestamp = start_timestamp
+            start_timestamp = start_timestamp + delta
+
+    # Invert timestamps if start is after end
+    if (
```
moshemorad, line 101: :/

holmes/plugins/toolsets/grafana/loki_api.py
```diff
@@ -72,73 +57,26 @@ def execute_loki_query(
...
-    pod_regex: str,
+    label_value: str,
+    filter_regexp: Optional[str],
     start: int,
     end: int,
-    pod_name_search_key: str = "pod",
+    label: str,
     namespace_search_key: str = "namespace",
-    limit: int = 1000,
+    limit: int = 200,
```
moshemorad, line 71: maybe make it configurable thorugh the toolset config?

tests/llm/test_ask_holmes.py
```diff
@@ -163,4 +164,5 @@ def ask_holmes(test_case: AskHolmesTestCase) -> LLMResult:
     chat_request = ChatRequest(ask=test_case.user_prompt)
 
     messages = build_chat_messages(chat_request.ask, [], ai=ai)
+    print(messages[0]["content"])
```
moshemorad: Left in purpose?

holmes/plugins/toolsets/datetime.py
nherment: I would not add an override for such tools b/c I don't see the value. Same for showing it. If it's a tool, it should be shown like the other tools.

holmes/plugins/toolsets/grafana/common.py
nherment: What would be the value? I checked and yes it would support these use cases but I don't see the value. We explicitly ask the LLM to provide a date in the proper format which it is doing pretty well

holmes/plugins/toolsets/grafana/loki_api.py
nherment: This is a lower API that is not used. I'd suggest keeping the LLM in control of the limit. It can already hone-in on the logs related to an issue thanks to the time range.

tests/llm/test_ask_holmes.py
nherment: Yes, removed. Thank you

holmes/plugins/prompts/_general_instructions.jinja2
nherment: No, this should stay. It tells the LLM to check the logs when investigating kubernetes issues and not assume everything is fine just because pods are healthy

holmes/plugins/toolsets/grafana/common.py
moshemorad: Up to you :)

moshemorad (approved):

PR #314, opened 2025-03-06
Multiple opensearch cluster suport

Avi-Robusta:
- Holmes would sometimes halucinate cluster names for multi configured opensearch clusters.
- The logic for get_client was broken for more than 1 cluster (would check var that wasnt there)
- misc other bug fixes

Screen Shot 2025-03-06 at 9 59 38


arikalon1 (approved):
LGTM

PR #315, opened 2025-03-06
[ROB-792 ] added node selector

Avi-Robusta:
verified via running helm chart locally 
Screen Shot 2025-03-06 at 13 32 21


arikalon1 (approved):
LGTM

PR #317, opened 2025-03-06
Upgrade ubuntu runner version

moshemorad:


.github/workflows/build-binaries-and-brew.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
         # We build on an older Ubuntu as pyinstaller binaries are forward-compatible not backwards-compatible
         # See https://pyinstaller.org/en/stable/usage.html?highlight=glibc#making-gnu-linux-apps-forward-compatible:~:text=The%20solution%20is%20to%20always%20build%20your%20app%20on%20the%20oldest%20version%20of%20GNU/Linux%20you%20mean%20to%20support.%20It%20should%20continue%20to%20work%20with%20the%20libc%20found%20on%20newer%20versions.
         # TODO: for similar reasons, we may want to build on older Windows/MacOS versions as well
-        os: [ubuntu-20.04, windows-latest, macos-latest]
+        os: [ubuntu-22.04, windows-latest, macos-latest]
```
moshemorad, line 14: @arikalon1 are we fine with having it only for 22.04? Looks good to me but WDYT?

arikalon1: I think it's fine

PR #319, opened 2025-03-07
fix: add logs prompt to workload health check + log enabled toolsets

nherment:


holmes/plugins/prompts/kubernetes_workload_ask.jinja2
```diff
@@ -46,6 +46,8 @@ In general:
 ** check the application aspects with kubectl_logs + kubectl_previous_logs and other relevant tools
 ** look for misconfigured ingresses/services etc
 
+{% include '_fetch_logs.jinja2' %}
```
itisallgood, line 49: I noticed that in `_fetch_logs.jinja2` we have the following text: 
```
* You have not been given access to tools to fetch kubernetes logs for nodes, pods, services or apps. This is likely a misconfiguration.
* If you need logs to answer questions or investigate issues, tell the user to configure the documentation and enable one of these toolsets:
** 'kubernetes/logs'
** 'grafana/loki'
```

Are we sure that Holmes actually does this and doesn't come up with error or something ?

itisallgood (approved):

holmes/plugins/prompts/kubernetes_workload_ask.jinja2
nherment: I see, thank for the question. Honestly I have not tested it and I'm not sure it matters so much. Having no logging tool configured would seriously limit Holmes in any case and this prompt here can't be worse than what we had before which assumed `kubectl logs ` was always present.

PR #330, opened 2025-03-16
ROB-858 holmes telemetry

moshemorad:


holmes/clients/robusta_client.py
```diff
@@ -0,0 +1,24 @@
+import logging
+from typing import Optional
+import requests
+from pydantic import BaseModel, ConfigDict
+from holmes.common.env_vars import ROBUSTA_API_ENDPOINT
+
+HOLMES_GET_INFO_URL = f"{ROBUSTA_API_ENDPOINT}/api/holmes/get_info"
+TIMEOUT = 0.5
```
arikalon1: seems a little short - 0.5 sec might fail often

holmes/config.py
```diff
@@ -211,6 +213,32 @@ class Config(RobustaBaseConfig):
 
     _server_tool_executor: Optional[ToolExecutor] = None
 
+    version: Optional[str] = Field(
```
arikalon1: What is the reason this is in the config?
Is it something the user can configure?

arikalon1 (approved):
nice work

I think i would move the version data out of the config

arikalon1 (approved):

PR #333, opened 2025-03-18
Update prometheus.py

alexinthesky:
We need a relative path there, as some prometheus backend ( for example mimir ) are exposed this way: `http://mimir-nginx.mimir/prometheus/`
and with an absolute path ,the /prometheus is stripped off.


nherment (approved):
Thank you @alexinthesky

PR #335, opened 2025-03-19
ROB-717 format stream message

RoiGlinik:


RoiGlinik:
Added SSE format needed by FE.
Added instructions and done messages.
Return the buggy litellm response - incorrect tool call

moshemorad (approved):

nherment (approved):

PR #338, opened 2025-03-19
feat: improve fetch_url to work with github pages and make llm use of prometheus more resilient to errors

nherment:
- Improve reliability of LLM interactions with prometheus toolset by allowing LLM to set `end=0` for range queries
- Internet toolset can now correctly fetch content from github URLs. 
  - Improved fetch_url tests are easier to debug
- Always log basic performance metrics


holmes/plugins/toolsets/utils.py
```diff
@@ -100,9 +100,9 @@ def process_timestamps_to_int(
     # Handle negative timestamps (relative to the other timestamp)
     if isinstance(start, int) and isinstance(end, int):
```
itisallgood, line 101: so what does happen when start is string ? is there any handling for that ? It seems like a possible scenario from 
`process_timestamps_to_int` definition

nherment: If start is a string it'll be transformed to an int line 97. This happens whether it's a date time string or if it's an epoch timestamp as a string.

itisallgood (approved):
Looks good to me, but I would add unit tests for `process_timestamps_to_int` as I can see it is used by Kafka, and maybe it could be nice to check all possible formats and scenarios with error cases

PR #341, opened 2025-03-21
ROB-945: prometheus toolset improvements

nherment:


.github/PULL_REQUEST_TEMPLATE/toolset.md
```diff
@@ -0,0 +1,10 @@
+
+Checklist for the toolset:
+
+- [ ] Toolset has unit tests where relevamt
```
moshemorad:
```suggestion
- [ ] Toolset has unit tests where relevant
```

holmes/plugins/toolsets/internet/internet.py
```diff
@@ -256,5 +261,5 @@ def __init__(self):
             tags=[
                 ToolsetTag.CORE,
             ],
-            is_default=True,
+            is_default="True",
```
moshemorad: Why? not sure i understand...

holmes/plugins/toolsets/kafka.py
```diff
@@ -454,14 +457,16 @@ def prerequisites_callable(self, config: Dict[str, Any]) -> bool:
...
-                    logging.error(
+                    message = (
                         f"Failed to set up Kafka client for {cluster.name}: {str(e)}"
                     )
+                    logging.error(message)
+                    errors.append(message)
 
-            return len(self.clients) > 0
-        except Exception:
+            return len(self.clients) > 0, "\n".join(errors)
```
moshemorad, line 466: Honestly this is confusing for me, we return True with errors and in the caller we just ignore those errors? Is that intended behavior? What should happen on Holmes Data Sources in the UI?

moshemorad (changes requested):

holmes/plugins/toolsets/internet/internet.py
nherment: Making type checking green. I don't think it's right to have put that type in `is_default` in the first place but it's there for now.

holmes/plugins/toolsets/kafka.py
nherment: Your understanding is correct and you are right it's not the right thing to do. There is a ticket this sprint to go over this specific kafka health check code and make sure we bubble up the correct info to the UI

holmes/plugins/toolsets/internet/internet.py
moshemorad: I think that it is by mistake, the arg in the internet tools set was set to str instead of bool. Why we can't fix it?

nherment: I'll revert this.

nherment: actually not. There is an issue further up in the file. Probably some code that was never refactored.

moshemorad (approved):

moshemorad (approved):

PR #345, opened 2025-03-23
Opensearch logs and traces

arikalon1:


holmes/core/llm.py
```diff
@@ -186,6 +191,7 @@ def completion(
             temperature=temperature,
             response_format=response_format,
             drop_params=drop_params,
+            thinking=thinking,
```
nherment, line 194: I would suggest renaming `tool_args` to `extra_args` and then:

```
        if THINKING:  # if model requires 'thinking', load it from env vars
            extra_args["thinking"] = json.loads(THINKING)
```

This is because litellm sometimes treat present but `None` arguments differently than when arguments are not present. I doubt it'll be an issue in this case though but still suggesting more defensive code.

holmes/core/supabase_dal.py
```diff
@@ -219,6 +219,27 @@ def get_issue_data(self, issue_id: Optional[str]) -> Optional[Dict]:
...
+            issue_data["start_timestamp"] = start_timestamp.strftime(
+                "%Y-%m-%dT%H:%M:%S.%fZ"
+            )
+            issue_data["end_timestamp"] = end_timestamp.strftime(
+                "%Y-%m-%dT%H:%M:%S.%fZ"
+            )
+            issue_data["start_timestamp_millis"] = int(
+                start_timestamp.timestamp() * 1000
+            )
+            issue_data["end_timestamp_millis"] = int(end_timestamp.timestamp() * 1000)
```
nherment, line 241: :+1:

holmes/plugins/toolsets/opensearch/opensearch_logs.py
```diff
@@ -0,0 +1,185 @@
...
+        return f'search logs: query="{params.get("query")}"'
+
+
+class OpenSearchLogsToolset(Toolset):
+    def __init__(self):
+        super().__init__(
+            name="opensearch/logs",
+            description="OpenSearch integration to fetch logs",
+            docs_url="",
+            icon_url="",
```
nherment: Should `docs_url` and `icon_url` be added now or later? `docs_url` might look like `https://docs.robusta.dev/master/configuration/holmesgpt/toolsets/opensearch_logs.html`

holmes/plugins/toolsets/opensearch/opensearch_logs.py
```diff
@@ -0,0 +1,185 @@
...
+            self.config = OpenSearchLogsConfig(
+                opensearch_logs_url=os.environ.get("OPENSEARCH_LOGS_URL"),
+                opensearch_auth_header=os.environ.get(
+                    "OPENSEARCH_LOGS_AUTH_HEADER", None
+                ),
+            )
+            return True
+        else:
+            self.config = OpenSearchLogsConfig(**config)
+            return True
```
nherment: I think the toolsets also need to implement `get_example_config` or the dynamic config in the UI will not be correct. e.g.

```
    def get_example_config(self):
        example_config = GrafanaLokiConfig(
            api_key="YOUR API KEY",
            url="YOUR GRAFANA URL",
            grafana_datasource_uid="<UID of the loki datasource to use>",
        )
        return example_config.model_dump()
```

holmes/plugins/toolsets/opensearch/opensearch_traces.py
```diff
@@ -0,0 +1,198 @@
...
+
+class TracesSearchQuery(BaseOpenSearchTracesTool):
+    def __init__(self, toolset: "OpenSearchTracesToolset"):
+        super().__init__(
+            name="traces_in_range_search",
+            description="Get traces in a specified time range for an opensearch query",
+            parameters={
+                "query": ToolParameter(
+                    description="An OpenSearch search query. It should be a stringified json, matching opensearch search syntax. "
+                    "The query must contain a 'range' on the startTimeMillis field. From a given startTimeMillis, until a given startTimeMillis. This is time in milliseconds",
```
nherment, line 103: Does this work well? Should holmes be given a tool to transform dates to millisecs?

holmes/plugins/toolsets/opensearch/opensearch_logs.py
```diff
@@ -0,0 +1,185 @@
...
+            tools=[
+                GetLogFields(toolset=self),
+                LogsSearchQuery(toolset=self),
+            ],
+            tags=[
+                ToolsetTag.CORE,
+            ],
+        )
+
+    def prerequisites_callable(self, config: dict[str, Any]) -> bool:
```
nherment, line 171: This should also do a health check on opensearch otherwise the UI shows the toolset as enabled and working even if it is misconfigured. I realised that yesterday from Natan trying out prometheus and reposrting issues.

holmes/plugins/toolsets/opensearch/opensearch_traces.py
```diff
@@ -0,0 +1,198 @@
...
+            self.config = OpenSearchTracesConfig(
+                opensearch_traces_url=os.environ.get("OPENSEARCH_TRACES_URL"),
+                opensearch_auth_header=os.environ.get(
+                    "OPENSEARCH_LOGS_AUTH_HEADER", None
+                ),
+            )
+            return True
+        else:
+            self.config = OpenSearchTracesConfig(**config)
+            return True
```
nherment: same comments as the logs toolset re. docs and icon urls, `get_example_config` and live health check to test the connection.

holmes/utils/holmes_sync_toolsets.py
```diff
@@ -12,8 +12,14 @@
...
-    statuses = {toolset.name: toolset.get_status().value for toolset in toolsets}
-    logging.info(f"Enabled toolsets: {statuses}")
+    enabled_toolsets = [
+        toolset.name for toolset in toolsets if toolset.get_status().value == "enabled"
+    ]
+    disabled_toolsets = [
+        toolset.name for toolset in toolsets if toolset.get_status().value != "enabled"
+    ]
+    logging.info(f"Enabled toolsets: {enabled_toolsets}")
+    logging.info(f"Disabled toolsets: {disabled_toolsets}")
```
nherment, line 22: :100:

nherment (approved):
Looks good. Made some suggestions

holmes/plugins/toolsets/opensearch/opensearch_traces.py
arikalon1: it works well on alerts, because I added those fields to the context
I'm calculating it when preparing the investigation context

holmes/core/llm.py
arikalon1: Looks like this behaves like the default, so I think it's ok in this case

holmes/plugins/toolsets/opensearch/opensearch_logs.py
arikalon1: fixed

holmes/plugins/toolsets/opensearch/opensearch_logs.py
arikalon1: good point, fixed

holmes/plugins/toolsets/opensearch/opensearch_logs.py
arikalon1: fixed

holmes/plugins/toolsets/opensearch/opensearch_traces.py
arikalon1: fixed

nherment (approved):

PR #348, opened 2025-03-25
ROB-968: initial coralogix toolset

nherment:


holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
```diff
@@ -0,0 +1,170 @@
...
+                "endDate": end,
+            },
+        }
+
+        response = execute_query(
+            base_url=self.toolset.config.base_url,
+            api_key=self.toolset.config.api_key,
+            query=query,
+        )
+        print(f"Fetching coralogix logs {query}")
```
moshemorad: Remove print

holmes/plugins/toolsets/coralogix/utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class CoralogixLabelsConfig(BaseModel):
+    pod: str = "kubernetes.pod_name"
+    namespace: str = "kubernetes.namespace_name"
+    app: str = "kubernetes.labels.app"
+
+
+class CoralogixConfig(BaseModel):
+    base_url: str = "https://ng-api-http.eu2.coralogix.com"
```
moshemorad: Are we sure that we want it by default since it looks like they have few different endpoints. WDYT?

holmes/plugins/toolsets/coralogix/utils.py
```diff
@@ -0,0 +1,166 @@
...
+) -> str:
+    """Extracts timestamp and log values from parsed JSON objects, sorted in ascending order (oldest first)."""
+    logs = []
+
+    for data in json_objects:
+        if isinstance(data, dict) and "result" in data and "results" in data["result"]:
+            for entry in data["result"]["results"]:
+                try:
+                    user_data = json.loads(entry.get("userData", "{}"))
+                    kubernetes = user_data.get("kubernetes", None)
```
moshemorad: Are we sure that those are indeed the generic use case? what is Kubernetes in that case?
log_message will always be in "log" field because in their docs I see there is also `message` field. would we like to do make those configurable?

holmes/plugins/toolsets/coralogix/utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def extract_logs(
+    json_objects: List[Dict[str, Any]], add_namespace_tag: bool, add_pod_tag: bool
+) -> str:
+    """Extracts timestamp and log values from parsed JSON objects, sorted in ascending order (oldest first)."""
+    logs = []
+
+    for data in json_objects:
+        if isinstance(data, dict) and "result" in data and "results" in data["result"]:
```
moshemorad, line 143: Please refactor this, extract the loop inside logic to fun and use return early. It is very hard to keep up with this length of loop and ifs.

holmes/plugins/toolsets/coralogix/utils.py
```diff
@@ -0,0 +1,166 @@
...
+                except json.JSONDecodeError:
+                    logging.error(
+                        f"Failed to decode userData JSON: {entry.get('userData')}"
+                    )
+        else:
+            logging.error(f"{data}")
+
+    logs.sort(key=lambda x: x[0])
+
+    formatted_logs = []
```
moshemorad: Consider to extract to method.

holmes/plugins/toolsets/coralogix/utils.py
```diff
@@ -0,0 +1,166 @@
...
+                    user_data = json.loads(entry.get("userData", "{}"))
+                    kubernetes = user_data.get("kubernetes", None)
+                    timestamp = normalize_datetime(user_data.get("time"))
+                    log_message = user_data.get("log", "")
+                    tags = format_kubernetes_info(
+                        kubernetes, add_namespace_tag, add_pod_tag
+                    )
+                    if log_message:
+                        logs.append(
+                            (timestamp, log_message, tags)
```
moshemorad: Consider using NamedTuple

holmes/plugins/toolsets/coralogix/utils.py
```diff
@@ -0,0 +1,166 @@
...
+    tags = []
+    if kubernetes:
+        if add_pod_tag and kubernetes.get("pod_name"):
+            tags.append(f'pod_name="{kubernetes.get("pod_name")}"')
+        if add_namespace_tag and kubernetes.get("namespace_name"):
+            tags.append(f'namespace_name="{kubernetes.get("namespace_name")}"')
+    return " ".join(tags)
+
+
+""" improves human readability of logs by indenting logs that span multiple lines.
```
moshemorad: Should this be part of the doc string of method `indent_multiline_log_message`?

holmes/plugins/toolsets/coralogix/utils.py
nherment: You're right. It's going to be error prone to have a default set to one of the regions. will remove.

holmes/plugins/toolsets/coralogix/utils.py
nherment: I'm not sure tbh. This is why I made this code flexible and if the `kubernetes` field is not there or has the wrong data the code will still work.
This just adds tags to each log line for Holmes to understand what pod and namespace are the logs from. The value is theoretical and there in case Holmes casts too wide a net and get the logs from many pods.

holmes/plugins/toolsets/coralogix/utils.py
nherment: yes, done as part of addressing the previous comment

holmes/plugins/toolsets/coralogix/utils.py
nherment: Ah right. It's inside the func in python...

holmes/plugins/toolsets/coralogix/utils.py
```diff
@@ -80,54 +76,81 @@ def normalize_datetime(date_str: str) -> str:
         date = datetime.fromisoformat(date_str.rstrip("Z"))
         # Ensure the timestamp has the maximum resolution (nanoseconds)
         normalized_date_time = date.strftime("%Y-%m-%dT%H:%M:%S.%f")
+        logging.error(f"normalized datetime {date_str} => {normalized_date_time}")
```
moshemorad: By mistake?

moshemorad (approved):

holmes/plugins/toolsets/coralogix/utils.py
nherment: I am debugging compatibility issues with older python versions. I will remove these debug lines once it's fixed.

Issue #352, opened 2025-03-27
Missing grafana folder in Homebrew installation causing HolmesGPT to fail


sourabhbhandari:
**Description:**

I installed HolmesGPT locally using Homebrew (brew tap robusta-dev/homebrew-holmesgpt && brew install holmesgpt). However, when trying to use it with Ollama, I encountered the following error:

holmes ask "what pods are unhealthy in my cluster?" --model="ollama_chat/llama3.2"
Failed to fetch holmes info                                      robusta_client.py:23
╭──────────────────────── Traceback (most recent call last) ────────────────────────╮
│ in ask:291                                                                        │
│                                                                                   │
│ in create_console_toolcalling_llm:418                                             │
│                                                                                   │
│ in create_console_tool_executor:301                                               │
│                                                                                   │
│ in load_builtin_toolsets:81                                                       │
│                                                                                   │
│ in load_python_toolsets:58                                                        │
│                                                                                   │
│ in __init__:275                                                                   │
│                                                                                   │
│ in _load_llm_instructions:375                                                     │
│                                                                                   │
│ in load_and_render_prompt:32                                                      │
│                                                                                   │
│ in load_prompt:22                                                                 │
╰───────────────────────────────────────────────────────────────────────────────────╯
FileNotFoundError: [Errno 2] No such file or directory:
**'/opt/homebrew/Cellar/holmesgpt/0.10.4-alpha/libexec/_internal/holmes/plugins/toolsets/grafana/toolset_grafana_tempo.jinja2'**

Upon checking, I found that the grafana folder is missing from the installed path:

**/opt/homebrew/Cellar/holmesgpt/0.10.4-alpha/libexec/_internal/holmes/plugins/toolsets/**

This causes the CLI tool to fail.

Steps to Reproduce:

1. Install HolmesGPT via Homebrew:

`brew tap robusta-dev/homebrew-holmesgpt && brew install holmesgpt`

2. Try running a query using Ollama:

`holmes ask "what pods are unhealthy in my cluster?" --model="ollama_chat/llama3.2"`

Observe the FileNotFoundError due to the missing grafana folder.

----------------------
Environment:

OS: macOS (Apple Silicon)

Installation method: Homebrew

HolmesGPT version: 0.10.4-alpha

Ollama version: ollama_chat/llama3.2
----------------------

moshemorad:
Thanks @sourabhbhandari we are looking into it.

zrks:
same for me
```
[...]

PR #353, opened 2025-03-27
ROB-997 add missing toolset files to binary build

nherment:


.github/workflows/build-binaries-and-brew.yaml
```diff
@@ -67,7 +67,21 @@ jobs:
       # regarding the tiktoken part of the command, see https://github.com/openai/tiktoken/issues/80
       # regarding the litellm part of the command, see https://github.com/pyinstaller/pyinstaller/issues/8620#issuecomment-2186540504
       run: |
-        pyinstaller holmes.py --add-data 'holmes/plugins/runbooks/*:holmes/plugins/runbooks' --add-data 'holmes/plugins/prompts/*:holmes/plugins/prompts' --add-data 'holmes/plugins/toolsets/*:holmes/plugins/toolsets' --hidden-import=tiktoken_ext.openai_public --hidden-import=tiktoken_ext --hiddenimport litellm.llms.tokenizers --hiddenimport litellm.litellm_core_utils.tokenizers --collect-data litellm
+
+        pyinstaller holmes.py \
+          --add-data 'holmes/plugins/runbooks/*:holmes/plugins/runbooks' \
+          --add-data 'holmes/plugins/prompts/*:holmes/plugins/prompts' \
+          --add-data 'holmes/plugins/toolsets/*:holmes/plugins/toolsets' \
+          --add-data 'holmes/plugins/toolsets/coralogix*:holmes/plugins/toolsets/coralogix' \
```
moshemorad, line 75: Thinking here loud, isn't there a better way to copy all folders recursively. My concern is that we will forget to add the new entries here every time. 

If this isn't supported i think we might consider using another hierarchy. e.g. put all template files in one folder and make sure to copy that. WDYT?

nherment: I think there are a couple of 'good' and comprehensive fixes:
1. Dynamically generate `--add-data` args by parsing the sources
2. Implement automated smoke tests for the cli

(2) would be extremely beneficial as we'd get notified at PR time.

moshemorad: Lets open tech debt for that for next sprint.

moshemorad (approved):

nherment:
Fixes #352

PR #357, opened 2025-03-28
feat: support for custom labels to prometheus queries

nherment:


holmes/core/openai_formatting.py
```diff
@@ -45,7 +45,7 @@ def format_tool_to_open_ai_standard(
     }
 
     # gemini doesnt have parameters object if it is without params
-    if tool_properties is None:
+    if tool_properties is None or tool_properties == {}:
```
nherment, line 48: This is unrelated with prometheus but makes holmes work with Gemini.

arikalon1 (approved):
nice work

PR #358, opened 2025-03-30
ROB-206: Change history tools

arikalon1:


holmes/core/supabase_dal.py
```diff
@@ -182,6 +183,67 @@ def sign_in(self) -> str:
...
+            if change_content:
+                changes_data.append(
+                    {
+                        "change": change_content["data"],
+                        "evidence_id": change_content["id"],
+                        **change,
+                    }
+                )
+
+        logging.debug(
```
moshemorad, line 242: Consider not using f-string. Since this will reevaluate before the check if the log level is enabled and it is waste.

holmes/plugins/toolsets/robusta/robusta.py
```diff
@@ -59,6 +65,53 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+        try:
+            changes = self._fetch_change_history(params)
+            if changes:
+                return yaml.dump(changes)
+            else:
+                return f"Could not find changes for {params}"
+        except Exception as e:
+            logging.error(e)
+            msg = f"There was an internal error while fetching changes for {params}. {str(e)}"
+            logging.error(msg)
```
moshemorad: Please change to exception (Or add stack trace) + duplicate error logging (Also 2 line above.

holmes/core/llm.py
```diff
@@ -192,6 +192,7 @@ def completion(
         thinking = None
         if THINKING:  # if model requires 'thinking', load it from env vars
             thinking = json.loads(THINKING)
+            litellm.modify_params = True
```
moshemorad, line 195: Why do we need this? Tried to understand from the docs but not sure i was able to.

moshemorad (approved):

holmes/core/llm.py
arikalon1: This THINKING and this one are for Anthropic
It needs it, otherwise there's an error with some of the tools

holmes/plugins/toolsets/robusta/robusta.py
arikalon1: fixed

holmes/core/supabase_dal.py
arikalon1: good point, fixed

Avi-Robusta (approved):

PR #362, opened 2025-04-02
ROB-1026: improve argocd

nherment:
- Add 2 evals for argoCD
- Add llm instructions for holmes to properly use argocd
- Change how toolset specific llm instructions are loaded so that yaml toolsets benefit from jinja2 templating


nherment:
docs: https://github.com/robusta-dev/robusta/pull/1774

PR #365, opened 2025-04-07
ROB-975: slackbot actions v1

nherment:


holmes/core/conversations.py
```diff
@@ -343,6 +343,39 @@ def build_issue_chat_messages(
     return conversation_history
 
 
+def add_or_update_system_prompt(
```
itisallgood: Could you please elaborate in the docs the cases when updating system prompt is useful? It will help others if they would need to handle something similar. Thanks

PR #366, opened 2025-04-08
ROB-1022 coralogix toolset returns url

nherment:
Some improvements to coralogix:

- The log tool returns 
  - A URL to replicate the query on the coralogix UI
  - The dataprime query
- The log level param has been removed. I'm not sure Holmes would use it well today and other logs toolsets don't support it. We may introduce it later if we find that Holmes need to quickly narrow down on errors although that may be solved better with an additional grep-like filter as opposed to log-level.
- The default time window for logs is extended to 24hrs. This is to help Holmes get logs when investigating pod related issues and the pod has been deleted.

:warning: should be merged after https://github.com/robusta-dev/holmesgpt/pull/365 because this branch was forked off of #365.

### DIFF
--- a/server.py
+++ b/server.py
@@ -337,12 +337,6 @@ def chat(chat_request: ChatRequest):
                     prompt="List the relevant runbooks and links used. Write a short summary for each",
                     pre_action_notification_text="Looking up and summarizing runbooks and links...",
                 ),
-                FollowUpAction(
-                    id="incidents",
-                    action_label="Incidents",
-                    prompt="List the related incidents",
-                    pre_action_notification_text="Searching for related incidents...",
-                ),
             ]
 
         llm_call = ai.messages_call(messages=messages)


holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
```diff
@@ -73,84 +78,104 @@ def __init__(self, toolset: BaseCoralogixToolset):
...
+            query_filters.append(f"{config.labels.app}:{app_name}")
 
         query_string = " AND ".join(query_filters)
+        query_string = f"source logs | lucene '{query_string}' | limit {log_count}"
+        return query_string
 
+    def _invoke(self, params: Any) -> StructuredToolResult:
+        if not self.toolset.config:
+            return StructuredToolResult(
+                status="success",
```
moshemorad: The status is success if the toolset isn't configured?

holmes/core/tools.py
```diff
@@ -26,6 +26,16 @@
 ToolsetPattern = Union[Literal["*"], List[str]]
 
 
+class StructuredToolResult(BaseModel):
+    schema_version: str = "robusta:v1.0.0"
+    status: Union[Literal["success"], Literal["no_data"], Literal["error"]]
```
moshemorad: Please change to str enum.

holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
```diff
@@ -73,84 +78,104 @@ def __init__(self, toolset: BaseCoralogixToolset):
...
                 raw_logs=response.text.strip(),
                 add_namespace_tag=add_namespace_tag,
                 add_pod_tag=add_pod_tag,
             )
+            return StructuredToolResult(
+                status="success" if logs else "no_data",
+                error=None if logs else "The query returned no logs",
+                data=logs,
+                query=query_string,
+                url=build_coralogix_link_to_logs(
```
moshemorad, line 150: Just to make sure, it is safe to run build_coralogix_link_to_logs if there is no data?

moshemorad (changes requested):

holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
nherment: good catch.

holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
nherment: Yes, it's actually one of the main use cases for having a URL

holmes/core/tools.py
```diff
@@ -94,10 +124,11 @@ def invoke(self, params: Dict) -> str:
         logging.info(
             f"Running tool {self.name}: {self.get_parameterized_one_liner(sanitize_params(params))}"
         )
-        return self._invoke(params)
+        result = self._invoke(params)
+        return format_tool_output(result)
 
     @abstractmethod
-    def _invoke(self, params: Dict) -> str:
+    def _invoke(self, params: Dict) -> Union[str, StructuredToolResult]:
```
itisallgood, line 133: Maybe it makes sense for all the tools to produce some kind of `StructuredToolResult` to avoid ambiguity ?

nherment: I agree it would simplify some of the code but I would say this is out of the scope of this task as it was specifically split to reduce the scope and avoid having to implement the structured result for all toolsets.

moshemorad (approved):

PR #367, opened 2025-04-08
Now holmes returns tools errors in a separate field

itisallgood:


nherment (changes requested):
@itisallgood I don't think that this approach is flexible and future proof. I'd think returning more complex structured data would be more powerful.

For example: 
- [structured data class](https://github.com/robusta-dev/holmesgpt/pull/366/files#diff-e6db47037b68ecbeb36b06891c2bda1c93e8157c1be40986daf404518d801056R29-R36)
- [invoke method refactor](https://github.com/robusta-dev/holmesgpt/pull/366/files#diff-e6db47037b68ecbeb36b06891c2bda1c93e8157c1be40986daf404518d801056R128)

PR #369, opened 2025-04-11
ROB-49: rabbitmq toolset

nherment:


tests/llm/fixtures/test_ask_holmes/38_rabbitmq_split_head/helm/send_messages.py
```diff
@@ -0,0 +1,177 @@
...
+        "payload": payload_base64,
+        "payload_encoding": "base64"
+    }
+    response = requests.post(publish_url, auth=auth, headers=headers, json=publish_payload)
+    if not check_response(response, "5. Publish Message"):
+        sys.exit(1)
+    print("\n")
+
+
+    # # 6. Get (Consume) the Message
```
RoiGlinik, line 162: why is this commented?

holmes/plugins/toolsets/rabbitmq/api.py
```diff
@@ -0,0 +1,411 @@
...
+            verify=config.verify_certs,
+        )
+        response.raise_for_status()
+        return response.json()
+    except requests.exceptions.RequestException as e:
+        logging.error(f"Request failed for {method} {url}: {e}")
+        raise  # Re-raise after logging for upstream handling
+
+
+def node_data_to_node_info(node_data: Dict) -> NodeInfo:
```
RoiGlinik, line 123: I think pydantic can do that kind of logic. not entirely sure. not blocking the PR but worth checking

RoiGlinik (approved):

tests/llm/fixtures/test_ask_holmes/38_rabbitmq_split_head/helm/send_messages.py
nherment: I want the messages to stay in the queue to ensure it's not trivial to reconcile nodes post-partition

holmes/plugins/toolsets/rabbitmq/api.py
nherment: Not with the current `NodeInfo` struct but you make a good point on that code. I modified the struct so that it can directly receive node_data with `NodeInfo(**node_data)`.

RoiGlinik (approved):

PR #371, opened 2025-04-14
Fix brew build

Avi-Robusta:
It didn't add all toolset directories to the binary by default - only manually configured directories
the robusta tool directory was added and broke the binary since it was missing the files,
Now will automatically add new tool directories


arikalon1 (approved):
LGTM

PR #372, opened 2025-04-14
Holmes' tools output changes

itisallgood:


holmes/core/tools.py
```diff
@@ -242,9 +255,12 @@ def __execute_subprocess(self, cmd) -> str:
...
                 stdin=subprocess.DEVNULL,
             )
-            return f"stdout:\n{result.stdout}\nstderr:\n{result.stderr}"
+            output = result.stdout + result.stderr
+            return output.strip(), 0, None
         except subprocess.CalledProcessError as e:
-            return f"Command `{cmd}` failed with return code {e.returncode}\nstdout:\n{e.stdout}\nstderr:\n{e.stderr}"
+            output = e.stdout + e.stderr
+            error_message = f"Command failed with return code {e.returncode}. stdout: {e.stdout.strip()}. stderr: {e.stderr.strip()}"
+            return output.strip(), e.returncode, error_message
```
nherment: I would simplify this code and avoid processing the result by combining stdout/stderr or writing an error message. I would argue that formatting the error is the responsibility of the `_invoke(self, params) -> StructuredToolResult` function.

The code here would be:

```python
result = subprocess.run(
                cmd,
                shell=True,
                capture_output=True,
                text=True,
                check=False, # do not throw, we just return the error code
                stdin=subprocess.DEVNULL,
                stdout=subprocess.PIPE,
                stderr=subprocess.STDOUT,
            )
return output.strip(), result.returncode
```

Then _invoke():

```python
    def _invoke(self, params) -> StructuredToolResult:
        if self.command is not None:
            raw_output = self.__invoke_command(params)
            raw_output, return_code = self.__invoke_command(params)
        else:
            raw_output = self.__invoke_script(params)
            raw_output, return_code = self.__invoke_script(params)

        ... apply_additional_instructions ...

        return StructuredToolResult(
            status=ToolResultStatus.SUCCESS
            if return_code == 0
            else ToolResultStatus.ERROR,
            error=f"Command failed with error code {return_code}",
            data=output_with_instructions,
            params=params,
        )
```


The advantages to this suggestions are:
- Error handling for yaml tools is done in a single place
- The tool output is left as-is without splitting stdout and stderr. It's behaving the same as a terminal, also maintaining the linearity between logs regardless of their log levels (stdout/stderr)
- StructuredToolResult consistently puts the raw tool output in the `data` field and keeps the error as a high level description of the issue

holmes/core/tools.py
```diff
@@ -35,6 +35,7 @@ class StructuredToolResult(BaseModel):
     schema_version: str = "robusta:v1.0.0"
     status: ToolResultStatus
     error: Optional[str] = None
+    return_code: Optional[int] = None
```
nherment, line 38: I don't think this field is necessary:
- the `error` field will always be set when the return code is non zero. The error field can contain the error code. For example `Command failed with exit code {exit_code}`
- `return_code` is specific to command line tools and like you suggested I'd try to make it as generic as possible

holmes/core/tools.py
```diff
@@ -35,6 +35,7 @@ class StructuredToolResult(BaseModel):
...
     data: Optional[Any] = None
     url: Optional[str] = None
     query: Optional[str] = None
```
nherment: Should `query` be renamed and re-used for holding the command line used in the YAML toolsets?
It's not yet used on the front-end. I think the `query` should hold the parametrized command used by the tool. For some tools it can be e.g. promQL, for command line tools it can be the actual command run.

Because it's not used by the UI I think it's ok if you say 'this is out of scope' for this PR.

holmes/plugins/toolsets/grafana/toolset_grafana_loki.py
```diff
@@ -112,7 +113,12 @@ def _invoke(self, params: Dict) -> str:
             end=end,
             limit=int(params.get("limit", 5000)),
         )
-        return "\n".join([format_log(log) for log in logs])
+        return StructuredToolResult(
+            status=ToolResultStatus.SUCCESS,
+            data="\n".join([format_log(log) for log in logs]),
+            return_code=0,
```
nherment: `return_code` should not be set for any tool that does not return an error code. I actually don't think it should be part of `StructuredToolResult`.

holmes/plugins/toolsets/internet/internet.py
```diff
@@ -193,15 +195,25 @@ def _invoke(self, params: Any) -> str:
 
         if not content:
             logging.error(f"Failed to retrieve content from {url}")
-            return ""
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=f"Failed to retrieve content from {url}",
+                return_code=-1,
```
nherment: I think we should not invent status codes like this. is it required?

holmes/plugins/toolsets/internet/notion.py
```diff
@@ -51,9 +52,19 @@ def _invoke(self, params: Any) -> str:
 
         if not content:
             logging.error(f"Failed to retrieve content from {url}")
-            return ""
-
-        return self.parse_notion_content(content)
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=f"Failed to retrieve content from {url}",
+                return_code=-1,
```
nherment: ditto regarding return code `-1`

tests/llm/utils/mock_toolset.py
```diff
@@ -62,15 +63,24 @@ def _auto_generate_mock_file(self, params: Dict):
 
         logging.info(f"Invoking tool {self.unmocked_tool}")
         output = self.unmocked_tool.invoke(params)
-        output = strip_ansi(output)
+        if output.status == ToolResultStatus.SUCCESS:
+            output = strip_ansi(output.data)
+        elif output.status == ToolResultStatus.ERROR:
+            output = strip_ansi(output.error)
```
nherment: A tool could still have `data` even if it failed. For example, a health check tool may 'fail' but still return detailed status information.

holmes/core/tool_calling_llm.py
```diff
@@ -247,12 +248,16 @@ def call(
                 for future in concurrent.futures.as_completed(futures):
                     tool_call_result: ToolCallResult = future.result()
                     tool_calls.append(tool_call_result)
+
+                    tool_call_result = tool_call_result.result.data
```
nherment: `tool_call_result.result` seems to still be typed as a `str` which seems incorrect?

nherment: Actually, both `StructuredToolResult` and `ToolCallResult` be merged together ? The new `StructuredToolResult` seem to be overlapping a lot with `ToolCallResult`.
I feel `StructuredToolResult` is introducing more complexity and confusion given we already have a structured tool call result. If so that's my bad...

@itisallgood and @moshemorad thoughts?

holmes/core/tool_calling_llm.py
```diff
@@ -547,8 +557,20 @@ def stream_analysis(it, peek_chunk):
 
                 for future in concurrent.futures.as_completed(futures):
                     tool_call_result: ToolCallResult = future.result()
+
+                    tool_response = tool_call_result.result.data
+                    if tool_call_result.status == ToolResultStatus.ERROR:
+                        tool_response = f"Tool execution failed with error: {tool_call_result.error_message}\n\nOutput:\n{tool_response}"
```
nherment: `tool_call_result.error_message` should already contain an introduction to the error. It is redundant to say there was an error before the actual error message. 
Also rather than reusing `tool_response` I would increase readability by referring to `tool_call_result.result.data` again:

```python
tool_response = tool_call_result.result.data
if tool_call_result.status == ToolResultStatus.ERROR:
    tool_response = f"{tool_call_result.error_message or 'Tool execution failed'}:\n\n{tool_call_result.result.data or ''}".strip()
```

nherment (changes requested):
Looking really good!

I have some questions about:
- Whether adding `return_code` is really necessary
- Simplifying how the yaml tool subprocess returns data
- Potentially merging `StructuredToolResult` with `ToolCallResult` which I'm aware has big implications for the work you have put in.

holmes/plugins/toolsets/grafana/toolset_grafana_loki.py
itisallgood: We initially thought that it would improve LLM's response, but I agree there is no reason to do it

holmes/core/tools.py
itisallgood: I would prefer to leave this field as an additional info, FE doesn't have to rely on it

holmes/core/tool_calling_llm.py
itisallgood: I would leave as it is. Let `StructuredToolResult` to just enrich the tool's response

holmes/core/tools.py
itisallgood: I updated the name and now I put commands there as well as queries

tests/llm/utils/mock_toolset.py
itisallgood: With updated logic, I think it is passing the whole output

holmes/core/tools.py
itisallgood: Updated

holmes/core/tool_calling_llm.py
```diff
@@ -289,7 +295,12 @@ def _invoke_tool(
             logging.error(
                 f"Tool call to {tool_name} failed with an Exception", exc_info=True
             )
-            tool_response = f"Tool call failed: {e}"
+            tool_response = StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=f"Tool call failed: {e}",
+                return_code=-1,
```
nherment: Why is return code set to `-1`? `return_code` is optional so I'd expect it to be None for every single case where we don't have a valid one.

holmes/plugins/toolsets/internet/notion.py
nherment: Seems it's still there :)

holmes/plugins/toolsets/internet/internet.py
nherment: Should `return_code=-1` be removed?

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -591,18 +662,33 @@ def _invoke(self, params: Any) -> str:
                     )
                 except json.JSONDecodeError:
                     pass
-                return (
-                    f"Query execution failed. HTTP {response.status_code}: {error_msg}"
+                return StructuredToolResult(
+                    status=ToolResultStatus.ERROR,
+                    error=f"Query execution failed. HTTP {response.status_code}: {error_msg}",
+                    return_code=-1,
```
nherment: Should `return_code=-1` be removed?

tests/llm/utils/mock_toolset.py
nherment: I think the code here actually needs to change by simply marshaling the tool output:

```
        output = self.unmocked_tool.invoke(params)
        content = output.model_dump_json(indent=2)
        with open(mock_file_path, "w") as f:
            f.write(mock_metadata_json + "\n")
            f.write(content)

        return output
```

tests/llm/utils/mock_toolset.py
```diff
@@ -62,15 +63,24 @@ def _auto_generate_mock_file(self, params: Dict):
...
 
-    def _invoke(self, params) -> str:
-        return self._auto_generate_mock_file(params)
+    def _invoke(self, params) -> StructuredToolResult:
+        return StructuredToolResult(
+            status=ToolResultStatus.SUCCESS,
+            data=self._auto_generate_mock_file(params),
+            return_code=0,
+            params=params,
+        )
```
nherment: And here just pass the result as it was before:

```
    def _invoke(self, params) -> str:
        return self._auto_generate_mock_file(params)
```

tests/llm/utils/mock_toolset.py
```diff
@@ -62,15 +63,24 @@ def _auto_generate_mock_file(self, params: Dict):
 
         logging.info(f"Invoking tool {self.unmocked_tool}")
         output = self.unmocked_tool.invoke(params)
-        output = strip_ansi(output)
```
nherment, line 65: `ToolMock` defined in this file likely needs to change so that `return_value` is the structured output.

nherment (changes requested):
Looks good. I think there are some remaining `return_code=-1` when it should probably ne `None` ?
In addition it seems the eval mocking needs to be updated to persist and load structured output.

nherment (approved):
lgtm!

PR #376, opened 2025-04-17
ROB-1032 llm selection

RoiGlinik:
Reads model credentials file with multiple models.
Runs Holmes requests with the specified model key. if no model requested uses the first one from the file.

does not effect cli as cli can work with config files as well.


holmes/config.py
```diff
@@ -41,6 +41,46 @@
 from enum import Enum
 
 DEFAULT_CONFIG_LOCATION = os.path.expanduser("~/.holmes/config.yaml")
+MODEL_CREDENTIALS_LOCATION = os.environ.get(
+    "MODEL_CREDENTIALS_LOCATION", "/etc/holmes/model/model_credentials.yaml"
+)
+
+
+def load_yaml_file(path: str, raise_error: bool = True) -> Dict:
```
itisallgood: I believe, we have yaml file loading logic in a lot of places. Could you move it out of `config.py` to a more appropriate place, so it can be reused by others ?

holmes/config.py
```diff
@@ -175,6 +188,17 @@ def parse_toolsets_file(
     return toolset_config
 
 
+def parse_models_file(path: str):
+    models = load_yaml_file(path, raise_error=False)
```
itisallgood, line 155: Shouldn't we raise errors and tell the user that credentials are wrong?

holmes/config.py
```diff
@@ -650,9 +675,26 @@ def load_from_file(cls, config_file: Optional[str], **kwargs) -> "Config":
...
+            )
+            api_key = model_info.pop("api_key", None)
+            self.model = model_info.pop("model", None)
+
+        return DefaultLLM(self.model, api_key, model_info)
+
+    def get_models_list(self) -> List[str]:
+        if self._models_credentials:
+            return json.dumps(list(self._models_credentials.keys()))
+        else:
```
itisallgood: redundant `else`
```
 if self._models_credentials:
            return json.dumps(list(self._models_credentials.keys()))
 return json.dumps([self.model])
```

itisallgood (changes requested):
Left some things that need to be addressed

helm/holmes/values.yaml
```diff
@@ -53,3 +53,5 @@ additionalVolumes: []
 additionalVolumeMounts: []
 
 priorityClassName: ""
+
+modelCredentials: {}
```
itisallgood: I misunderstood the logic, but what happens if we have modelCredentials defined along with additional envs?  What does happen in a case when modelCredentials fail? Do we raise an error and we try with envs?

RoiGlinik: if model credentials fails and is empty, it will work with env vars.

holmes/config.py
RoiGlinik: I could add some env var to contorl the flow, but now it always tries to read the file so i dont want it to raise errors for users who dont use it

server.py
```diff
@@ -355,7 +356,7 @@ def chat(chat_request: ChatRequest):
 
 @app.get("/api/model")
 def get_model():
```
moshemorad, line 358: Where this is being used?

moshemorad (approved):

itisallgood (approved):

server.py
RoiGlinik: runner uses this endpoint

PR #377, opened 2025-04-17
ROB-1072: add eval + tests for coralogix

nherment:


.gitignore
```diff
@@ -1,6 +1,5 @@
 .DS_STORE
 tmp/
-config.yaml
```
moshemorad: Please revert, It might include sensitive data and each one of us use it differently.
I saw that you might added it due to the tests config. so just make sure it won't include also config.yaml that we have on root folder. Or change the name of the file in the tests.

At least this is how I work with the --config flag. If you think it is only me we can proceed, but double check with others.

moshemorad (changes requested):

.gitignore
nherment: done. Renamed the tests configs to `toolsets.yaml`

moshemorad (approved):

PR #385, opened 2025-04-23
Structured output feature flag

itisallgood:


holmes/core/tool_calling_llm.py
```diff
@@ -45,16 +49,22 @@ class ToolCallResult(BaseModel):
...
     tool_name: str
     description: str
-    result: StructuredToolResult
+    result: Union[StructuredToolResult, str]
     size: Optional[int] = None
 
     def as_dict(self):
+        # TODO: remove this after FE is ready
+        if isinstance(self.result, StructuredToolResult):
+            result = self.result.model_dump()
```
nherment, line 58: Shouldn't this use something like the function below? ToolCallResult.result use to be a string and `model_dump()` returns a dict.

```python
def format_structured_tool_result(tool_call_result:StructuredToolResult):
   tool_response = tool_call_result.result.data
   if tool_call_result.result.status == ToolResultStatus.ERROR:
     tool_response = f"{tool_call_result.result.error or 'Tool execution failed'}:\n\n{tool_call_result.result.data or ''}".strip()
   return tool_call_result
```

nherment: Well, this is actually the same as `_get_formatted_tool_call_response_content` defined below. `_get_formatted_tool_call_response_content` does not need to be a class method but can be an independent function which then allows you to use it here.

holmes/core/tool_calling_llm.py
```diff
@@ -245,22 +255,35 @@ def call(
...
                     tool_call_result: ToolCallResult = future.result()
-                    tool_calls.append(tool_call_result)
 
-                    tool_response = tool_call_result.result.data
-                    if tool_call_result.result.status == ToolResultStatus.ERROR:
-                        tool_response = f"{tool_call_result.result.error or 'Tool execution failed'}:\n\n{tool_call_result.result.data or ''}".strip()
+                    tool_response_content = (
+                        self._get_formatted_tool_call_response_content(tool_call_result)
+                    )
+                    # TODO: remove this after FE is ready
```
nherment: Either remove the TODO or explain further on what condition this can be removed

holmes/core/tool_calling_llm.py
```diff
@@ -245,22 +255,35 @@ def call(
...
+                    tool_call_response = tool_call_result.as_dict()
+                    if HOLMES_STRUCTURED_OUTPUT_CONVERSION_FEATURE_FLAG:
+                        tool_call_response["result"] = tool_response_content
+
+                    tool_calls.append(tool_call_response)
```
nherment, line 268: This used to be `tool_call_result` directly before the struct tool result change. Why does it need to change?

holmes/core/tool_calling_llm.py
```diff
@@ -568,31 +591,37 @@ def stream_analysis(it, peek_chunk):
...
+                    tool_response_content = (
+                        self._get_formatted_tool_call_response_content(tool_call_result)
+                    )
 
                     message_to_append = {
                         "tool_call_id": tool_call_result.tool_call_id,
                         "role": "tool",
                         "name": tool_call_result.tool_name,
-                        "content": tool_response,
+                        "content": tool_response_content,
```
nherment, line 604: :+1:

holmes/core/tool_calling_llm.py
```diff
@@ -568,31 +591,37 @@ def stream_analysis(it, peek_chunk):
...
+                            "role": "tool",
+                            "name": tool_call_result.tool_name,
+                            "content": tool_response_content,
+                        }
+                    else:
+                        result_dict = {
+                            "tool_call_id": tool_call_result.tool_call_id,
+                            "role": "tool",
+                            "name": tool_call_result.tool_name,
+                            "result": tool_call_result.result.model_dump(),
```
nherment: Shouldn't this be a string with the same value of `_get_formatted_tool_call_response_content`

nherment (approved):

nherment (approved):

PR #387, opened 2025-04-24
ROB-1177: coralogix enhanced logs

nherment:
The main goal of this PR is to adopt the same approach as the loki toolset, using the pair `resource_type` and `resource_name` for arguments to fetching coralogix logs.

In addition:
- Tags are removed from the output. Seeing how Holmes actually uses logs and makes decision based on that I have removed the tags from the output. This helps reducing the token count and I expect no impact on Holmes at all.
- Added URL and query to the structured output. It was removed by mistake in the previous PR.
- Added URL and query to the beginning of the stringified log results so that user can still access the coralogix query by copy pasting the link (e.g. from slack)


arikalon1 (approved):
LGTM

PR #393, opened 2025-05-02
chore(toolset): add Azure Kubernetes Service toolset

mainred:
This PR adds a list of read operations to Azure Kubernetes Service Services. Also, a cloud_provider tool is added to determined the cloud provider of the Kubernetes cluster in current context before we can have a toolset somewhere to host cloud provider related tools


arikalon1 (approved):
Thanks for the PR @mainred 

Very nice work!

Can you please sign the CLA, so we can merge it?

PR #395, opened 2025-05-03
Support asking clarifying questions and allow user to follow up after investigation

lightme16:
To make HolmesGPT more helpful, I added support for asking for clarifications. If a user asks, "Why did my deployment fail?" HolmesGPT will prompt them with, "What's the deployment name?" and "What’s the namespace?". Only after that it would start investigation.

Also, after the investigation wraps up, users can follow up on the report. They can ask more questions since that's often needed. You’ll want to keep iterating once HolmesGPT finds any problems.


arikalon1 (dismissed):
Thanks for the PR @lightme16 

Great enhancement for Holmes!
Nice and clean

arikalon1 (dismissed):

arikalon1 (dismissed):

arikalon1 (dismissed):

PR #404, opened 2025-05-06
DONGMERGE chore: allow customer to specify the AKS env in aks toolset

mainred:
Normally, a user has to specify the cluster name, resource group and subscription ID to locate an AKS cluster. Considering it's not possible to extract these AKS info from the Kubernetes resources, To store the context on the cluster info, we require the user to set these environment RESOURCE_GROUP_NAME, CLUSTER_NAME and SUBSCRIPTION_ID to load aks toolset. This saves the effort to provide the same information every time they want to debug an AKS cluster.

### DIFF
--- a/holmes/plugins/toolsets/aks.yaml
+++ b/holmes/plugins/toolsets/aks.yaml
@@ -6,7 +6,10 @@ toolsets:
     prerequisites:
       - command: "az aks --help"
       - command: "kubectl version --client"
-
+      - env:
+          - RESOURCE_GROUP_NAME
+          - CLUSTER_NAME
+          - SUBSCRIPTION_ID
     tools:
       - name: "cloud_provider"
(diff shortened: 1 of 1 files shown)


arikalon1 (approved):
thanks for the PR @mainred 

It is more user friendly this way

arikalon1:
@mainred I merged the PR, then saw the prefix `DONGMERGE chore` 
What did you mean by that?

mainred:
@arikalon1 It's ok to merge it. I wanted to leverage this change to provide AKS cluster info as the context of the following asks. But https://github.com/robusta-dev/holmesgpt/pull/395 this PR can help request the cluster info, and interactive conversation will wrap this cluster info in the following questions, which means we have ways to store the cluster info context.
The interactive conversation stores the cluster info in the scope of a conversation, while another new conversation will have to specify the cluster info once again, but we can use this PR to get the cluster info the env.

Thank you for merging this PR.

PR #407, opened 2025-05-08
ROB-846: allow direct conneciton to loki and tempo

nherment:


nherment:
docs: https://github.com/robusta-dev/robusta/pull/1808

holmes/plugins/toolsets/grafana/common.py
```diff
@@ -3,12 +3,21 @@
...
+    requests through grafana. In this case `url` should be the grafana URL.
+    If `grafana_datasource_uid` is not set, it is assumed that the `url` is the
+    systems' URL
+    """
+
     api_key: Optional[str] = None
     headers: Optional[Dict[str, str]] = None
     url: str
-    grafana_datasource_uid: str
+    grafana_datasource_uid: Optional[str] = None
```
moshemorad, line 20: Thinking here loud, If anyway what grafana does it is just proxy to the real service behind the data source why we want grafata/X toolsets?

Why not change those toolsets to LokiToolsets etc and either in the docs explain how to set grafana data source url or have GrafanaLokiToolset which inheritance/composite the LokiToolset and modify it accordingly? 

The only reason i think that we would like to actually keep the GrafanaXToolset is for backward compatibility and other specific logic like the health (Which honestly i think we also need to add more support for the actual system health check if it is only proxy)

moshemorad (changes requested):

holmes/plugins/toolsets/grafana/common.py
nherment: I kept Grafana Loki and Grafana Tempo for a few reasons:
- It was simpler to implement and maintains backward compatibility
- They work very similarly so it does make sense to have a common base toolset between both
- The official system names are indeed 'Grafana Loki' and 'Grafana Tempo' so I don't think it's very confusing to name the toolsets accordingly

There may be an argument to rename the toolsets to `grafana_loki/logs` and `grafana_tempo/traces` to differentiate them from future 'grafana_loki/admin' or 'grafana_loki/health' related toolsets that do not provide logs.

wdyt @moshemorad ?

moshemorad: Sounds good. My only concern with rename is backward compatibility? 
We might handle it by replacing string in the yaml WDYT? if we go that way please also add warning for deprecating the support.

moshemorad (approved):

PR #408, opened 2025-05-09
ROB-1267: initial implementation

nherment:


holmes/plugins/toolsets/kubernetes_logs.py
```diff
@@ -0,0 +1,283 @@
...
+                ) + self._fetch_pod_logs(params=params, previous=False)
+            else:
+                all_logs = self._fetch_pod_logs(
+                    params=params, containers=pod.containers, previous=True
+                )
+                all_logs = all_logs + self._fetch_pod_logs(
+                    params=params, containers=pod.containers, previous=False
+                )
+
+            if params.filter_pattern:
```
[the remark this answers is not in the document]
nherment: No, that's on purpose. Both current and previous logs can be present if a pod is restarted as the pod can retain its name.

holmes/plugins/toolsets/kubernetes_logs.py
```diff
@@ -0,0 +1,283 @@
...
+                all_logs = self._filter_logs(all_logs, params.filter_pattern)
+
+            if params.limit and params.limit < len(all_logs):
+                all_logs = all_logs[-params.limit :]
+
```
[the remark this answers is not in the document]
nherment: No.

holmes/plugins/toolsets/logging_api.py
```diff
@@ -0,0 +1,206 @@
...
+    pod: str
+    namespace: str
+
+
+class LoggingConfig(BaseModel):
+    """Base configuration for all logging backends"""
+
+    labels: LoggingLabelsConfig
+    # Common parameters like timeouts, defaults, etc.
+
```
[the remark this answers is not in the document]
nherment: no

tests/llm/fixtures/test_ask_holmes/08_sock_shop_frontend/test_case.yaml
```diff
@@ -17,3 +17,4 @@ after_test: |
   kubectl delete -f ./helm/trigger-carts-issue.yaml
 evaluation:
   correctness: 1
+generate_mocks: False
```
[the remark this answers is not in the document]
nherment: no

tests/llm/fixtures/test_ask_holmes/20_long_log_file_search/fetch_logs.txt
```diff
@@ -0,0 +1,3 @@
+{"toolset_name":"kubernetes/logs","tool_name":"fetch_logs","match_params":{"namespace":"default","pod_name":"network-connector","filter":"Peer down"}}
+{"schema_version": "robusta:v1.0.0", "status": "success", "error": null, "return_code": null, "data": null, "url": null, "invocation": null, "params": {"namespace": "default", "pod_name": "network-connector", "start_time": null, "end_time": null, "filter_pattern": "Peer down", "limit": 2000}}
+Peer down
```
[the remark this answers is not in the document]
nherment: incorrect

holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
```diff
@@ -150,3 +156,84 @@
...
+                ]
+            except re.error:
+                # If regex is invalid, return unfiltered logs with a warning
+                pass
+
+        # Apply limit
+        if limit and limit < len(filtered_logs):
+            filtered_logs = filtered_logs[-limit:]  # Get the most recent logs
+
+        # Format the logs
```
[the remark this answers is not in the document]
nherment: @coderabbitai that makes no sense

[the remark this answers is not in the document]
nherment: no, that's not how it works

holmes/plugins/toolsets/logging_api.py
```diff
@@ -0,0 +1,205 @@
...
+
+class FetchLogsParams(BaseModel):
+    namespace: str
+    pod_name: str
+    start_time: Optional[str] = None
+    end_time: Optional[str] = None
+    filter_pattern: Optional[str] = None
+    limit: Optional[int] = None
+
```
moshemorad: I think that since pod_name is required it means that this is specificly for fetching pod logs. Can we make it specific? Do you think it is good idea? are you hacing future plan and this is why the name is general?
```suggestion
class FetchPodLogsParams(BaseModel):
    namespace: str
    pod_name: str
    start_time: Optional[str] = None
    end_time: Optional[str] = None
    filter_pattern: Optional[str] = None
    limit: Optional[int] = None

```

holmes/plugins/toolsets/logging_api.py
```diff
@@ -0,0 +1,205 @@
...
+                    type="string",
+                    required=False,
+                ),
+            },
+        )
+        self._toolset = toolset
+
+    def _invoke(self, params: dict) -> StructuredToolResult:
+        structured_params = FetchLogsParams(
+            namespace=get_param_or_raise(params, "namespace"),
```
moshemorad: Do we really need get_param_or_raise? wouldn't it fail anyway on the pydantic validation of the field?

holmes/plugins/toolsets/logging_api.py
```diff
@@ -0,0 +1,205 @@
...
+            params=structured_params,
+        )
+
+        return result
+
+    def get_parameterized_one_liner(self, params: dict) -> str:
+        """Generate a one-line description of this tool invocation"""
+        namespace = params.get("namespace", "unknown-namespace")
+        pod_name = params.get("pod_name", "unknown-pod")
+        filter_desc = (
```
moshemorad: We don't want to add here also the other params? I think that for example start_time & end_time might also provide relevant info. WDYT?

holmes/plugins/toolsets/logging_api.py
```diff
@@ -0,0 +1,205 @@
...
+            f" with filter '{params.get('filter')}'" if params.get("filter") else ""
+        )
+        return f"Fetching logs for pod {pod_name} in namespace {namespace}{filter_desc}"
+
+
+def process_time_parameters(
```
moshemorad: Not in use?

holmes/plugins/toolsets/git.py
```diff
@@ -69,6 +69,9 @@ def _sanitize_error(self, error_msg: str) -> str:
         return error_msg.replace(self.git_credentials, "[REDACTED]")
 
     def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
```
moshemorad: Is it relevant to logging? doesn't look like it. Can you please move it to other PR? Thanks.

holmes/plugins/prompts/_fetch_logs.jinja2
```diff
@@ -1,44 +1,12 @@
-{%- set loki_ts = toolsets | selectattr("name", "equalto", "grafana/loki") | first -%}
-{%- set coralogix_ts = toolsets | selectattr("name", "equalto", "coralogix/logs") | first -%}
-{%- set k8s_ts = toolsets | selectattr("name", "equalto", "kubernetes/logs") | first -%}
-{%- set opensearch_ts = toolsets | selectattr("name", "equalto", "opensearch/logs") | first -%}
-
 # Logs
```
moshemorad, line 7: What do you think about moving it to the toolset level instead and load it from the relevant logs toolset?

holmes/plugins/toolsets/kubernetes_logs.py
```diff
@@ -0,0 +1,284 @@
...
+        except Exception as e:
+            return False, f"Kubernetes client initialization error: {str(e)}"
+
+    def get_example_config(self):
+        return LoggingConfig().model_dump()
+
+    def _initialize_client(self):
+        if not self._api_client:
+            try:
+                try:
```
moshemorad: Please extract to method and add logs in debug level. I think that logging is something that we start and put more attention to. It will also help us debug stuff much easier.

holmes/plugins/toolsets/kubernetes_logs.py
```diff
@@ -0,0 +1,284 @@
...
+            )
+        except Exception as e:
+            logging.exception(f"Error fetching logs for pod {params.pod_name}")
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=f"Error fetching logs: {str(e)}",
+            )
+
+    def _find_pod(self, namespace: str, pod_name: str) -> Optional[Pod]:
+        if not self._core_v1_api:
```
moshemorad, line 154: Not sure i fully understand. In which case we can get to here and the _core_v1_api isn't initialized? 
If it is safe guard i think that we atleast need to add here warning log.

holmes/plugins/toolsets/kubernetes_logs.py
```diff
@@ -0,0 +1,284 @@
...
+                    container_name=container_name,
+                    previous=previous,
+                )
+
+                # Add container name prefix only for multi-container pods
+                # This matches kubectl behavior
+                container_logs = [f"{container_name}: {log}" for log in container_logs]
+                pod_logs.extend(container_logs)
+
+        if not pod_logs:
```
moshemorad, line 192: Just to make sure that it is be design, if we try to call fetch_logs for each container and none of them return logs because no logs exists we still call this? Right?
If this is indeed the case can you elaborate the use case for it?

holmes/plugins/toolsets/kubernetes_logs.py
```diff
@@ -0,0 +1,284 @@
...
+    """
+    # match ISO 8601 format (YYYY-MM-DDTHH:MM:SS[.fffffffff]Z)
+    timestamp_pattern = re.compile(
+        r"^(?P<ts>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2}))"
+    )
+    filtered_lines_content: List[str] = []
+
+    for line in logs:
+        # Non-string entries are passed through as-is
+        if not isinstance(line, str):
```
moshemorad: What is this case? can you add also a commnt to document it?

holmes/plugins/toolsets/kubernetes_logs.py
```diff
@@ -0,0 +1,284 @@
...
+    def _format_logs(self, logs: List[str]) -> str:
+        """Format logs for output"""
+        if not logs:
+            return "No logs found"
+
+        # Simple join without line numbers to match kubectl logs format
+        return "\n".join([log for log in logs if log.strip()])
+
+
+def filter_log_lines_by_timestamp_and_strip_prefix(
```
moshemorad: Shouldn't this be part of the toolset class? do we think we will need to use it in other cases?

holmes/plugins/toolsets/kubernetes_logs.py
```diff
@@ -0,0 +1,284 @@
...
+            if params.filter_pattern:
+                all_logs = self._filter_logs(all_logs, params.filter_pattern)
+
+            if params.limit and params.limit < len(all_logs):
+                all_logs = all_logs[-params.limit :]
+
+            # Format output
+            formatted_logs = self._format_logs(all_logs)
+
+            return StructuredToolResult(
```
moshemorad, line 132: I think that in formatted logs we are return str even if no logs are found. In the structure tool response we also added support for no data.
We need to also support that case. e.g. no logs the status should not be success but no data.

holmes/plugins/toolsets/logging_api.py
nherment: I don't think that;s the right approach. The tool is called `fetch_logs`, the toolset function is called `fetch_logs`. It makes sense to me that the params are called `FetchLogsParams` and not `FetchPodLogsParams`.

holmes/plugins/toolsets/logging_api.py
nherment: pydantic validation generates atrocious validation messages.
Linting will fail if the param is passed as is.

holmes/plugins/prompts/_fetch_logs.jinja2
nherment: Why would that be better?

holmes/plugins/toolsets/git.py
nherment: It's relevant because evals are broken because of this.
Because I touched the file I also fixed the linting. Is it really necessay to move it to another PR?

holmes/plugins/toolsets/kubernetes_logs.py
nherment: Yes this should really not be necessary but it's due to how the toolsets have been designed to take the config. It's a tech debt we've been dragging for a long time. I'll add the log

holmes/plugins/toolsets/kubernetes_logs.py
nherment: Good question, I added a comment to explain it.
It's a way to be resilient, in particular for previous logs. The container list is a 'current' container list (as far as I know I can't get a 'previous container' list). This means that fetching previous logs with specific containers will work in most cases but may fail to fetch logs if the container names have changed

holmes/plugins/toolsets/kubernetes_logs.py
nherment: If it is needed as common code then it'll be moved there. It's not needed today so it should stay here. wdyt?

holmes/plugins/toolsets/kubernetes_logs.py
nherment: It's defensive code. should actually be `filtered_lines_content.append(str(line))`. Fixed

holmes/plugins/toolsets/logging_api.py
nherment: I've actually simplified this. A single string in a line is a great way to describe a command line but a terrible way to describe used params.
params are returned to the UI as-is and if that is important to users there should be a way to display params in a better format. Long lines can be clipped in the SaaS UI and hard to see.

holmes/plugins/toolsets/kubernetes_logs.py
nherment: Added a debug line. I don't think it's best to extract to another method. This method here is small and it would make the code harder to read to split it further.

holmes/plugins/toolsets/logging_api.py
moshemorad: What im concern with the current implementation is that we need to remember to add the get_param_or_raise for all required params from now on. To be honest it is something that we will forget.

So either we need to have some function that checks the models and validate it or add logic that catch pydantic ValidationError and handle it as we expected. You can see example here: https://docs.pydantic.dev/latest/errors/errors/#customize-error-messages

If you think that we can handle it in another sub task it is OK by me but let's open it and make sure we are doing that.

holmes/plugins/prompts/_fetch_logs.jinja2
moshemorad: We won't need to remember to add here the prompt for every toolset being added. When the toolset loaded it will add relevant prompt instead. If you think otherwise you can ignore.

moshemorad (changes requested):
Few comments

holmes/plugins/toolsets/logging_api.py
moshemorad: I probably wasn't clear enough. What i meant is that from my pov we need to change the tool to be called FetchPodLogs, and align all names to it. 
The main reason is we agreed the the api will be very specific, we want to start with the most simple and naive implementation and improve from there. LogsToolset is a good name but the tools names should specific to the cases. e.g. FetchPodLogs, FetchDeploymentLogs etc. Otherwise how the LLM will know to select the relevant tools?

The the result will be that we have LogsToolset which has all relevant api (tools) we expect
LogsToolset
- FetchPodLogs
- FetchDeploymentLogs
- FetchStatefulesetLogs 
- etc

And all Specific implementation logs toolset will need to implement those capabilities.

holmes/plugins/toolsets/kubernetes_logs.py
```diff
@@ -0,0 +1,281 @@
...
+            self._api_client = client.ApiClient()
+            self._core_v1_api = client.CoreV1Api(self._api_client)
+        except Exception:
+            logging.error("Failed to initialize Kubernetes client", exc_info=True)
+
+    def fetch_logs(self, params: FetchLogsParams) -> StructuredToolResult:
+        try:
+            pod = self._find_pod(params.namespace, params.pod_name)
+            all_logs = []
+            if not pod or not pod.containers:
```
moshemorad: Why we want to fetch logs if 'not pod'? e.g. if we couldn't find the pod how we can fetch it logs?

holmes/plugins/toolsets/logging_api.py
nherment: I think the more likely evolution for this tool is to still be called `fetch_logs` and have the key arguments be `resource_type` and `resource_name`.

holmes/plugins/toolsets/logging_api.py
nherment: Yes, agree, and this can be part of a toolset refactor. Added to https://robusta-dev.atlassian.net/browse/ROB-1243

holmes/plugins/prompts/_fetch_logs.jinja2
nherment: We won't need to remember anything once we migrate all tools because the logs related prompt will be the same across all logging toolsets.

holmes/plugins/toolsets/kubernetes_logs.py
nherment: A terminated pod may return nothing when `_find_pod` is called.

holmes/plugins/toolsets/logging_api.py
moshemorad: And in that case it we wouldn't name it something more specific like FetchK8SLogs or FetchResourceLogs? I think that the whole idea of this task was to make thing very simple and straight forward. FetchLogs is to generic.

holmes/plugins/toolsets/kubernetes_logs.py
moshemorad: In that case logs are indeed still available?

.github/workflows/llm-evaluation.yaml
moshemorad, line 1: This is a feature branch we should only merge it after finished all the tasks.

holmes/plugins/toolsets/kubernetes_logs.py
nherment: yes, sometimes

moshemorad (dismissed):

holmes/plugins/toolsets/opensearch/opensearch_logs.py
```diff
@@ -1,347 +1,151 @@
...
-            if nested_properties:
-                self._extract_fields_from_properties(
-                    nested_properties, f"{full_path}.", field_details, index_name
+        try:
+            start_time = None
+            end_time = None
+            if params.start_time or params.end_time:
+                start_time, end_time = process_time_parameters(
+                    params.start_time, params.end_time
                 )
```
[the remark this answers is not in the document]
nherment: No

holmes/plugins/toolsets/coralogix/api.py
```diff
@@ -54,35 +53,32 @@ def health_check(domain: str, api_key: str) -> Tuple[bool, str]:
...
+def build_query_string(config: CoralogixConfig, params: FetchPodLogsParams) -> str:
     query_filters = []
-    if namespace_name:
-        query_filters.append(f"{config.labels.namespace}:{namespace_name}")
-    query_filters.append(f"{label}:/{resource_name}/")
+    query_filters.append(f'{config.labels.namespace}:"{params.namespace}"')
+    query_filters.append(f'{config.labels.pod}:"{params.pod_name}"')
+
+    if params.match:
+        query_filters.append(f'{config.labels.log_message}:"{params.match}"')
```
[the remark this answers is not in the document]
nherment: search API is read only,

tests/llm/utils/mock_toolset.py
```diff
@@ -89,24 +92,50 @@
...
+            if span:
+                span.log(
+                    input=params,
+                    output=str(e),
+                )
+            raise
+        finally:
+            if span:
+                span.end()
+        return result
```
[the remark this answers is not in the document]
nherment: The code here ensures span.end() is called

holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
```diff
@@ -26,80 +24,69 @@
...
+            name="coralogix/logs",
+            description="Toolset for interacting with Coralogix to fetch logs",
+            docs_url="https://docs.robusta.dev/master/configuration/holmesgpt/toolsets/coralogix_logs.html",
+            icon_url="https://avatars.githubusercontent.com/u/35295744?s=200&v=4",
+            prerequisites=[CallablePrerequisite(callable=self.prerequisites_callable)],
+            tools=[
+                PodLoggingTool(self),
+            ],
+            tags=[ToolsetTag.CORE],
+        )
```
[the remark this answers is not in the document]
nherment: design issue with toolsets. by design `fetch_pod_logs` should not be called unless prerequisites are called..

.github/workflows/build-and-test.yaml
```diff
@@ -11,13 +11,21 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
+      - name: Install Poetry
```
moshemorad: Looks like this PR for some reason contains other changes from master branch, Can you please fix that since the PR is already huge and it is very hard to understand which change is relevant and which isn't. 

Thanks

nherment: I think this was because the branch was not up to date.

holmes/plugins/toolsets/logging_api.py
nherment: We aligned on fetch_pod_logs. Resolving this comment.

PR #414, opened 2025-05-13
Adding new tools to aks-node-health toolset

pavneeta:


holmes/plugins/toolsets/aks-node-health.yaml
```diff
@@ -13,43 +13,46 @@ toolsets:
...
-        description: "Fetches kubelet logs from a specific node in the AKS cluster."
-        user_description: "get kubelet logs for node {{ NODE_NAME }}"
-        command: |
-          ssh {{ NODE_NAME }} 'journalctl -u kubelet'
-
-      - name: "check_control_plane_connectivity"
-        description: "Verifies if the node can communicate with the control plane."
-        user_description: "check control plane connectivity for node {{ NODE_NAME }}"
+        description: "Fetches kubelet logs from a specific node in the AKS cluster using Azure CLI VMSS run command."
+        user_description: "get kubelet logs for node {{ NODE_NAME }} in the AKS cluster"
```
arikalon1, line 23: Should the parameters in the `user description` match the command parameters ?

arikalon1 (approved):
thanks for the PR @pavneeta 

Nice work
Left one minor comment

holmes/plugins/toolsets/aks-node-health.yaml
arikalon1: just to clarify, since the `NODE_NAME` is in the user description, but not in the command parameters, it might appear empty when describing the command

PR #419, opened 2025-05-14
ROB-1288 toolsset mcp

RoiGlinik:
## Summary by CodeRabbit

- **New Features**
  - Added support for integrating and invoking remote MCP tools via a new toolset.
  - Users can now dynamically load and manage tools from an MCP server.

- **Tests**
  - Introduced tests to verify tool schema parsing, tool creation, error handling for unreachable servers, and successful tool loading from a mocked server response.

- **Chores**
  - Updated Python version requirement to 3.10 and added a new MCP dependency.
  - Adjusted CI workflow to exclude Python 3.9 from testing matrix.


holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+
+        return parameters
+
+    def get_parameterized_one_liner(self, params: Dict) -> str:
+        return f"Call mcp server {self.url} tool {self.name} with params {str(params)}"
+
+
+class MCPToolset(Toolset):
+    url: str
```
moshemorad: Can we use pydantic AnyUrl?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -0,0 +1,127 @@
...
+                required=key in required_list,
+            )
+
+        return parameters
+
+    def get_parameterized_one_liner(self, params: Dict) -> str:
+        return f"Call mcp server {self.url} tool {self.name} with params {str(params)}"
+
+
+class MCPToolset(Toolset):
```
moshemorad: Thinking here loud, MCPToolset feels to wide e.g. local & remote are both MCPToolset but they have different args. Do you think it is a good idea to make it more specific e.g. RemoteMCPToolset?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @field_validator("url")
+    def sse_endpoint(cls, v):
+        if not v.endswith("/sse"):
+            v = v.rstrip("/") + "/sse"
+
+        return v
+
+    @model_validator(mode="after")
+    def set_tools_prerequisites(self):
```
moshemorad: Any particular reason to do it that way and not how it is done today by init methoud?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -0,0 +1,127 @@
...
+        self.prerequisites = [CallablePrerequisite(callable=self.init_server_tools)]
+        return self
+
+    # used as a CallablePrerequisite, config added for that case.
+    def init_server_tools(self, config: dict[str, Any]) -> Tuple[bool, str]:
+        try:
+            tools_result = asyncio.run(self._get_server_tools())
+            self.tools = [
+                MCPTool.create(self.url, tool, self.headers)
+                for tool in tools_result.tools
```
moshemorad, line 105: What if the server doesn't support any tools? Is it means anything? Not sure we it is really something that we need to be bothered by but still thought about it so raised it.
Maybe add warning log?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -0,0 +1,127 @@
...
+            ]
+            return (True, "")
+        except Exception as e:
+            # using e.args, the asyncio wrapper could stack another exception this helps printing them all.
+            return (
+                False,
+                f"Failed to load mcp server {self.name} {self.url} tools: {str(e.args)}",
```
moshemorad: Not sure i understand e.args indeed contains the tools?

tests/test_mcp_toolset.py
```diff
@@ -0,0 +1,91 @@
...
+        ),
+        "side": ToolParameter(type="string", required=True),
+        "limit_price": ToolParameter(type="number", required=False),
+    }
+    tool_params = MCPTool.parse_input_schema(input_schema_sample)
+
+    assert tool_params == expected_ouput
+
+
+def test_parse_mcp_tool():
```
moshemorad, line 8: Consider to join this test with the one above. Not sure that how this is currently implelemtned gives any value. (We only checking part of the field and not all of them)
Also the tool should create the parms so it feels like it should be joined. WDYT?

tests/test_mcp_toolset.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_mcpserver_1tool(monkeypatch):
+    mcp_toolset = MCPToolset(
+        url="http://0.0.0.0/3005",
+        name="test_mcp",
+        description="demo mcp with 2 simple functions",
+    )
+
+    async def mock_get_server_tools():
```
moshemorad, line 62: Thinking here loud. Do you think it might be worth to mock the http response instead of the func so we will also check the actual http handling?

moshemorad (changes requested):

holmes/plugins/toolsets/mcp/toolset_mcp.py
RoiGlinik: yea its a typo

holmes/plugins/toolsets/mcp/toolset_mcp.py
RoiGlinik: i used the model_post_init i think its a bit cleaner

tests/test_mcp_toolset.py
RoiGlinik: For now ill skip it, I have a test case for an error in that case

moshemorad:
@coderabbitai review

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -14,13 +14,14 @@
 from mcp.types import Tool as MCP_Tool
 
 import asyncio
-from pydantic import field_validator, model_validator, Field
+from pydantic import Field, AnyUrl, field_validator
 from typing import Tuple
+import logging
 
 
 class MCPTool(Tool):
```
moshemorad: Shouldn't we also rename this to RemoteMCPTool?

moshemorad (approved):

vishiy:
hi - mcp tool set support is only for agent ? or will it be for cli too ?

arikalon1:
hi @vishiy 

MCP toolset will be supported for both the CLI and the agent

PR #422, opened 2025-05-16
fix: issues with git toolset

nherment:


itisallgood (approved):
I've already seen those changes and they fixed a few errors in my PRs. LGTM

PR #428, opened 2025-05-18
ROB-1290 parse mcp

RoiGlinik:
## Summary by CodeRabbit

- **New Features**
  - Added support for remote toolsets via URL in toolset configuration.
  - Introduced integration with remote MCP servers to dynamically load and invoke tools.
- **Improvements**
  - Enhanced header management for remote toolsets, allowing headers to be dynamically retrieved from configuration.
- **Tests**
  - Added tests to verify correct handling of HTTP headers in remote toolsets and loading of remote MCP toolsets with environment variable substitution.
- **Chores**
  - Updated server startup process to perform synchronization before server launch.


holmes/core/tools.py
```diff
@@ -115,6 +115,7 @@ class ToolsetTag(str, Enum):
     CORE = "core"
     CLUSTER = "cluster"
     CLI = "cli"
+    MCP = "mcp"
```
moshemorad: I dont think that this is what tags mean, we need a better way to load mcp toolsets.

holmes/config.py
```diff
@@ -612,9 +613,15 @@ def merge_and_override_bultin_toolsets_with_toolsets_config(
                 toolsets_with_updated_statuses[toolset.name].override_with(toolset)
             else:
                 try:
-                    validated_toolset = YAMLToolset(
-                        **toolset.model_dump(exclude_none=True)
-                    )
+                    tags = toolset.tags
+                    if ToolsetTag.MCP in tags:
+                        validated_toolset = RemoteMCPToolset(
```
moshemorad, line 617: Can we try to check if the field url is included and then try to parse it as remote mcp tool instead? Usage of tags here feels like we are twisting the meaning of them.

If any objections let's discuss it.

server.py
```diff
@@ -107,7 +104,7 @@ async def lifespan(app: FastAPI):
         }
     )
 
-app = FastAPI(lifespan=lifespan)
+app = FastAPI()
```
moshemorad, line 107: Why we changed it?

tests/test_mcp_toolset.py
```diff
@@ -78,3 +78,24 @@ async def mock_get_server_tools():
...
+
+
+def test_mcpserver_no_headers():
+    mcp_toolset1 = RemoteMCPToolset(
+        url="http://0.0.0.0/3005",
+        name="test_mcp",
+        description="demo mcp with 2 simple functions",
+    )
+
+    assert mcp_toolset1.get_headers() is None
```
moshemorad, line 101: Please also add test that loaded the toolset from the config flow.

server.py
RoiGlinik: The lifespan function is async and runs in a loop which also loads the toolsets.

The new MCP async code for loading tools was causing issues within this flow , so I changed it to be sync in Holmes main function which simplifies the flow in general.

moshemorad: Can you please share what was failed? Is it the new creation of event loop?

moshemorad (approved):

PR #436, opened 2025-05-22
feat: test case and POC for a solution

nherment:


holmes/core/prompt.py
```diff
@@ -0,0 +1,8 @@
+from pathlib import Path
+
+
+def append_file_to_user_prompt(user_prompt: str, file_path: Path) -> str:
+    f = file_path.open("r")
+    user_prompt += f"\n\nAttached file '{file_path.absolute()}':\n{f.read()}"
```
mainred: Not closely related to this change, but should we just append the user prompt with the content of the file? what's the benefit of `Attached file {fie_name}` inside the user prompt?

holmes/plugins/toolsets/bash/bash_instructions.jinja2
```diff
@@ -0,0 +1,14 @@
+# Bash commands
+
+The tool `run_bash_command` allows you to run many kubectl commands:
```
mainred, line 3: bash command only runs kubectl commands? I prefer an explicit name if we don't have plan to introduce commands other than kubectl?

mainred: Hmmm, I think we are explicitly saying kubectl is currently supported command, we'll introduce more prompt when we support more commands in the future.

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+
+    def _build_kubectl_command(self, params: dict, pod_name: str) -> str:
+        namespace = params.get("namespace", "default")
+        image = get_param_or_raise(params, "image")
+        command_str = get_param_or_raise(params, "command")
+        return f"kubectl run {pod_name} --image={image} --namespace={namespace} --rm --attach --restart=Never -i --tty -- {command_str}"
+
+    def _invoke(self, params: Dict[str, Any]) -> StructuredToolResult:
+        timeout = params.get("timeout", 60)
```
mainred, line 81: What will be returned and feed to llm when the timeout reached?

holmes/plugins/toolsets/bash/kubectl/kubectl_describe.py
```diff
@@ -0,0 +1,66 @@
...
+    )
+    parser.add_argument(
+        "--field-selector",
+        type=regex_validator("field selector", SAFE_SELECTOR_PATTERN),
+    )
+    parser.add_argument("--include-uninitialized", action="store_true")
+
+
+def stringify_describe_command(cmd: Any) -> str:
+    parts = ["kubectl", "describe", cmd.resource_type]
```
mainred, line 46: what's the valid resource type scope?

holmes/plugins/toolsets/bash/kubectl/kubectl_events.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _validate_for_object(value: str) -> str:
+    """Validate the --for object parameter."""
+    MAX_FOR_OBJ_SIZE = 253
+    for_object_pattern = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9\-_./:]*$")
+
+    if not for_object_pattern.match(value):
+        raise argparse.ArgumentTypeError(f"Invalid for_object: {value}")
+    if len(value) > MAX_FOR_OBJ_SIZE:
```
mainred, line 45: Not sure if truncate the event when max size reach is a better idea instead of raising errors, otherwise we cannot get events on a resource exploded with events.

holmes/plugins/toolsets/bash/kubectl/kubectl_get.py
```diff
@@ -0,0 +1,108 @@
+from typing import Any
```
mainred, line 1: Will llm pick the simplified kubectl get command in kubernetes/core instead this one?

holmes/core/prompt.py
nherment: That's how it's been working. A couple notes:
- This change does not change how it was working before. This function has been extracted so that our mock framework can behave the same way as the CLI and append files in the same way to the user prompt
- This behaves similarly to Holmes SaaS that adds account specific instructions (runbooks, etc.) to the initial user prompt.

holmes/plugins/toolsets/bash/bash_instructions.jinja2
nherment: You're correct and I think you raise a good point.

The main thinking leading to this tool being called `run_bash_command` was to make it easier for the LLM to find the right tool. i.e. I suspect it's easier for the LLM to use this tool to run for example `kubectl ... | grep ...` than if the tool was called `kubectl`.

It's a hunch though as I did not objectively verify this.

Over the long term, I do think a bash tool is the right approach as long as the **runbooks and instructions for the LLM are the same as the ones provided to human devops**.

Another consideration: one issue that we are going to have is that today, tool configuration is at the toolset level. This is not very compatible with having a `bash` tool that does mostly everything.

Finally I have doubt whether a single `bash` tool with instructions on what this tool can do is better than specific tools (`kubectl` etc.) like you mention. As the capabilities grow (Azure CLI, etc.) it may be less effective to have a single tool and it will be better to split the capabilities.

holmes/plugins/toolsets/bash/bash_toolset.py
nherment: Further below there is a `except subprocess.TimeoutExpired:` which returns a dedicated error message to the LLM

holmes/plugins/toolsets/bash/kubectl/kubectl_describe.py
nherment: `resource_type` is checked by `validate_kubectl_command` in `holmes/plugins/toolsets/bash/kubectl/__init__.py`.

The list of valid resource types is in `holmes/plugins/toolsets/bash/kubectl/constants.py`

holmes/plugins/toolsets/bash/kubectl/kubectl_events.py
nherment: I'm probably missing something, this does throw if the filter is too long?

holmes/plugins/toolsets/bash/kubectl/kubectl_get.py
nherment: Yes it may. Having multiple tools doing the same thing can lower the LLM's accuracy. 

On the eval for the DNS instructions, Holmes will loose ~10% of accuracy if both the `bash` and the `kubernetes/core` toolsets are enabled.

holmes/core/prompt.py
```diff
@@ -0,0 +1,8 @@
+from pathlib import Path
+
+
+def append_file_to_user_prompt(user_prompt: str, file_path: Path) -> str:
+    f = file_path.open("r")
```
moshemorad: Please use context manager. I suspect that the files that are opened there aren't being closed.

holmes/plugins/toolsets/bash/common/stringify.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    Escape shell arguments to prevent injection.
+    Uses single quotes for safety, escaping any single quotes in the content.
+    """
+    escaped_args = []
+
+    for arg in args:
+        # If argument is safe (contains only alphanumeric, hyphens, dots, underscores, equals, slash, comma, colon)
+        # no escaping needed
+        if arg and all(c.isalnum() or c in ".-_=/,:" for c in arg):
```
moshemorad: Consider change ".-_=/,:" to frozenset and const it on module level.
A. no need to recreate this string every call in the loop
B. Set is more efficient then looping on the string and compare values.

holmes/plugins/toolsets/bash/common/stringify.py
```diff
@@ -0,0 +1,22 @@
+def escape_shell_args(args: list[str]) -> list[str]:
```
moshemorad, line 6: I saw that we use shlex on other places can we use `quote` from there instead of implement by ourselves? do we have more logic then that here and i missing it?
Can we use shelx instead in all the flow?

holmes/plugins/toolsets/bash/kubectl/kubectl_get.py
```diff
@@ -0,0 +1,108 @@
...
+        exit_on_error=False,  # Important for library use
+    )
+    parser.add_argument(
+        "resource_type", type=whitelist_validator("resource type", VALID_RESOURCE_TYPES)
+    )
+    parser.add_argument(
+        "resource_name",
+        nargs="?",
+        default=None,
+        type=regex_validator("namespace", SAFE_NAME_PATTERN),
```
moshemorad:
```suggestion
        type=regex_validator("resource_name", SAFE_NAME_PATTERN),
```

holmes/plugins/toolsets/bash/kubectl/kubectl_get.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+    parser.add_argument(
+        "-n", "--namespace", type=regex_validator("namespace", SAFE_NAMESPACE_PATTERN)
+    )
+    parser.add_argument("-A", "--all-namespaces", action="store_true")
+    parser.add_argument(
+        "-l", "--selector", type=regex_validator("selector", SAFE_SELECTOR_PATTERN)
+    )
+    parser.add_argument(
+        "--field-selector", type=regex_validator("selector", SAFE_SELECTOR_PATTERN)
```
moshemorad:
```suggestion
        "--field-selector", type=regex_validator("field-selector", SAFE_SELECTOR_PATTERN)
```

holmes/plugins/toolsets/bash/kubectl/kubectl_get.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+    parser.add_argument(
+        "-o", "--output", type=whitelist_validator("output", VALID_OUTPUT_FORMATS)
+    )
+    parser.add_argument(
+        "--sort-by", type=regex_validator("namespace", SAFE_NAME_PATTERN)
```
moshemorad:
```suggestion
        "--sort-by", type=regex_validator("sort-by", SAFE_NAME_PATTERN)
```

holmes/plugins/toolsets/bash/kubectl/kubectl_get.py
```diff
@@ -0,0 +1,108 @@
...
+def stringify_get_command(cmd: Any) -> str:
+    parts = ["kubectl", "get", cmd.resource_type]
+
+    # Add resource name if specified
+    if cmd.resource_name:
+        parts.append(cmd.resource_name)
+
+    if cmd.all_namespaces:
+        parts.append("--all-namespaces")
+    if cmd.namespace:
```
moshemorad, line 78: This feels very confusing, i think that if both are provided we need to raise error. What we expect to happen if both were provided?

holmes/plugins/toolsets/bash/kubectl/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+        help="Action to perform (e.g., get, apply, delete)",
+    )
+    create_kubectl_get_parser(action_subparsers)
+    create_kubectl_describe_parser(action_subparsers)
+    create_kubectl_top_parser(action_subparsers)
+    create_kubectl_events_parser(action_subparsers)
+    create_kubectl_logs_parser(action_subparsers)
+
+
+def validate_kubectl_command(cmd: Any) -> None:
```
moshemorad, line 47: Why do we need this validaiton if we also has is as part of each command parser. e.g. in get we also validate those.

But there we aren't validaing the len so maybe we would like to also add this validation there?

holmes/plugins/toolsets/bash/kubectl/kubectl_events.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    parser.add_argument("--types", type=lambda x: _validate_event_types(x))
+    parser.add_argument("-w", "--watch", action="store_true")
+    parser.add_argument("--no-headers", action="store_true")
+
+
+def _validate_for_object(value: str) -> str:
+    """Validate the --for object parameter."""
+    MAX_FOR_OBJ_SIZE = 253
+    for_object_pattern = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9\-_./:]*$")
```
moshemorad: Please move both out of the method unless there is a good reason to recompile the regex every call.

holmes/plugins/toolsets/bash/kubectl/kubectl_events.py
```diff
@@ -0,0 +1,89 @@
...
+    parser.add_argument(
+        "-l",
+        "--selector",
+        type=regex_validator("selector", re.compile(r"^[a-zA-Z0-9\-_.=,!()]+$")),
+    )
+    parser.add_argument(
+        "--field-selector",
+        type=regex_validator("field selector", re.compile(r"^[a-zA-Z0-9\-_.=,!()]+$")),
+    )
+    parser.add_argument(
```
moshemorad: Out of couriosity moving only the funciton name instead of lambda won't work? just trying to understand why we need the lambda here.

holmes/plugins/toolsets/bash/kubectl/kubectl_events.py
```diff
@@ -0,0 +1,89 @@
...
+    if len(value) > MAX_FOR_OBJ_SIZE:
+        raise argparse.ArgumentTypeError(
+            f"for_object too long. Max allowed size is {MAX_FOR_OBJ_SIZE} but received {len(value)}"
+        )
+    return value
+
+
+def _validate_event_types(value: str) -> str:
+    """Validate the --types parameter."""
+    VALID_EVENT_TYPES = {"Normal", "Warning"}
```
moshemorad: Extract from method

holmes/plugins/toolsets/bash/kubectl/kubectl_run.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def validate_image_and_commands(
+    image: str, container_command: str, config: Optional[BashExecutorConfig]
+) -> None:
+    """
+    Validate that the image is in the whitelist and commands are allowed.
+    Raises ArgumentTypeError if validation fails.
+    """
+    if not config or not config.kubectl or not config.kubectl.allowed_images:
```
moshemorad, line 17: It is littlebit confusing that we allow config to be None and then checking for it. Maybe remove the Optional from the type hint?

holmes/plugins/toolsets/bash/bash_instructions.jinja2
moshemorad: @nherment what we decided on then?

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -0,0 +1,303 @@
...
+        super().__init__(
+            name="kubectl_run_image",
+            description=(
+                "Executes `kubectl run <name> --image=<image> ... -- <command>` return the result"
+            ),
+            parameters={
+                "image": ToolParameter(
+                    description="The image to run",
+                    type="string",
+                    required=True,
```
moshemorad, line 50: Consider: add here the available images from the config as enum.

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -0,0 +1,303 @@
...
+                shell=True,
+                executable="/bin/bash",
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                text=True,
+                timeout=timeout,
+                check=False,
+            )
+
+            result_data = (
```
moshemorad: Thinking here loud. 
Don't we want to seperate the stdout & stderr from each other and take only the relevant output?
Another thing. we added support for no_data in status. In the current way we adding the full kubectl command will be confusing. Why we adding it there? can we also make sure to return no_data in relevant cases?

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -0,0 +1,303 @@
...
+
+        image = get_param_or_raise(params, "image")
+        command_str = get_param_or_raise(params, "command")
+
+        namespace = params.get("namespace")
+
+        if namespace and not re.match(SAFE_NAMESPACE_PATTERN, namespace):
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                data=f"Error: The namespace is invalid. Valid namespaces must match the following regexp: {SAFE_NAMESPACE_PATTERN}",
```
moshemorad: Aren't we should put the error message on the error field?

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -0,0 +1,303 @@
...
+        except (argparse.ArgumentError, ValueError) as e:
+            logging.info(f"Refusing LLM tool call {command_str}", exc_info=True)
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=f"Refusing to execute bash command. Only some commands are supported and this is likely because requested command is unsupported. Error: {str(e)}",
+                params=params,
+            )
+
+        try:
+            process = subprocess.run(
```
moshemorad: Can we extract the call to subprocess and reuse in all invoke calls? I think that they are basicly the same right? 
same as the structure result etc...

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -0,0 +1,303 @@
...
+        command = params.get("command", "N/A")
+        display_command = command[:200] + "..." if len(command) > 200 else command
+        return display_command
+
+
+class BashExecutorToolset(BaseBashExecutorToolset):
+    def __init__(self):
+        super().__init__(
+            name="bash",
+            enabled=True,  # Default state; can be overridden by global config.
```
moshemorad: I remebmer that we said that this will be disabled by default. Am i wrong? if yes ignore.

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -0,0 +1,303 @@
...
+        template_file_path = os.path.abspath(
+            os.path.join(os.path.dirname(__file__), "bash_instructions.jinja2")
+        )
+        self._load_llm_instructions(jinja_template=f"file://{template_file_path}")
+
+    def prerequisites_callable(self, config: dict[str, Any]) -> tuple[bool, str]:
+        if config:
+            self.config = BashExecutorConfig(**config)
+        else:
+            self.config = BashExecutorConfig()
```
moshemorad, line 207: If no image or config was provided what is the value of this toolset being enabled? In that case isn't it make will only make calls to fail and impact the actual result?

holmes/config.py
```diff
@@ -54,24 +54,55 @@ class SupportedTicketSources(str, Enum):
...
-        msg = f"ENV var replacement {env_var_key} does not exist for param: {value}"
-        logging.error(msg)
-        raise Exception(msg)
+    env_patterns = re.findall(r"{{\s*env\.([^}]*)\s*}}", value)
+
+    # Find all non-env patterns ({{ anything_else }})
+    all_patterns = re.findall(r"{{\s*([^}]*)\s*}}", value)
+    non_env_patterns = [p for p in all_patterns if not p.strip().startswith("env.")]
+
+    if non_env_patterns and not env_patterns:
```
moshemorad: Can you elaborate what is this use case? e.g. what is non_env_patterns that we are fine with keeping? when we have such a thing?

holmes/config.py
```diff
@@ -54,24 +54,55 @@ class SupportedTicketSources(str, Enum):
...
+        return ""
+
+    if not env_patterns and not non_env_patterns:
+        return value
+
+    # If the entire string is just one env pattern, handle it specially (should raise on missing)
+    full_pattern_match = re.match(r"^\s*{{\s*env\.([^}]*)\s*}}\s*$", value)
+    if full_pattern_match:
```
moshemorad: Why we need this special case? isn't the general match and replace will be good enough?

.gitignore
```diff
@@ -162,7 +162,6 @@ playwright.png
 .deepeval*
 pyrightconfig.json
 
-*.AUTOGENERATED
```
moshemorad, line 165: Why we removed it? since presented in the ineer gitignore?

tests/plugins/toolsets/bash/test_unsafe_commands.py
```diff
@@ -0,0 +1,241 @@
...
+        """Test that dangerous commands raise the expected exception when processed."""
+        match = (
+            re.escape(partial_error_message_content)
+            if partial_error_message_content
+            else None
+        )
+        with pytest.raises(expected_exception, match=match):
+            make_command_safe(dangerous_command, config=TEST_CONFIG)
+
+    def test_mixed_safe_and_unsafe_commands_raise_error(self):
```
moshemorad: Can also add cases for running with &&?

tests/plugins/toolsets/bash/test_safe_commands.py
```diff
@@ -0,0 +1,303 @@
...
+                "kubectl get pods --namespace default",
+            ),
+            # Quote normalization in grep
+            ('kubectl get pods | grep "nginx"', "kubectl get pods | grep nginx"),
+            ("kubectl get pods | grep 'nginx'", "kubectl get pods | grep nginx"),
+        ],
+    )
+    def test_kubectl_round_trip(self, input_command: str, expected_output: str):
+        output_command = make_command_safe(input_command, config=TEST_CONFIG)
+        print(f"* EXPECTED:\n{expected_output}")
```
moshemorad: Please replace to logging

tests/llm/test_ask_holmes.py
```diff
@@ -164,7 +181,9 @@ def ask_holmes(test_case: AskHolmesTestCase) -> LLMResult:
             expected_tools.append(tool_mock.tool_name)
 
     tool_executor = ToolExecutor(mock.enabled_toolsets)
+    enabled_toolsets = [t.name for t in tool_executor.enabled_toolsets]
 
+    print(f"** ENABLED TOOLSETS **\n{', '.join(enabled_toolsets)}")
```
moshemorad: please remove

tests/llm/utils/mock_toolset.py
```diff
@@ -56,6 +66,9 @@ def _get_mock_file_path(self):
 
     def _auto_generate_mock_file(self, params: Dict):
         mock_file_path = self._get_mock_file_path()
+
+        mock_file_path = deconflict_file_name(mock_file_path)
```
moshemorad: Can we make the logic from make it eval reliable? or we will just override it when we will merge it?

tests/llm/utils/mock_toolset.py
```diff
@@ -149,6 +162,7 @@ def _load_toolsets_definitions(self, run_live) -> List[ToolsetYamlFromConfig]:
             toolsets_definitions = parse_toolsets_file(
                 path=config_path, raise_error=run_live
             )
+            print(f"** ** {toolsets_definitions}")
```
moshemorad: Please remove

tests/llm/fixtures/test_ask_holmes/42_dns_issues_steps_old_tools/test_case.yaml
```diff
@@ -0,0 +1,27 @@
...
+  The answer MUST NOT tell the user to do any of the above actions as next steps but should have run these actions itself.
+  The answer MUST include confirm it tested DNS resolution from a pod and it showed DNS issues
+before_test: |
+  kubectl apply -f./manifest.yaml
+  sleep 10
+after_test: kubectl delete -f./manifest.yaml
+
+evaluation:
+  correctness:
+    expected_score: 0
```
moshemorad, line 23: So just to make sure i understand correctly. we know that this eval doesn't work and we want still merge it?

moshemorad (changes requested):

holmes/plugins/toolsets/bash/bash_instructions.jinja2
nherment: I want to keep it as is for now. We can address these questions once it comes to adding more bash capabilities to Holmes.

holmes/plugins/toolsets/bash/kubectl/kubectl_events.py
nherment: @mainred resolving this, please reopen if needed.

tests/llm/fixtures/test_ask_holmes/42_dns_issues_result_new_tools/toolsets.yaml
```diff
@@ -0,0 +1,20 @@
...
+          - image: busybox
+            allowed_commands:
+              - "cat .*"
+              - "nslookup .*"
+              - "sh .*"
+          - image: "registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3"
+            allowed_commands:
+              - "cat .*"
+              - "nslookup .*"
+              - "sh .*"
```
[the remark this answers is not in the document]
nherment: I prefer the test to stay simple

holmes/main.py
```diff
@@ -1,5 +1,6 @@
 # ruff: noqa: E402
 import os
+from holmes.core.prompt import append_file_to_user_prompt
```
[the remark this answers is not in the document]
nherment: fixed.

tests/llm/fixtures/test_ask_holmes/42_dns_issues_steps_new_all_tools/dns_troubleshooting_instructions.md
```diff
@@ -0,0 +1,67 @@
...
+*   "Pods in namespace 'secure-ns' cannot resolve any hostnames. `/etc/resolv.conf` in these pods is missing the correct `nameserver` entry. This is likely due to a misconfiguration in the pod's `dnsPolicy` or the underlying node's DNS setup."
+*   "External DNS resolution is failing cluster-wide. The CoreDNS ConfigMap shows that the `forward` plugin is pointing to an incorrect upstream DNS server IP address."
+*   "DNS lookups for 'service-a.namespace-b' are timing out. A NetworkPolicy in 'namespace-b' is blocking egress traffic on port 53 to the kube-dns service."
+
+## Recommend Remediation Steps (Based on Docs)
+*   **CRITICAL:** ALWAYS refer to the official Kubernetes DNS debugging guide for detailed troubleshooting and solutions:
+    *   Main guide: https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/
+    *   CoreDNS specific: https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/ (for CoreDNS customization which might be relevant)
```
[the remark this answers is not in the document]
nherment: current indentation is correct

holmes/plugins/toolsets/bash/parse_command.py
```diff
@@ -0,0 +1,99 @@
...
+            for cmd in safe_commands
+        ]
+
+        return " | ".join(safe_commands_str)
+    except SystemExit:
+        # argparse throws a SystemExit error when it can't parse command or arguments
+        # This ideally should be captured differently by ensuring all possible args
+        # are accounted for in the implementation for each command.
+        # When falling back, we raise a generic error
+        raise ValueError("The command failed to be parsed for safety")
```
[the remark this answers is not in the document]
nherment: addressed.

tests/llm/utils/mock_utils.py
```diff
@@ -225,3 +238,15 @@ def load_investigate_request(test_case_folder: Path) -> InvestigateRequest:
...
+def load_include_files(
+    test_case_folder: Path, include_files: Optional[list[str]]
+) -> str:
+    extra_prompt: str = ""
+    if include_files:
+        for file_path_str in include_files:
+            file_path = Path(test_case_folder.joinpath(file_path_str))
+            extra_prompt = append_file_to_user_prompt(extra_prompt, file_path)
+
+    return extra_prompt
```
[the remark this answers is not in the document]
nherment: python is >= 3.10. Ignoring this

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -0,0 +1,303 @@
...
+            )
+        try:
+            safe_command_str = make_command_safe(command_str, self.toolset.config)
+        except (argparse.ArgumentError, ValueError) as e:
+            logging.info(f"Refusing LLM tool call {command_str}", exc_info=True)
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=f"Refusing to execute bash command. Only some commands are supported and this is likely because requested command is unsupported. Error: {str(e)}",
+                params=params,
+            )
```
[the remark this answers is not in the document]
nherment: Changed to `error=` across the board

holmes/plugins/toolsets/bash/kubectl/kubectl_get.py
nherment: This will be passed through to the actual kubectl which will throw return an error

holmes/plugins/toolsets/bash/kubectl/__init__.py
nherment: This validates common args

nherment: > But there we aren't validaing the len so maybe we would like to also add this validation there?

I think less is more. If it can cause a security issue then we should validate but otherwise I don't think we should

holmes/plugins/toolsets/bash/kubectl/kubectl_run.py
nherment: Because we only know if this is needed at runtime. It's good information to bubble up to the LLM which can then inform the user that they can configure the toolset to allow running images.

holmes/plugins/toolsets/bash/bash_toolset.py
nherment: I have been wanting to be able to do so for a long time. The current design of toolsets and tools mean they are initialized before the config is set.
It is possible to reset the params once the config is set but I think the correct approach is to refactor how Toolsets are initialized and configured. There is a whole lot of ugly code that is caused by this obsolete design.

I don't think hacking a solution to do this is worth the effort.

holmes/plugins/toolsets/bash/bash_toolset.py
nherment: > Don't we want to seperate the stdout & stderr from each other and take only the relevant output?

We should replicate what a user/llm would see if executing the command in a shell. Often, stdout and stderr are intermingled and the relative positioning of each can be a helpful indicator for debugging. stderr is often (if not always) clearly identified as such by the meaning of the text within it.

> can we also make sure to return no_data in relevant cases?

We likely should because the SaaS UI relies on it. I think it's potentially hard and unreliable to implement for all use cases. I'll add it for the simple one where the return string is empty and not an error.

holmes/plugins/toolsets/bash/bash_toolset.py
nherment: yes, good catch!

holmes/plugins/toolsets/bash/bash_toolset.py
nherment: Indeed.

holmes/plugins/toolsets/bash/bash_toolset.py
nherment: It'll only fail for `kubectl run ...` and then will inform the user (through the output) that the toolset needs to be configured. My understanding is that this is inline with previous decision to ask the user to configure holmes if something is not working as expected. It's the reason it's been designed that way.

holmes/config.py
nherment: I changed the specs and simplified the code. There was a bunch of code that was removing any double curly braces unless they were a well formed `{{ env.... }}` string.
Now they are left as-is and only the correct strings are replaced.

holmes/config.py
nherment: yes it is. Code is simpler now

.gitignore
nherment: replaced with `tests/llm/fixtures/.gitignore`

tests/plugins/toolsets/bash/test_unsafe_commands.py
nherment: I added a negative test case for &&

tests/llm/test_ask_holmes.py
nherment: This is useful.

tests/plugins/toolsets/bash/test_safe_commands.py
nherment: This is useful. I'll comment it out.

tests/llm/utils/mock_toolset.py
nherment: I adoped the reliable evals logic as it's better

tests/llm/fixtures/test_ask_holmes/42_dns_issues_steps_old_tools/test_case.yaml
nherment: The eval works. `expected_score: 0` just means Holmes is not always correct.

holmes/plugins/toolsets/bash/parse_command.py
```diff
@@ -0,0 +1,103 @@
...
+        if part == "&&":
+            raise ValueError(
+                'Double ampersand "&&" is not a supported way to chain commands. Run each command separately.'
+            )
+        else:
+            current_command.append(part)
+
+    if current_command:
+        commands_list.append(current_command)
+
```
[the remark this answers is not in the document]
nherment: I fixed the code.

holmes/plugins/toolsets/bash/common/stringify.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Escape shell arguments to prevent injection.
+    Uses shlex.quote for safe shell argument quoting.
+    """
+    escaped_args = []
+
+    for arg in args:
+        # If argument is safe (contains only alphanumeric, hyphens, dots, underscores, equals, slash, comma, colon)
+        # no escaping needed
+        if arg and all(c.isalnum() or c in SAFE_SHELL_CHARS for c in arg):
```
moshemorad, line 16: Isn't shlex.quote already doing this logic as well? 
I see that it search for safe character.

tests/llm/fixtures/test_ask_holmes/42_dns_issues_steps_old_tools/test_case.yaml
moshemorad: It keeps confusing me. sorry

holmes/plugins/toolsets/bash/bash_toolset.py
moshemorad: Not fixed?

holmes/plugins/toolsets/bash/common/stringify.py
nherment: Yes, it's a bit duplicated.

PR #445, opened 2025-05-23
Cli logging improvements

aantn:
This cleans up cli logs a bit. It involves some flow changes as well to toolset configuration, so it should be carefully reviewed and tested. I did a sanity check on the cli, but did not check the in-cluster server behaviour.


holmes/core/tools.py
```diff
@@ -399,6 +401,8 @@ def interpolate_command(self, command: str) -> str:
         return interpolated_command
 
     def check_prerequisites(self):
+        self._status = ToolsetStatusEnum.ENABLED
```
aantn, line 404: Please check this carefully! I *think* it's correct, but I don't have a full understanding of how toolset.enabled interacts with toolset._status (I think enabled indicates the user intent and _status is the result of the checks so this is fine.) I also only tested with the local cli not the in-cluster path

arikalon1: @aantn can you explain why you added it?

aantn: `self._status` starts as disabled, so without this the function returns after the first iteration due to the check at the end.

arikalon1 (approved):
great improvement, thank you

arikalon1 (dismissed):

arikalon1 (approved):

PR #447, opened 2025-05-26
Update aks-node-health.yaml

aritraghosh:
Adding and changing a few things in the aks-node health toolset


aritraghosh:
@mainred  Can you review

holmes/plugins/toolsets/aks-node-health.yaml
```diff
@@ -20,23 +20,23 @@ toolsets:
...
 
-      - name: "check_kubelet_logs"
-        description: "Fetches kubelet logs from a specific node in the AKS cluster."
-        user_description: "get kubelet logs for node {{ NODE_NAME }}"
+      - name: "get_node_events"
+        description: "Fetches recent events for a specific node to surface warnings and errors."
+        user_description: "get events for node {{ NODE_NAME }}"
         command: |
-          ssh {{ NODE_NAME }} 'journalctl -u kubelet'
+          kubectl get events --field-selector involvedObject.kind=Node,involvedObject.name={{ NODE_NAME }} --sort-by='.lastTimestamp'
```
mainred, line 27: I think it's more like a runbook instead of tools in toolsets. 
A better practice for the aks node health case I think is we introduce a prompt/runbook like the dns trouble guide we have, and leverage the existing tools if we have, or introduce new tools to get resources.

https://github.com/robusta-dev/holmesgpt/blob/5cc6a34c18ac548e0bfa5d6a80213bbb5baebc4c/holmes/plugins/toolsets/kubernetes.yaml#L41-L43

holmes/plugins/toolsets/aks-node-health.yaml
```diff
@@ -20,23 +20,23 @@ toolsets:
...
 
-      - name: "check_control_plane_connectivity"
-        description: "Verifies if the node can communicate with the control plane."
-        user_description: "check control plane connectivity for node {{ NODE_NAME }}"
+      - name: "check_node_resource_usage"
+        description: "Shows CPU/memory usage for a specific node (requires metrics-server)."
+        user_description: "get resource usage for node {{ NODE_NAME }}"
         command: |
-          kubectl exec {{ NODE_NAME }} -- ping <control-plane-ip>
+          kubectl top node {{ NODE_NAME }}
```
mainred, line 33: top node is also support but not with node name filter, we can ` kubectl top node {{ NODE_NAME }}` in the same toolset as the existing kubectl_top_nodes

https://github.com/robusta-dev/holmesgpt/blob/5cc6a34c18ac548e0bfa5d6a80213bbb5baebc4c/holmes/plugins/toolsets/kubernetes.yaml#L162-L165

holmes/plugins/toolsets/aks-node-health.yaml
mainred: what do you think @arikalon1

aritraghosh:
Closing this one based on discussion with @pavneeta , he will create a new one to replace this

cc @mainred

PR #459, opened 2025-05-28
feat: cache toolset status and add toolset management tool command

mainred:
- use isort to sort the imported packaged
- introduce toolset_manager to manage toolsets, which was Config's job.
- only enable build-in toolsets specified in the config file
- use load_toolsets_config as the only entrypoint to load initialize and validate a toolset from a config/definition
- introduce cli command to list and refresh the toolset status from local cache

```
 ./dist/holmes/holmes toolset --help

 Usage: holmes toolset [OPTIONS] COMMAND [ARGS]...

 toolset management commands

╭─ Options ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help          Show this message and exit.                                                                                                           │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ list      List build-int and custom toolsets status of CLI                                                                                            │
│ refresh   Refresh build-in and custom toolsets status of CLI                                                                                          │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

```
./dist/holmes/holmes toolset list
                                                   Toolset Status
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━┓
┃ Name                             ┃ Status   ┃ Enabled ┃ Type       ┃ Path                                ┃ Error ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━┩
│ confluence                       │ disabled │ False   │ built-in   │ None                                │ None  │
│ docker/core                      │ disabled │ False   │ built-in   │ None                                │ None  │
│ helm/core                        │ disabled │ False   │ built-in   │ None                                │ None  │
│ kubernetes/logs                  │ disabled │ False   │ built-in   │ None                                │ None  │
│ servicenow                       │ disabled │ False   │ built-in   │ None                                │ None  │
│ aks/node-health                  │ disabled │ False   │ built-in   │ None                                │ None  │
│ aws/security                     │ disabled │ False   │ built-in   │ None                                │ None  │
│ aws/rds                          │ disabled │ False   │ built-in   │ None                                │ None  │
│ argocd/core                      │ disabled │ False   │ built-in   │ None                                │ None  │
│ slab                             │ disabled │ False   │ built-in   │ None                                │ None  │
│ aks/core                         │ disabled │ False   │ built-in   │ None                                │ None  │
│ kubernetes/core                  │ enabled  │ True    │ built-in   │ None                                │ None  │
│ kubernetes/live-metrics          │ disabled │ False   │ built-in   │ None                                │ None  │
│ kubernetes/kube-prometheus-stack │ disabled │ False   │ built-in   │ None                                │ None  │
│ kubernetes/krew-extras           │ disabled │ False   │ built-in   │ None                                │ None  │
│ kubernetes/kube-lineage-extras   │ disabled │ False   │ built-in   │ None                                │ None  │
│ internet                         │ disabled │ False   │ built-in   │ None                                │ None  │
│ robusta                          │ disabled │ False   │ built-in   │ None                                │ None  │
│ opensearch/status                │ disabled │ False   │ built-in   │ None                                │ None  │
│ grafana/loki                     │ disabled │ False   │ built-in   │ None                                │ None  │
│ grafana/tempo                    │ disabled │ False   │ built-in   │ None                                │ None  │
│ newrelic                         │ disabled │ False   │ built-in   │ None                                │ None  │
│ grafana/grafana                  │ disabled │ False   │ built-in   │ None                                │ None  │
│ notion                           │ disabled │ False   │ built-in   │ None                                │ None  │
│ kafka/admin                      │ disabled │ False   │ built-in   │ None                                │ None  │
│ datadog                          │ disabled │ False   │ built-in   │ None                                │ None  │
│ prometheus/metrics               │ disabled │ False   │ built-in   │ None                                │ None  │
│ datetime                         │ enabled  │ True    │ built-in   │ None                                │ None  │
│ opensearch/logs                  │ disabled │ False   │ built-in   │ None                                │ None  │
│ opensearch/traces                │ disabled │ False   │ built-in   │ None                                │ None  │
│ coralogix/logs                   │ disabled │ False   │ built-in   │ None                                │ None  │
│ rabbitmq/core                    │ disabled │ False   │ built-in   │ None                                │ None  │
│ git                              │ disabled │ False   │ built-in   │ None                                │ None  │
│  ig/core                          │ enabled  │ True    │ customized │ /home/azureuser/inspect_gadget.yaml │ None  │
```


holmes/core/toolset_manager.py
```diff
@@ -0,0 +1,314 @@
...
+            if toolsets_from_config:
+                self.add_or_merge_yaml_toolsets(
+                    toolsets_from_config,
+                    toolsets_by_name,
+                )
+
+        # custom toolset should not override built-in toolsets
+        # to test the new change of built-in toolset, we should make code change and re-compile the program
+        custom_toolsets = self.load_custom_toolsets()
+        for custom_toolset in custom_toolsets:
```
mainred: server is using custom_toolset to enable built-in toolsets and add extra customized toolset.
We need to consider add custom config toolset overriding built-in toolset back to not block server.

https://github.com/robusta-dev/holmesgpt/blob/master/helm/holmes/templates/toolset-config.yaml

holmes/core/toolset_manager.py
```diff
@@ -0,0 +1,314 @@
...
+            for _, toolset_config in self.toolsets.items():
+                logging.debug(f"Processing toolset config: {toolset_config}")
+                toolset_config["enabled"] = self.get_toolset_definition_enabled(
+                    toolset_config
+                )
+            toolsets_from_config = load_toolsets_config(
+                toolsets=self.toolsets, strict_check=False
+            )
+            if toolsets_from_config:
+                self.add_or_merge_yaml_toolsets(
```
mainred: There's some misunderstanding and less documented part for tool_config.
From the original code, both toolset and custom_toolset can override built-in toolset and introduce new customized toolset, which seems to be redundant.

holmes/config.py
```diff
@@ -205,6 +148,32 @@ def model_post_init(self, __context: Any) -> None:
...
+        cli_options = {k: v for k, v in kwargs.items() if v is not None and v != []}
+
+        if config_from_file is None:
+            return cls(**cli_options)
+
+        logging.debug(f"Overriding config from cli options {cli_options}")
+        merged_config = config_from_file.dict()
+        merged_config.update(cli_options)
+        return cls(**merged_config)
+
```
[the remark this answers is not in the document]
mainred: We rely on config_from_file  from the input to decide the following actions instead of overriding its value to None

holmes/core/tools.py
```diff
@@ -118,6 +94,8 @@ class ToolsetTag(str, Enum):
 
 
 class ToolsetType(str, Enum):
+    BUILTIN = "built-in"
+    CUSTOMIZED = "customized"
```
moshemorad: Can we please change the terminology to custom instead?

holmes/core/tools.py
```diff
@@ -345,14 +326,15 @@ class Toolset(BaseModel):
...
 
     # warning! private attributes are not copied, which can lead to subtle bugs.
     # e.g. l.extend([some_tool]) will reset these private attribute to None
 
-    _path: Optional[str] = PrivateAttr(None)
-    _status: ToolsetStatusEnum = PrivateAttr(ToolsetStatusEnum.DISABLED)
-    _error: Optional[str] = PrivateAttr(None)
+    # status fields that be cached
+    type: ToolsetType = ToolsetType.BUILTIN
+    path: Optional[FilePath] = None
```
moshemorad, line 335: Missed rename in line 492

holmes/core/tools.py
```diff
@@ -345,14 +326,15 @@ class Toolset(BaseModel):
...
-    type: Optional[ToolsetType] = None
 
     # warning! private attributes are not copied, which can lead to subtle bugs.
     # e.g. l.extend([some_tool]) will reset these private attribute to None
 
-    _path: Optional[str] = PrivateAttr(None)
-    _status: ToolsetStatusEnum = PrivateAttr(ToolsetStatusEnum.DISABLED)
-    _error: Optional[str] = PrivateAttr(None)
+    # status fields that be cached
+    type: ToolsetType = ToolsetType.BUILTIN
```
moshemorad: Are we sure that we want to provide BUILTIN as default value? I think that this should be always provided so maybe it would be better to not have a default for that?

holmes/core/tools.py
```diff
@@ -477,6 +447,16 @@ def _load_llm_instructions(self, jinja_template: str):
...
+    # YamlToolset is loaded from a YAML file specified by the user and should be enabled by default
+    # Built-in toolsets are exception and should be disabled by default when loaded
+    enabled: bool = True
+    prerequisites: List[
+        Union[
+            StaticPrerequisite,
+            ToolsetCommandPrerequisite,
+            ToolsetEnvironmentPrerequisite,
+        ]
+    ] = []  # type: ignore
```
moshemorad: Why the `type: ignore` is needed here? I tried to removed it but didn't saw any issue so i might missing something.

holmes/main.py
```diff
@@ -52,6 +58,13 @@
     help="Generate new integrations or test data",
 )
 app.add_typer(generate_app, name="generate")
+toolset_app = typer.Typer(
+    add_completion=False,
+    name="toolset",
+    no_args_is_help=True,
+    help="toolset management commands",
```
moshemorad:
```suggestion
    help="Toolset management commands",
```

holmes/core/toolset_manager.py
```diff
@@ -0,0 +1,394 @@
...
+    custom_toolsets_from_cli: Optional[List[FilePath]]
+
+    toolset_status_location: FilePath
+
+    def __init__(
+        self,
+        toolsets: Optional[dict[str, dict[str, Any]]] = None,
+        custom_toolsets: Optional[List[FilePath]] = None,
+        custom_toolsets_from_cli: Optional[List[FilePath]] = None,
+        toolset_status_location: FilePath = DEFAULT_TOOLSET_STATUS_LOCATION,
```
moshemorad: Default value doesn't aligned with the type hint of the var.

holmes/main.py
```diff
@@ -126,19 +139,15 @@ def init_logging(verbose_flags: Optional[List[bool]] = None):
 )
 opt_model: Optional[str] = typer.Option(None, help="Model to use for the LLM")
 opt_config_file: Optional[Path] = typer.Option(
-    None,
+    DEFAULT_CONFIG_LOCATION,
```
moshemorad: I think that this change causing issue when the file not exists. Tried to run it locally (I don't have the config file) and it result in the following error.

╭─────────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ────────────────────────────────────────────────────────────────────────────────────────╮
│ /Users/moshemorad/robusta/holmesgpt/holmes/main.py:378 in ask                                                                                                                                                    │
│                                                                                                                                                                                                                  │
│    375 │   Ask any question and answer using available tools                                                                                                                                                     │
│    376 │   """                                                                                                                                                                                                   │
│    377 │   console = init_logging(verbose)  # type: ignore                                                                                                                                                       │
│ ❱  378 │   config = Config.load_from_file(                                                                                                                                                                       │
│    379 │   │   config_file,                                                                                                                                                                                      │
│    380 │   │   api_key=api_key,                                                                                                                                                                                  │
│    381 │   │   model=model,                                                                                                                                                                                      │
│                                                                                                                                                                                                                  │
│ /Users/moshemorad/robusta/holmesgpt/holmes/config.py:169 in load_from_file                                                                                                                                       │
│                                                                                                                                                                                                                  │
│   166 │   │                                                                                                                                                                                                      │
│   167 │   │   cli_options = {k: v for k, v in kwargs.items() if v is not None and v != []}                                                                                                                       │
│   168 │   │                                                                                                                                                                                                      │
│ ❱ 169 │   │   if config_from_file is None:                                                                                                                                                                       │
│   170 │   │   │   return cls(**cli_options)                                                                                                                                                                      │
│   171 │   │                                                                                                                                                                                                      │
│   172 │   │   logging.debug(f"Overriding config from cli options {cli_options}")                                                                                                                                 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
UnboundLocalError: cannot access local variable 'config_from_file' where it is not associated with a value

holmes/main.py
```diff
@@ -319,9 +328,8 @@ def ask(
     # common options
     api_key: Optional[str] = opt_api_key,
     model: Optional[str] = opt_model,
-    config_file: Optional[str] = opt_config_file,  # type: ignore
+    config_file: Optional[Path] = opt_config_file,  # type: ignore
```
moshemorad: Is the `type: ignore` still needed?

holmes/main.py
```diff
@@ -319,9 +328,8 @@ def ask(
...
     custom_toolsets: Optional[List[Path]] = opt_custom_toolsets,
-    allowed_toolsets: Optional[str] = opt_allowed_toolsets,
```
moshemorad, line 325: This is a change in behivour. Until now by definition we enabled user to decide which tools are enabled through arg, now it means that we only change enabled tools by config file. Is that intended?

holmes/config.py
```diff
@@ -170,7 +96,12 @@ class Config(RobustaBaseConfig):
     opsgenie_query: Optional[str] = None
 
     custom_runbooks: List[FilePath] = []
-    custom_toolsets: List[FilePath] = []
+
+    # custom_toolsets is defined in the config file, and can override built-in toolsets as stable custom toolsets
```
moshemorad: Can you please elaborate with examples? not sure I fully understand the use case here. 
When running from CLI is there a flow that i would have both of the vars configured? if so can you please explain the use case?

tests/plugins/toolsets/test_load_config.py
```diff
@@ -0,0 +1,93 @@
...
+    "GRAFANA_URL": "https://my-grafana.com/",
+}
+
+
+def test_load_toolsets_from_config():
+    original_env = os.environ.copy()
+
+    try:
+        for key, value in env_vars.items():
+            os.environ[key] = value
```
moshemorad: Please consider using monkeypatch.setenv or mock.patch.dict

holmes/plugins/toolsets/__init__.py
```diff
@@ -81,13 +77,79 @@ def load_python_toolsets(dal: Optional[SupabaseDal]) -> List[Toolset]:
...
+    """
+
+    if not toolsets:
+        return []
+
+    loaded_toolsets: list[Toolset] = []
+    if is_old_toolset_config(toolsets):
+        message = "Old toolset config format detected, please update to the new format: https://docs.robusta.dev/master/configuration/holmesgpt/custom_toolsets.html"
+        logging.warning(message)
+        raise ValueError(message)
```
moshemorad, line 127: Im not sure that we want to change this behivour yet. @arikalon1 WDYT? can we deprecate it already?

holmes/core/toolset_manager.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+class ToolsetManager:
+    """
+    ToolsetManager is responsible for managing toolset locally.
+    It can refresh the status of all toolsets and cache the status to a file.
+    It also provides methods to get toolsets by name and to get the list of all toolsets.
+    """
+
+    toolsets: Optional[dict[str, dict[str, Any]]]
```
moshemorad: Since this isn't paydantic object those are class level variable. Currently their are unassigned so won't be part of the class. But it could cause confusion. Is there a case that we would like to have those vars in the class level? if not please consider to remove.

holmes/core/toolset_manager.py
```diff
@@ -0,0 +1,394 @@
...
+        if os.path.isfile(CUSTOM_TOOLSET_LOCATION):
+            if self.custom_toolsets is None:
+                self.custom_toolsets = []
+            self.custom_toolsets.append(CUSTOM_TOOLSET_LOCATION)
+
+        self.custom_toolsets_from_cli = custom_toolsets_from_cli
+        self.toolset_status_location = toolset_status_location
+
+    @property
+    def cli_tool_tags(self) -> List[str]:
```
moshemorad: Please consider to return set instead of list since used for in checking.

holmes/core/toolset_manager.py
```diff
@@ -0,0 +1,394 @@
...
+            dal, refresh_status=refresh_status
+        )
+        return [
+            ts
+            for ts in toolsets_with_status
+            if any(tag in self.cli_tool_tags for tag in ts.tags)
+        ]
+
+    # TODO(mainred): cache and refresh periodically toolset status for server if necessary
+    def list_enabled_server_toolsets(
```
moshemorad: Looks like for some reason when running with this version on my local cluster the datetime toolset isn't being enabled by default.

holmes/core/toolset_manager.py
```diff
@@ -0,0 +1,394 @@
...
+        for custom_toolset_from_cli in custom_toolsets_from_cli:
+            if custom_toolset_from_cli in toolsets_status_by_name.keys():
+                raise Exception(
+                    f"Toolset {custom_toolset_from_cli.name} from cli is already defined in existing toolset"
+                )
+        all_toolsets_with_status.extend(custom_toolsets_from_cli)
+
+        return all_toolsets_with_status
+
+    def list_enabled_console_toolsets(
```
moshemorad: The current implementation cause it to disable all toolset by default.
This is different behiour then our initial intantion. We would like that every new user will have toolsets that are enabled by default to start and get the hand of holmes before messing with toolsets config. This is quite a change and i think that we need to be aligned on that.
@arikalon1 WDYT?

moshemorad (changes requested):
There are few changes that effect the default user expiriance. we need to align on that.

holmes/core/tools.py
mainred: yeah, in the code path, we always set the toolset type. I can make it optional, and let the loading toolset part set the toolset type.

holmes/core/tools.py
mainred: Do you know how can we detect this error from any lint?

holmes/core/toolset_manager.py
mainred: We don't recommend using list for in checking?

holmes/core/tools.py
mainred: Good point, it's an over-code here.

holmes/core/toolset_manager.py
mainred: Thanks for sharing the intention of the design.

My intention is to disable all tools by default, and customer has to enable the tools explicitly. This allows us to reduce the chance the loading unprepared toolset. But since we have toolset management cli, I think we can enable  them by default, and let toolset management command to manage the lifecycle of the toolsets.

holmes/core/toolset_manager.py
mainred: Yes, I disabled all built-in toolset by default.

holmes/main.py
mainred: ~I can  change the code to make it make the configurable optional~

mainred: It's because I used the var before it's initialized. Will create the config_from_file first

holmes/plugins/toolsets/__init__.py
mainred: I did not change the behavior by following https://github.com/robusta-dev/holmesgpt/blob/master/holmes/config.py#L73

mainred: I think it's good to remove it if the format is not supported anymore after we announce deprecating it several releases.

holmes/core/tools.py
moshemorad: I added mypy to pre commit and it should catch those. I will try to understand why it isn't raising error on that.

mainred:
Had call with @moshemorad, the notes from the call

```
print toolset is load from cache during ask command
fix the toolset table issue
update the custom-toolset flag description to the toolset is for experimental and not made permanent
add doc to explain the customer-facing changes.
```

mainred:
Could you please take a look at this precommit check failure? where should I add this dependency tabulate? @@moshemorad Thanks in advance.

I have updated the pyproject.toml to include the package.

how should I deal with this build error? [feat: cache toolset status and add toolset management tool command · robusta-dev/holmesgpt@b505f81](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459)
```
holmes/main.py:5: error: Library stubs not installed for "tabulate"  [import-untyped]
[161](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459#step:4:166)
holmes/main.py:5: note: Hint: "python3 -m pip install types-tabulate"
[162](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459#step:4:167)
holmes/main.py:5: note: (or run "mypy --install-types" to install all missing stub packages)
[163](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459#step:4:168)
holmes/main.py:5: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
[164](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459#step:4:169)
Found 1 error in 1 file (checked 85 source files)
```

holmes/core/tools.py
moshemorad: Pushed a fix for mypy to master that should catch those kind of things

mainred:
I have resolved code conflicts multiple times since my change touch the config.py. I'll focus on the PR comments if there are incoming ones, and until all looks good, I'll resolve the conflicts.
cc @moshemorad

moshemorad:
> Could you please take a look at this precommit check failure? where should I add this dependency tabulate? @@moshemorad Thanks in advance.
> 
> I have updated the pyproject.toml to include the package.
> 
> how should I deal with this build error? [feat: cache toolset status and add toolset management tool command · robusta-dev/holmesgpt@b505f81](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459)
> 
> ```
> holmes/main.py:5: error: Library stubs not installed for "tabulate"  [import-untyped]
> [161](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459#step:4:166)
> holmes/main.py:5: note: Hint: "python3 -m pip install types-tabulate"
> [162](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459#step:4:167)
> holmes/main.py:5: note: (or run "mypy --install-types" to install all missing stub packages)
> [163](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459#step:4:168)
> holmes/main.py:5: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
> [164](https://github.com/robusta-dev/holmesgpt/actions/runs/15464595950/job/43533650601?pr=459#step:4:169)
> Found 1 error in 1 file (checked 85 source files)
> ```

Currently add `# type: ignore`

moshemorad:
> I have resolved code conflicts multiple times since my change touch the config.py. I'll focus on the PR comments if there are incoming ones, and until all looks good, I'll resolve the conflicts. cc @moshemorad

The fix is due to the missed error when we changed private attr in the Toolset Model. So it might raise more issue. Let me know if you bump into anything that need my help.

mainred:
> Let me know if you bump into anything that need my help.

This PR is ready for review. Please take another look. Thanks.

mainred:
Please ignore the conflicts lets focus on the PR change first. I will definitely resolve the conflicts after addressing all comments to reduce the chance of resolving conflicts.

holmes/config.py
mainred: re-phrase the comments for custom_toolsets and custom_toolsets_from_cli, please take a look.

mainred:
@moshemorad could you please help find the root cause of llm evaluation error. The following error message raised from the pipeline seems to be common issue? Thanks.
`httpx.LocalProtocolError: Illegal header value b'Bearer '

moshemorad:
> @moshemorad could you please help find the root cause of llm evaluation error. The following error message raised from the pipeline seems to be common issue? Thanks. `httpx.LocalProtocolError: Illegal header value b'Bearer '

Looks like for some reason it isn't adding the OpenAI key, trying to understand why.

moshemorad:
@mainred Looks like the issue is due to forked repos don't use the original repo secrets.
https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

@arikalon1 @nherment any idea what we want to do in those cases?

holmes/core/tools.py
```diff
@@ -323,7 +301,10 @@ class ToolsetEnvironmentPrerequisite(BaseModel):
 class Toolset(BaseModel):
     model_config = ConfigDict(extra="forbid")
     experimental: bool = False
-    enabled: bool = False
+
+    # Built-in toolset is enabled when 1. it's specified in the config and 2. 'enabled' is not set to false
+    # Custom toolset is enabled when 'enabled' is not set to false
+    enabled: bool = True
```
moshemorad: This change causing to scenario when we are trying to load a lot of irrelevant tools in the server and most of the in error state. 
image

We need to make sure that in server it will be as today.
image

moshemorad (changes requested):

holmes/core/tools.py
moshemorad: Fixed and pushed

moshemorad (approved):

PR #467, opened 2025-05-30
Cleaning up the AKS-node-health toolset

pavneeta:
1. Cleans up the description vs the command naming
2. removes the vmss commands since they don't work without machine ID
3. added top resource commands.


pavneeta:
replaces this PR since there were conflicts with another PR, and the VMSS commands would not work - https://github.com/robusta-dev/holmesgpt/pull/414

arikalon1 (approved):
Thanks for the PR @pavneeta 

Nice work!

PR #497, opened 2025-06-07
HolmesGPT standalone docs

pavangudiwada:
https://robusta-dev.github.io/holmesgpt/ <- See the docs here


aantn:
TODO:

- [x] Remove redundant markdown pages in github - e.g. https://github.com/robusta-dev/holmesgpt/blob/master/docs/installation.md
- [x] Update links from README.md to go to docs, not redundant github links (which are being removed)
- [ ] Add a file in the docs/ directory like _how_to_build_docs.md - make sure this file is not picked up by mkdocs itself and included in the build
- [ ] Move all docs related files (including mkdocs.yml into docs/ subdir)
- [x] Cleanup all mkdocs warnings

pavangudiwada:
> * [ ]  Move all docs related files (including mkdocs.yml into docs/ subdir)

Can't move it because mkdocs mandates docs to be in a child directory. 

Config value 'docs_dir': The 'docs_dir' should not be the parent directory of the config file. Use a child directory instead so that the 'docs_dir' is a sibling of the config file.

aantn (approved):
Excellent work.

aantn (approved):

PR #500, opened 2025-06-07
fix(test): check classifier model pattern for Azure OpenAI

mainred:
By default, we want to share the llm model used for holmes and classifier, but the model used for Azure OpenAI follows the pattern azure/deployment-name, which cannot be passed to azure_deployment in `openai.AzureOpenAI`.
This PR extracts the deployment name from the model and validates the variable classifier_model follows the name pattern as MODEL


moshemorad (approved):
LGTM

Issue #473, opened 2025-06-03
Pick the right runbooks based on prompt (planning phase)


julia-yin:
### What would you like to be added?

Based on the user's prompt, have a "pre-planning phase" where LLM (or some other algorithm) can pick the right runbook(s) which contain specific troubleshooting steps for different issues/scenarios.

### Why is this needed?

As we are building runbooks to address DNS troubleshooting, node NotReady, pod errors etc. we need a good way to make sure the runbooks used are accurate. The user should not need to specify a runbook to use when running "holmes ask".

aantn:
@julia-yin thanks for creating this! It's probably the most important task on our roadmap!

There are a few possible implementation approaches for this. Should we have a session to discuss? We've been researching what might be most effective and would love to discuss.

mainred:
Fixed by https://github.com/robusta-dev/holmesgpt/pull/547

PR #533, opened 2025-06-17
feat(runbook): llm select the correct runbook from catalog

mainred:
fix: #473

## Changes in this PR:

- allo user to specify customized runbook under `holmes ask`
- introduce a RunbookCatalogManager, which returns runbook per user question by LLM
- RunbookCatalogManager will return custom runbook if any
- Add a runbook catalog, which contains internal runbooks written in markdown

## impacted function

only `holmes ask` CLI is expected to be impacted

## Why use LLM to pick the runbook

The original thought is to introduce a runbook tool to fetch runbook from the runbook catalog folder, but considering we'll have more runbooks to be introduced and selecting the runbook by the either label or tag is limited, introducing LLM to pick the matched one should be the final solution. Also, it's best to select the runbook after customer specify their question, and the following actions should be based on runbook.  With all these in mind, I decided to use LLM to pick the runbook during the early phase of the investigation. This is similar to we have achieved in `holmes investigate`

## Tests

```
➜  holmesgpt git:(runbook-selection) ✗ ./dist/holmes/holmes ask "detect why the k8s pod client under namespace test-ns cannot resolve dns" --max-steps 20
Failed to fetch holmes info                                                                                                          robusta_client.py:23
loading toolset status from cache                                                                                                  toolset_manager.py:216
User: detect why the k8s pod client under namespace test-ns cannot resolve dns
Found runbook for question 'detect why the k8s pod client under namespace test-ns cannot resolve dns':                            tool_calling_llm.py:181
networking/dns_troubleshooting_instructions.md
Running tool kubectl_find_resource: kubectl get -A --show-labels -o wide pod | grep client                                                   tools.py:125
Running tool kubectl_describe: kubectl describe pod client -n test-ns                                                                        tools.py:125
Running tool kubectl_logs_all_containers: kubectl logs client -n test-ns --all-containers                                                    tools.py:125
Token limit exceeded. Truncating tool responses.                                                                                  tool_calling_llm.py:201
Running tool kubectl_describe: kubectl describe pod coredns- -n kube-system                                                                  tools.py:125
Running tool kubectl_describe: kubectl describe service kube-dns -n kube-system                                                              tools.py:125
Running tool kubectl_get_yaml: kubectl get -o yaml pod client -n test-ns                                                                     tools.py:125
Running tool trace_dns_gadget: fetch DNS queries and responses of pod test-ns/client for 30 seconds                                          tools.py:125
Running tool kubectl_get_by_kind_in_namespace: kubectl get --show-labels -o wide networkpolicy -n test-ns                                    tools.py:125
Running tool kubectl_get_yaml: kubectl get -o yaml networkpolicy default-deny-egress -n test-ns                                              tools.py:125
tool_calling_llm.call - completed in 9 iterations - 74078ms                                                                      performance_timing.py:41
AI: Pod client in namespace test-ns cannot resolve DNS because the default-deny-egress NetworkPolicy blocks all outbound traffic, including DNS requests to
CoreDNS (10.0.0.10). No egress rules are defined, so DNS queries never reach the DNS server.

To fix: Add an egress rule to allow UDP/TCP traffic to 10.0.0.10 on port 53 in the default-deny-egress NetworkPolicy.
```


holmes/core/runbooks.py
```diff
@@ -24,3 +28,76 @@ def get_instructions_for_issue(self, issue: Issue) -> List[str]:
...
+        else:
+            logging.debug(f"Runbook link from LLM: {runbookAbsLink}")
+        runbook_folder = get_runbook_folder()
+
+        runbookPath = os.path.join(runbook_folder, runbookAbsLink)
+        try:
+            with open(runbookPath, "r") as file:
+                content = file.read()
+                if len(content.strip()) == 0:
+                    print(f"Warning: The runbook '{runbookPath}' is empty.")
```
nilo19: why do we want to check if the file is empty?

holmes/core/runbooks.py
```diff
@@ -24,3 +28,76 @@ def get_instructions_for_issue(self, issue: Issue) -> List[str]:
...
+When no runbook matches the user question, you should return empty string.
+Here is the catalog of available runbooks:
+{self.catalog.model_dump_json()}.
+                """,
+            },
+            {"role": "user", "content": f"{question}"},
+        ]
+        response = self.ai.completion(messages, temperature=0)
+        # when no runbook matches the user question, llm returns ""
+        runbookAbsLink = response.choices[0].message.content.strip(' "')  # type: ignore
```
nilo19: prefer runbook_abs_link to keep aligned with other variable names

holmes/core/runbooks.py
mainred: Normally the runbook should not be empty, but checking the emptiness of the runbook and logging the warning is to handle exception. What do you think?

PR #534, opened 2025-06-17
Fix server loading tools

moshemorad:


mainred:
Thanks for fixing the issue.

Issue #532, opened 2025-06-17
bash toolset breaks


mainred:
### What happened?

```
Running tool kubectl_run_image: kubectl run <pod_name> --image=busybox --namespace=test-ns --rm --attach --restart=Never -i -- 'cat          tools.py:125
/etc/resolv.conf'
Tool call to kubectl_run_image failed with an Exception                                                                           tool_calling_llm.py:327
Traceback (most recent call last):
  File "holmes/core/tool_calling_llm.py", line 313, in _invoke_tool
    tool_response = tool.invoke(tool_params)
  File "holmes/core/tools.py", line 128, in invoke
    result = self._invoke(params)
  File "holmes/plugins/toolsets/bash/bash_toolset.py", line 94, in _invoke
    validate_image_and_commands(
  File "holmes/plugins/toolsets/bash/kubectl/kubectl_run.py", line 17, in validate_image_and_commands
    if not config or not config.kubectl or not config.kubectl.allowed_images:
AttributeError: 'dict' object has no attribute 'kubectl'
Running tool kubectl_run_image: kubectl run <pod_name> --image=busybox --namespace=test-ns --rm --attach --restart=Never -i -- 'cat          tools.py:125
/etc/resolv.conf; nslookup kube-dns.kube-system.svc.cluster.local; nslookup www.google.com'
Tool call to kubectl_run_image failed with an Exception                                                                           tool_calling_llm.py:327
Traceback (most recent call last):
  File "holmes/core/tool_calling_llm.py", line 313, in _invoke_tool
    tool_response = tool.invoke(tool_params)
  File "holmes/core/tools.py", line 128, in invoke
    result = self._invoke(params)
  File "holmes/plugins/toolsets/bash/bash_toolset.py", line 94, in _invoke
    validate_image_and_commands(
  File "holmes/plugins/toolsets/bash/kubectl/kubectl_run.py", line 17, in validate_image_and_commands
    if not config or not config.kubectl or not config.kubectl.allowed_images:
AttributeError: 'dict' object has no attribute 'kubectl'
``


also 
```
Running tool run_bash_command: 'kubectl get events -A'                                                                       tools.py:125
tool_calling_llm.call - completed in 2 iterations - 2513ms                                                       performance_timing.py:41
AI: Unable to fetch cluster events due to a missing or invalid bash command. Ensure bash is installed and properly configured.      
```

### What did you expect to happen?

All bash tool should be successfully executed.

[...]

PR #535, opened 2025-06-18
fix(toolset): fix toolset config is not initialized when check prerequisites is not called

mainred:
We rely on check prerequisites to initialize the toolset configuration, but when the toolset is loaded from the cache, the prerequisites is not checked, and toolset configuration is not initialized.
Fix this by adding init_config method to initialize the toolset configuration when check prerequisites is not called.
fix: #532

### DIFF
--- a/holmes/core/toolset_manager.py
+++ b/holmes/core/toolset_manager.py
@@ -204,8 +204,10 @@ class ToolsetManager:
         toolset_tags: Optional[List[ToolsetTag]] = None,
     ) -> List[Toolset]:
         """
-        Load the toolset status from the cache file.
-        If the file does not exist, return an empty list.
+        Load the toolset with status from the cache file.
+        1. load the built-in toolsets
+        2. load the custom toolsets from config, and override the built-in toolsets
+        3. load the custom toolsets from CLI, and raise error if the custom toolset from CLI conflicts with existing toolsets
         """
 
         if not os.path.exists(self.toolset_status_location) or refresh_status:
@@ -240,6 +242,13 @@ class ToolsetManager:
                     cached_status.get("type", ToolsetType.BUILTIN)
                 )
                 toolset.path = cached_status.get("path", None)
+            # check prerequisites for only enabled toolset when the toolset is loaded from cache
+            if (
+                toolset.enabled
+                and toolset.status == ToolsetStatusEnum.ENABLED
+                and using_cached
+            ):
+                toolset.check_prerequisites()  # type: ignore
 
         # CLI custom toolsets status are not cached, and their prerequisites are always checked whenever the CLI runs.
         custom_toolsets_from_cli = self._load_toolsets_from_paths(
@@ -253,6 +262,9 @@ class ToolsetManager:
                 raise ValueError(
                     f"Toolset {custom_toolset_from_cli.name} from cli is already defined in existing toolset"
                 )
+            # status of custom toolsets from cli is not cached, and we need to check prerequisites every time the cli runs.
+            custom_toolset_from_cli.check_prerequisites()
+
         all_toolsets_with_status.extend(custom_toolsets_from_cli)
         if using_cached:
             num_available_toolsets = len(
@@ -402,4 +414,5 @@ class ToolsetManager:
             if new_toolset.name in existing_toolsets_by_name.keys():
                 existing_toolsets_by_name[new_toolset.name].override_with(new_toolset)
             else:
+                existing_toolsets_by_name[new_toolset.name] = new_toolset
                 existing_toolsets_by_name[new_toolset.name] = new_toolset


nherment:
Thank you for this @mainred.
There is already a mechanism to initialize the toolsets which consists in calling `check_prerequisites()`. Why not reuse that?

I agree the current approach is not ideal and we want to steer away from it but I also think that adding another function to initialize the config will make toolset initialization more confusing than it already is.

mainred:
check_prerequisites takes time, while our toolset cache is to reduce the loading time taken by the check_prerequisites. So I instead create a init_config method, and assume that the check has been done during the status refresh stage

holmes/plugins/toolsets/git.py
```diff
@@ -68,18 +70,14 @@ def _sanitize_error(self, error_msg: str) -> str:
             return error_msg
         return error_msg.replace(self.git_credentials, "[REDACTED]")
 
-    def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
-        if not config and not (os.getenv("GIT_REPO") and os.getenv("GIT_CREDENTIALS")):
+    def prerequisites_callable(self) -> Tuple[bool, str]:
+        if not self.config and not (
+            os.getenv("GIT_REPO") and os.getenv("GIT_CREDENTIALS")
+        ):
```
nherment: It looks like the env checks belong to `init_config`

holmes/plugins/toolsets/git.py
```diff
@@ -89,6 +87,15 @@ def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
             logging.exception("GitHub prerequisites failed.")
             return False, ""
 
+    def init_config(self):
+        self.git_repo = os.getenv("GIT_REPO") or self.config.get("git_repo")
```
nherment: Should this check if `self.config` exists

nherment:
> check_prerequisites takes time, while our toolset cache is to reduce the loading time taken by the check_prerequisites. So I instead create a init_config method, and assume that the check has been done during the status refresh stage

ok makes sense.

In that case the current change can be improve by making sure the init_config method is idempotent. For example in the case of prometheus, calling `init_config` twice would raise an Exception because the line `self.config = PrometheusConfig(**self.config)` will fail if self.config is already a `PrometheusConfig`.

What about changing `init_config` to a setter by passing the config dict as an argument: `def init_config(config:dict): ...`. This separates the config from the time consuming prereq checks but maintain the current (flawed) logic without introducing new potential bugs.

nherment:
Also note that the pre-commit checks fail

holmes/plugins/toolsets/git.py
mainred: I think I have changed git toolset. It was reverted somehow.

mainred: We can discuss. In prerequisite check,  we check if the config and env is correct. in  init_config, should we also check these prerequisites? Or assume the enabled toolset has meet the requirement.

holmes/plugins/toolsets/git.py
mainred: Yeah, let me add the check.

holmes/plugins/toolsets/git.py
nherment: I think you can assume it has met the requirements because you want to skip the prerequisites liveness checks.
There is some caveats that the config dict may not match the expected pydantic BaseModel and in that case the toolset should go in the proper 'failed' state with a meaningful error message.

mainred:
@nherment could you please take another look?

holmes/plugins/toolsets/internet/internet.py
```diff
@@ -244,17 +244,19 @@ def __init__(
             docs_url=docs_url,
         )
 
-    def prerequisites_callable(self, config: Dict[str, Any]) -> Tuple[bool, str]:
-        if not config:
-            return True, ""
-        self.additional_headers = config.get("additional_headers", {})
+    def prerequisites_callable(self) -> Tuple[bool, str]:
```
nherment: `prerequisites_callable` can be removed if it's not checking anything

holmes/core/tools.py
```diff
@@ -448,6 +448,15 @@ def check_prerequisites(self):
     def get_example_config(self) -> Dict[str, Any]:
         return {}
 
+    @abstractmethod
+    def init_config(self):
+        """
+        Initialize the toolset configuration.
+        CallablePrerequisite can be used to initialize the configuration of a toolset, and validate the configuration prerequisites,
+        but when prerequisites check is not required, use init_config to initialize the configuration.
```
nherment: This comment seems incorrect?
`init_config` should always be used to initialize and validate the config.

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -200,9 +199,12 @@ def _reload_llm_instructions(self):
...
-        if config:
-            self.config = BashExecutorConfig(**config)
+    def prerequisites_callable(self) -> tuple[bool, str]:
+        self.init_config()
+        return True, ""
+
+    def init_config(self):
+        if self.config:
+            self.config = BashExecutorConfig(**self.config)
         else:
```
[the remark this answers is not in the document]
nherment: I really think expecting self.config to be of different types is the wrong approach. I reiterate the comment I made before that init_config should be idempotent and self.config should only have one data type and should not be both a dict and a `BashExecutorConfig`. This applies to all toolsets.

holmes/plugins/toolsets/datadog.py
```diff
@@ -133,18 +136,21 @@ def __init__(self):
...
-        if not config:
+    def prerequisites_callable(self) -> Tuple[bool, str]:
+        if not self.config:
             return False, ""
 
         try:
-            dd_config = DatadogConfig(**config)
-            self.dd_api_key = dd_config.dd_api_key
-            self.dd_app_key = dd_config.dd_app_key
+            self.init_config()
```
nherment: Is it not better to call `toolset.init_config()` within `toolset.check_prerequisites()`?

What happens where there are multiple prerequisites (possible by design)? There is an implicit guarantee thyat prerequisites are run in order but it's just not ideal to have to call `init_config` within a callable prerequisite.

nherment: Once `init_config` is called inside `check_prerequisites` it is important to catch any config validation error and return an appropriate error message that helps users remediating the error.

holmes/plugins/toolsets/kafka.py
```diff
@@ -499,12 +499,12 @@ def __init__(self):
...
 
-    def prerequisites_callable(self, config: Dict[str, Any]) -> Tuple[bool, str]:
-        if not config:
+    def prerequisites_callable(self) -> Tuple[bool, str]:
+        if not self.config:
             return False, TOOLSET_CONFIG_MISSING_ERROR
         errors = []
         try:
-            kafka_config = KafkaConfig(**config)
+            kafka_config = KafkaConfig(**self.config)
```
nherment: This should be an indication that the current design is incorrect.

Why does prerequisites callable need to initialize and validate the config when it's the role of `init_config`? It's repeated code and requires consistent error handling (user should see the same error for the same misconfiguration).

Rather here the pattern should be the one established. init_config validates and creates a typed config which is then used in `prerequisites_callable`.

holmes/plugins/toolsets/kafka.py
```diff
@@ -538,6 +538,26 @@ def prerequisites_callable(self, config: Dict[str, Any]) -> Tuple[bool, str]:
...
+            if cluster.kafka_security_protocol:
+                admin_config["security.protocol"] = cluster.kafka_security_protocol
+            if cluster.kafka_sasl_mechanism:
+                admin_config["sasl.mechanisms"] = cluster.kafka_sasl_mechanism
+            if cluster.kafka_username and cluster.kafka_password:
+                admin_config["sasl.username"] = cluster.kafka_username
+                admin_config["sasl.password"] = cluster.kafka_password
+
+            client = AdminClient(admin_config)
+            self.clients[cluster.name] = client  # Store in dictionary
```
nherment: The code here is a copy paste of the code in `prerequisites_callable`. It should be extracted in its a private method to avoid duplication.

holmes/plugins/toolsets/opensearch/opensearch_utils.py
```diff
@@ -28,12 +29,12 @@ def get_example_config(self) -> Dict[str, Any]:
...
 
-    def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
+    def prerequisites_callable(self) -> Tuple[bool, str]:
         env_url = os.environ.get("OPENSEARCH_LOGS_URL", None)
         env_index_pattern = os.environ.get("OPENSEARCH_LOGS_INDEX_NAME", "*")
-        if not config and not env_url:
+        if not self.config and not env_url:
             return False, "Missing opensearch traces URL. Check your config"
-        elif not config and env_url:
+        elif not self.config and env_url:
```
nherment: It's the role of init_config to consolidate a valid config so there is no longer any need to check for env vars here so the code here is duplicated unnecessarily .

nherment (changes requested):
making ti clear that I'm requesting changes. Happy to jump on a call and discuss these

holmes/plugins/toolsets/grafana/base_grafana_toolset.py
```diff
@@ -33,11 +33,11 @@ def __init__(
 
     def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
         if not config:
-            logging.debug(f"Grafana config not provided {self.name}")
+            logging.exception(f"Grafana config not provided {self.name}")
```
nherment: The docs for `logging.exception()` mention it should only be called from an exception handler. Isn't `logging.error` or `logging.warning` more appropriate here?

holmes/plugins/toolsets/kafka.py
```diff
@@ -499,45 +499,37 @@ def __init__(self):
...
+            if cluster.kafka_security_protocol:
+                admin_config["security.protocol"] = cluster.kafka_security_protocol
+            if cluster.kafka_sasl_mechanism:
+                admin_config["sasl.mechanisms"] = cluster.kafka_sasl_mechanism
+            if cluster.kafka_username and cluster.kafka_password:
+                admin_config["sasl.username"] = cluster.kafka_username
+                admin_config["sasl.password"] = cluster.kafka_password
+
+            client = AdminClient(admin_config)
+            self.clients[cluster.name] = client
```
nherment: It's great that the code is no longer duplicated. However I think there is a regression: the reporting of configuration errors is no longer complete as the code now would only report  a failure for the first cluster that is misconfigued as opposed to all of them.

One of our big focus is to smooth the onboarding and ease of configuration for our users and I feel this is a backward step.

holmes/plugins/toolsets/opensearch/opensearch.py
```diff
@@ -243,3 +230,12 @@ def get_example_config(self) -> Dict[str, Any]:
...
         return example_config.model_dump()
+
+    def init_config(self, config: dict[str, Any]):
+        os_config = OpenSearchConfig(**config)
+        for cluster in os_config.opensearch_clusters:
+            logging.info("Setting up OpenSearch client")
+            cluster_kwargs = cluster.model_dump()
+            client = OpenSearchClient(**cluster_kwargs)
+            if client.client.cluster.health(params={"timeout": 5}):
+                self.clients.append(client)
```
nherment: Putting the health check inside the config is fine by me be isn't what you want to avoid by separating the config?

What is important though is to make sure that if more than one cluster cannot be initialized that all the failures are reported to the user for the same reason that I mentioned about the Kafka clusters initialisation.

Looks like there is a bug here as well and I've recorded it separately: there is no error if a cluster reports unhealthy but it won't be part of the initialized clients.

holmes/plugins/toolsets/opensearch/opensearch_traces.py
```diff
@@ -138,11 +139,11 @@ def _invoke(self, params: Any) -> StructuredToolResult:
             logging.debug(f"opensearch traces search query: {full_query}")
             headers = {"Content-Type": "application/json"}
             headers.update(
-                add_auth_header(self._toolset.opensearch_config.opensearch_auth_header)
+                add_auth_header(self._toolset.typed_config.opensearch_auth_header)  # type: ignore
             )
 
             logs_response = requests.get(
-                url=get_search_url(self._toolset.opensearch_config),
+                url=get_search_url(self._toolset.typed_config),  # type: ignore
```
nherment: Why are `type: ignore` added here in multiple places? Typing is here as a safety mechanism against bugs that we should not ignore.

Would adding a not None check at the beginning of the invoke method help?

```
            if not self._toolset.typed_config:
                return StructuredToolResult(
                    status=ToolResultStatus.ERROR,
                    error=TOOLSET_CONFIG_MISSING_ERROR,
                    params=params
                )
```

holmes/plugins/toolsets/opensearch/opensearch_logs.py
```diff
@@ -140,8 +140,8 @@ def _get_fields_using_mappings(
     ) -> StructuredToolResult:
         """Use the OpenSearch getMappings API to retrieve fields (new implementation)"""
         url = urljoin(
-            self._toolset.opensearch_config.opensearch_url,
-            f"/{self._toolset.opensearch_config.index_pattern}/_mapping",
+            self._toolset.typed_config.opensearch_url,  # type: ignore
+            f"/{self._toolset.typed_config.index_pattern}/_mapping",  # type: ignore
```
nherment: Why are `type: ignore` added here in multiple places? Typing is here as a safety mechanism against bugs that we should not ignore.

Would adding a not None check at the beginning of the invoke method help?

```
            if not self._toolset.typed_config:
                return StructuredToolResult(
                    status=ToolResultStatus.ERROR,
                    error=TOOLSET_CONFIG_MISSING_ERROR,
                    params=params
                )
```

holmes/plugins/toolsets/opensearch/opensearch_utils.py
```diff
@@ -30,25 +33,24 @@ def get_example_config(self) -> Dict[str, Any]:
 
     def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
         env_url = os.environ.get("OPENSEARCH_LOGS_URL", None)
-        env_index_pattern = os.environ.get("OPENSEARCH_LOGS_INDEX_NAME", "*")
         if not config and not env_url:
             return False, "Missing opensearch traces URL. Check your config"
-        elif not config and env_url:
-            self.config = OpenSearchIndexConfig(
+        self.init_config(config)
+        return opensearch_health_check(self.typed_config)  # type: ignore
```
nherment: Same comment about `type: ignore`. Let's avoid adding these and do the necessary type checks at runtime.

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -816,3 +825,16 @@ def get_example_config(self):
...
+            self.typed_config = PrometheusConfig(**config)
+        else:
+            prometheus_url = os.environ.get("PROMETHEUS_URL")
+            if not prometheus_url:
+                prometheus_url = self.auto_detect_prometheus_url()
+            self.typed_config = PrometheusConfig(
+                prometheus_url=prometheus_url,
+                headers=add_prometheus_auth(os.environ.get("PROMETHEUS_AUTH_HEADER")),
+            )
+            logging.info(f"Prometheus auto discovered at url {prometheus_url}")
```
nherment: There is a regression here about error handling when prometheus auto discovery fails.
I understand that `self._is_healthy()` will report a config error but it's a generic message. The failure message that was there before this change `"Unable to auto-detect prometheus. Define prometheus_url in the configuration for tool prometheus/metrics"` is very helpful to users that rely on auto discovery and we should keep it.

tests/utils/toolsets.py
```diff
@@ -1,17 +1,17 @@
...
     return False, "Callable check failed"
 
 
-def callable_failure_no_message(config: Dict[str, Any]) -> Tuple[bool, str]:
+def callable_failure_no_message(onfig: dict[str, Any]) -> Tuple[bool, str]:
     return False, ""
 
 
-def failing_callable_for_test(config: dict):
+def failing_callable_for_test(onfig: dict[str, Any]):
```
nherment: typos `onfig` => `config`

tests/test_mcp_toolset.py
```diff
@@ -49,10 +48,10 @@ def test_mcpserver_unreachable():
     assert (
         False,
         "Failed to load mcp server test_mcp http://0.0.0.0:3009/sse ('unhandled errors in a TaskGroup', [ConnectError('All connection attempts failed')])",
-    ) == mcp_toolset.init_server_tools(config=None)
+    ) == mcp_toolset.init_server_tools(None)
 
 
-def test_mcpserver_1tool(monkeypatch):
+def test_mcpserver_tool(monkeypatch):
```
nherment: :+1:

holmes/core/toolset_manager.py
```diff
@@ -238,6 +241,9 @@ def load_toolset_with_status(
                     cached_status.get("type", ToolsetType.BUILTIN)
                 )
                 toolset.path = cached_status.get("path", None)
+            # Initialize the config for enabled toolsets
+            if toolset.enabled and toolset.status == ToolsetStatusEnum.ENABLED:
+                toolset.init_config(toolset.config)  # type: ignore
```
nherment: This looks like a valid type check failure that should not be ignored

```
Argument of type "dict[str, Any] | None" cannot be assigned to parameter "config" of type "dict[str, Any]" in function "init_config"
  Type "dict[str, Any] | None" is not assignable to type "dict[str, Any]"
```

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -25,7 +24,7 @@
 
 
 class BaseBashExecutorToolset(Toolset):
-    config: Optional[BashExecutorConfig] = None
+    typed_config: Optional[BashExecutorConfig] = None
```
nherment: :+1:

holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
```diff
@@ -145,8 +153,19 @@ def __init__(self):
...
             return False, TOOLSET_CONFIG_MISSING_ERROR
-        self.config = CoralogixConfig(**config)
-        if self.config.api_key:
-            return health_check(domain=self.config.domain, api_key=self.config.api_key)
-        else:
-            return False, "Missing configuration field 'api_key'"
+        try:
+            self.init_config(config)
+
+            if self.typed_config.api_key:  # type: ignore
```
nherment: `# type: ignore` to remove

holmes/plugins/toolsets/rabbitmq/toolset_rabbitmq.py
```diff
@@ -146,31 +148,18 @@ def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
...
-            logging.info("Loaded RabbitMQ config from environment variables.")
 
         try:
-            self.config = RabbitMQConfig(**config)
+            self.init_config(config)
         except Exception as e:
             return (False, f"Failed to parse RabbitMQ configuration: {str(e)}")
 
-        return self._check_clusters_config(self.config)
+        return self._check_clusters_config(self.typed_config)  # type: ignore
```
nherment: `# type: ignore` to remove

nherment (changes requested):
Hi @mainred ,

There is a previous comment (albeit minor) that is unaddressed: https://github.com/robusta-dev/holmesgpt/pull/535/files#r2162958028

Some other comments, mostly around not ignoring type checks and around reporting config errors to users.

Overall the approach looks good. Well done!

holmes/plugins/toolsets/grafana/base_grafana_toolset.py
mainred: Let me revert this change. config is not set is possible and the check just fails.

holmes/plugins/toolsets/kafka.py
mainred: Yeah, I agree. But I did not find a better way to organize all the errors since errors are raised in init_config, and it seems to be no better way to let init_config pass the errors  to prerequisites_callable

holmes/plugins/toolsets/opensearch/opensearch_traces.py
mainred: Thanks for the suggestion.  I did not know the none type check can help avoid adding `type: ignore`

holmes/core/toolset_manager.py
mainred: Should we allow init_config initialize the type configurations without user config?

nherment: I think so. Sometimes the is no config and/or no config is valid as there are defaults like the internet toolset.

holmes/plugins/toolsets/kafka.py
nherment: It's not great but would it work to gather all errors and then raise an exception? It's not too different than how a pydantic validation error would behave. It's also somewhat simple to implement.

holmes/plugins/toolsets/prometheus/prometheus.py
mainred: added auto-detect back.

holmes/plugins/toolsets/kafka.py
mainred: Done. thanks.

holmes/plugins/toolsets/opensearch/opensearch.py
mainred: @arikalon1 I think we need your opinion around handling these cases.

holmes/plugins/toolsets/bash/bash_toolset.py
mainred: Instead of reusing config, which might be confused with Toolset.config, I use typed_config for specific toolsets for a typed config.
FYI @arikalon1  We had discussions around this variable naming, but now we are on the same page to use typed_config to be specific.

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -200,9 +199,16 @@ def _reload_llm_instructions(self):
...
-    def prerequisites_callable(self, config: dict[str, Any]) -> tuple[bool, str]:
+    def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
+        try:
+            self.init_config(config)
+        except Exception as e:
+            logging.exception("Failed to initialize BashExecutorToolset configuration")
+            return False, str(e)
+        return True, ""
+
+    def init_config(self, config: Optional[dict[str, Any]]):
```
mainred: Another concern is related to init_config.
prerequisites_callable calls init_config and can catch the exception when config is not well set. But in init_config, which can be also called when the status is cached and prerequisites_callable is not called https://github.com/robusta-dev/holmesgpt/pull/535/files#diff-bfd1b7e4ad5810c10197015d94b5d2aa9f3ad74a937f531f25954306858b08b0R246. The exception is not well handled, since we assume the prerequisites have been checked, but it is not also so considering some configurations are set by environment variables, and environment variables can be cleaned up after console restarts.

cc @arikalon1  @nherment

holmes/plugins/toolsets/opensearch/opensearch.py
arikalon1: @mainred I think we need to be aligned with the previous behavior
I think previously, it was checking health on all clusters.
If a cluster is healthy - it's added to the list.
If not - the error is returned.

Currently, I think there are 2 issues:
1. The loop in line 240, is not with try/except - so any error will not be caught and will stop the execution
2. I think there's another loop, in line 247 - looks like almost the same code. Is it duplicated? or, is this on purpose?

holmes/plugins/toolsets/opensearch/opensearch_traces.py
```diff
@@ -130,6 +131,12 @@ def __init__(self, toolset: "OpenSearchTracesToolset"):
     def _invoke(self, params: Any) -> StructuredToolResult:
         err_msg = ""
         try:
+            if not self._toolset.typed_config:
```
arikalon1: @mainred Why do we need this check?
How can we get to `_invoke`, if no config was provided?
Wouldn't the toolset be disabled?

holmes/plugins/toolsets/bash/bash_toolset.py
arikalon1: @mainred with `init_config`, if the configuration is invalid, it should always raise an exception? And then the tool is not initialized?

holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
```diff
@@ -51,33 +52,37 @@ def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
...
-        return health_check(domain=self.config.domain, api_key=self.config.api_key)
+        return health_check(
+            domain=self.typed_config.domain, api_key=self.typed_config.api_key
+        )
 
-    @property
-    def coralogix_config(self) -> Optional[CoralogixConfig]:
-        return self.config
+    def init_config(self, config: Optional[dict[str, Any]]):
+        if not config:
```
arikalon1: @mainred here, for example, `init_config` doesn't raise an exception when it's not initialized.
Will the Coralogix tool appear enabled or disabled?

holmes/plugins/toolsets/kafka.py
```diff
@@ -499,45 +499,47 @@ def __init__(self):
...
-                    self.clients[cluster.name] = client  # Store in dictionary
-                except Exception as e:
-                    message = (
-                        f"Failed to set up Kafka client for {cluster.name}: {str(e)}"
-                    )
-                    logging.error(message)
-                    errors.append(message)
-
-            return len(self.clients) > 0, "\n".join(errors)
+            self.init_config(config)
```
arikalon1: @mainred here as well
If no config, `init_config` doesn't raise an exception. But we return True. 
Is this correct?

holmes/plugins/toolsets/opensearch/opensearch_traces.py
mainred: typed_config is an optional object. If we don't check the emptiness, we need to add `type: ignore`.

holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
mainred: we expect prerequisite_call to check the enablement of the tool, in which init_config is called.

holmes/plugins/toolsets/bash/bash_toolset.py
arikalon1: @nherment @mainred Can you share the reason for choosing a different member, instead of overriding the `config` member on the base class (like in an OOP manner)?

PR #542, opened 2025-06-19
speed up cli startup

aantn:


holmes/config.py
```diff
@@ -169,12 +171,15 @@ def load_from_file(cls, config_file: Optional[Path], **kwargs) -> "Config":
         cli_options = {k: v for k, v in kwargs.items() if v is not None and v != []}
 
         if config_from_file is None:
-            return cls(**cli_options)
+            result = cls(**cli_options)
+        else:
+            logging.debug(f"Overriding config from cli options {cli_options}")
+            merged_config = config_from_file.dict()
+            merged_config.update(cli_options)
+            result = cls(**merged_config)
```
aantn, line 189: Note that without these changes, model_post_init() is called twice (once here and once above for load_model_from_file()) and all the logging happens twice

mainred: To clarify, if config_from_file is not None, `model_post_init()` is inevitably called twice, but this change logs only once by calling `log_useful_info` once.

holmes/clients/robusta_client.py
```diff
@@ -1,18 +1,20 @@
...
-TIMEOUT = 2
+TIMEOUT = 0.3
 
 
 class HolmesInfo(BaseModel):
     model_config = ConfigDict(extra="ignore")
     latest_version: Optional[str] = None
 
 
+@cache
```
mainred, line 16: Will this reduce the call to holmes across `holmes ask` commands?

holmes/config.py
aantn: Exactly.

holmes/clients/robusta_client.py
aantn: @mainred can you explain the question further?

mainred: Normally we call fetch_holmes_info once for each `holmes ask`, so if the cache does not work other, following, `holmes ask`, it brings no benefit.

aantn: @mainred it helps here because `model_post_init` is called twice in certain flows, so without this cache, `fetch_holmes_info` is also called twice which can be expensive.

holmes/clients/robusta_client.py
```diff
@@ -1,18 +1,20 @@
...
 def fetch_holmes_info() -> Optional[HolmesInfo]:
```
aantn, line 17: @moshemorad is it safe to cache this for the server too? (i.e. are there cases where the model list might change and holmes might reload its config but the holmes pod would not restart?)

mainred (dismissed):

holmes/clients/robusta_client.py
moshemorad: @aantn Yeah, it is currently only used to compare to latest version. if the pod already running version cannot change.
No harm in that case.

aantn: Thanks!

aantn:
@mainred do you mind re-approving? (nothing substantial changed, but I had to resolve a merge conflict so the review is stale)

arikalon1 (dismissed):

PR #543, opened 2025-06-19
More tool logging improvements

aantn:
Feel free to suggest an alternative format. I don't think we've gotten it quite right yet!

### DIFF
--- a/holmes/core/tools.py
+++ b/holmes/core/tools.py
@@ -134,8 +134,14 @@ class Tool(ABC, BaseModel):
             if hasattr(result, "get_stringified_data")
             else str(result)
         )
+        if len(output_str) == 0:
+            preview = "<empty>"
+        elif len(output_str) > 80:
+            preview = f"{(output_str[:80] + "...")!r}"
+        else:
+            preview = f"{output_str!r}"
         logging.info(
-            f"   Finished in {elapsed:.2f}s, output length: {len(output_str):,} characters, preview ⬇\n   {output_str[:80]!r}..."
+            f"|--- Finished in {elapsed:.2f}s, output length: {len(output_str):,} characters, preview ⬇\n     {preview}"
         )
         # return format_tool_output(result)
         return result


mainred (approved):

holmes/core/tools.py
```diff
@@ -134,8 +134,15 @@ def invoke(self, params: Dict) -> StructuredToolResult:
...
+        if len(output_str) == 0:
+            preview = "<empty>"
+        elif len(output_str) > 80:
+            clipped = output_str[:80] + "..."
+            preview = f"{clipped!r}"
+        else:
+            preview = f"{output_str!r}"
         logging.info(
-            f"   Finished in {elapsed:.2f}s, output length: {len(output_str):,} characters, preview ⬇\n   {output_str[:80]!r}..."
+            f"|--- Finished in {elapsed:.2f}s, output length: {len(output_str):,} characters, preview ⬇\n     {preview}"
```
mainred, line 146: The output looks like this
```
loading toolset status from cache                                                                 toolset_manager.py:216
User: detect why the k8s pod client under namespace test-ns cannot resolve dns
Running tool kubectl_find_resource: kubectl get -A --show-labels -o wide pod | grep client                  tools.py:126
|--- Finished in 1.37s, output length: 236 characters, preview ⬇                                            tools.py:144
     'test-ns             client                                                   1/1...'
Running tool trace_dns_gadget: fetch DNS queries and responses of pod test-ns/client for 30 seconds         tools.py:126
|--- Finished in 42.11s, output length: 20,165 characters, preview ⬇                                                                         tools.py:144
     'time="2025-06-24T00:54:47Z" level=warning msg="aks-nodepool1-16826556-vmss000002...'
```

mainred (approved):
Looks good.  Thanks for consistently refining the output of logging

holmes/core/tools.py
moshemorad: @aantn in theory since we are running in thread pool with more then one workter in the stream_call it might be printed out of order right? Not sure if it is deal breaker or not but wanted to make sure we are atleast taking it into considaration.

aantn: Yes, good catch! This problem is already present on master and this PR doesn't make it worse in any way, so I'm going to merge this and open a separate github issue to track

PR #544, opened 2025-06-19
Fix version warning from pipx

### LINKED ISSUE #466: holmes reports wrong version, when installed with pipx from github
When installing holmes from pipx ([according to docs here](https://github.com/robusta-dev/holmesgpt/blob/master/docs/installation.md#pip-and-pipx)) there are two version-related problems:

1. holmes reports its version during installation as 0.1.0, which is incorrect. (The incorrect number is likely taken from pyproject.toml)

Image

2. holmes thinks there is a newer version:

Image

aantn:
Fixes #466 
Note: this is only a partial fix, as now holmes will disable version checks altogether when installed from a github source version.

I think this is OK, as once we push to pypi users will be instructed to install from pypi and the version check will work for them.

### DIFF
--- a/holmes/config.py
+++ b/holmes/config.py
@@ -115,11 +115,18 @@ class Config(RobustaBaseConfig):
 
     @property
     def is_latest_version(self) -> bool:
-        if self._holmes_info and self._holmes_info.latest_version and self._version:
-            return self._version.startswith(self._holmes_info.latest_version)
+        if (
+            not self._holmes_info
+            or not self._holmes_info.latest_version
+            or not self._version
+        ):
+            # We couldn't resolve version, assume we are running the latest version
+            return True
+        if self._version.startswith("dev-"):
+            # dev versions are considered to be the latest version
+            return True
 
-        # We couldn't resolve version, assume we are running the latest version
-        return True
+        return self._version.startswith(self._holmes_info.latest_version)
 
     @property
     def toolset_manager(self) -> ToolsetManager:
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "holmesgpt"
-version = "0.1.0"
+version = "0.0.0"
```
mainred, line 3: Normally, we should bump the version to the latest tag in master and each release should have the corresponding  version, but as we currently recommend master branch, and version should follow [pep-0440](https://peps.python.org/pep-0440/), we can temporarily use 0.0.0 for master branch.

holmes/config.py
```diff
@@ -144,9 +151,7 @@ def model_post_init(self, __context: Any) -> None:
 
         if not self.is_latest_version and self._holmes_info:
             logging.warning(
-                "You are running version %s of holmes, but the latest version is %s. Please update to the latest version.",
-                self._version,
-                self._holmes_info.latest_version,
+                f"You are running version {self._version} of holmes, but the latest version is {self._holmes_info.latest_version}. Please update.",
```
mainred, line 154: We may return warning when the user tries the officially released version, and we suggest them to do the upgrad, and we can consider the other cases as test?

pyproject.toml
aantn: Yeah, agreed. If we bumped the number during releases in the file itself it would solve a few issues here. Do you have an example of how another project does it? (Not going to hold up this pr for that, but for future improvement.)

holmes/config.py
aantn: Can you explain?

mainred: I mean normally the user should pick the official releases or the other cases as test or dev. We maintain the official release and suggest them to do upgrade in case it's not latest and give this warning, but for the other cases, we may skip the warning for these dev or test cases?

aantn: Yes, exactly. That should be the behaviour right now with these changes. (When 0.0.0 is present - or when we have the information from git, we assume a dev version.)

We just need to have a way to install officially releases, which #418 will do. (Right now you can install from brew also.)

mainred (approved):
Thanks for your change.

PR #547, opened 2025-06-20
feat(runbook): introduce runbook toolset to fetch internal runbooks

mainred:
This PR introduces a toolset Runbook to fetch the built-in runbooks, the catalog are integrated into the system prompt. I closed the PR to let llm pick the runbook and append it to the initial steps before executing any tools  https://github.com/robusta-dev/holmesgpt/pull/533, but it's not already necessary to follow a runbook and the runbook selected might be misleading depending only on the initial user prompt.
The result of toolset will not be appended to the user prompt, but as the other tool result, it will be part of the assistant prompt

fix: #473


holmes/plugins/prompts/_general_instructions.jinja2
```diff
@@ -33,6 +33,8 @@
 
 {% include '_fetch_logs.jinja2' %}
 
+{% include '_runbook_instructions.jinja2' %}
+
```
mainred: @aantn @arikalon1 I can your help to check if adding the catalog here will break the other scenarios. I tried to change generic_ask.jinja2 which is the system prompt of `holmes ask`.

holmes/plugins/prompts/_runbook_instructions.jinja2
```diff
@@ -0,0 +1,15 @@
...
+
+{%- for runbook in runbooks.catalog -%}
+
+desciption: {{ runbook.description }}
+
+link: {{ runbook.link }}
+
+{%- endfor -%}
+
+Always try to find the matched runbooks from the available runbooks to troubleshoot the issue by comparing the runbook description with the user prompt.
```
mainred: @aantn  @arikalon1 I need your help review the prompt of the runbooks.

mainred:
@nilo19 sorry I closed the earlier PR, and after I force pushed the branch and try to reopen that PR, it's not allowed anymore. Please review this PR.

holmes/plugins/prompts/_general_instructions.jinja2
arikalon1: @mainred why did you choose to change `_general_instructions.jinja2` and not `generic_ask.jinja2` ?

`_general_instructions.jinja2` is used in all Holmes investigations flows, and I think it would be good to have runbooks in all of them.
But I'm asking, because you wrote you tried to change only the `ask` prompt.

I think it won't break the other scenarios, but after running the evals we'll be able to verify it works well.

holmes/main.py
```diff
@@ -1,10 +1,10 @@
 # ruff: noqa: E402
 import os
-from holmes.core.prompt import append_file_to_user_prompt
 from collections import OrderedDict
```
arikalon1: Can you please move all the imports after line 16?

This code, loads custom certificates (it's used by some users that runs Holmes on environments with Proxies with self signed certificates)

Doing imports before loading the certificates, might indirectly start some http client, without the custom certificate, which will cause it to fail on those environments.
So we add imports, only after loading the custom certificates, if exists

holmes/plugins/prompts/_runbook_instructions.jinja2
arikalon1: The best way to check if the prompt is good, is to add evals for it.
I would add both positive and negative evals. (I would verify a runbook is selected when required, and not selected when is not needed)

I think it's important for the LLM to say, if's it's not doing a runbook step, because of missing tool, or anther reason.

I haven't tested it, but this is the prompt I would use (it's a little shorter, because I think the LLM is smart)

```
ALWAYS try to find runbooks that can provide troubleshooting instructions.
To get the runbook details, use the `fetch_runbook` tool.
ALWAYS follow the steps described in the runbook.
If you decided not to follow one or more steps, ALWAYS explain why
```

holmes/plugins/toolsets/runbook/runbook_fetcher.py
```diff
@@ -0,0 +1,77 @@
...
+
+# TODO(mainred): currently we support fetch runbooks hosted internally, in the future we may want to support fetching
+# runbooks from external sources as well.
+class RunbookFetcher(Tool):
+    toolset: "RunbookToolset"
+
+    def __init__(self, toolset: "RunbookToolset"):
+        super().__init__(
+            name="fetch_runbook",
+            description="Reads a runbook from the runbook catalog",
```
arikalon1: I think a better description might be:
`Get runbook content by runbook path. Use this to get troubleshooting steps for incidents`

arikalon1 (approved):
Thanks for the PR @mainred 

Very nice work!

Left some small comments

holmes/main.py
mainred: Sure. I saw the NOTE in L14-15, and did try to add imports after L16, but this change moves imports from L3 to L7 by isort automatically. I think it's ok to keep this change, because it does not impact the certificate loading logic, and it fixes the import sequence(group)?

holmes/plugins/prompts/_general_instructions.jinja2
mainred: Sorry, I should give my evidence.
Because `generic_ask.jinja2` is used by generic_ticket.jinjia2 in [ `holmes investigate ticket`](https://github.com/robusta-dev/holmesgpt/blob/master/holmes/main.py#L767), not for ask only, and describing the runbook tool in _general_instructions.jinja2 around [{% include '_toolsets_instructions.jinja2' %}](https://github.com/robusta-dev/holmesgpt/blob/master/holmes/plugins/prompts/_general_instructions.jinja2#L32). But these evidence are not strong. let me move it generic_ask.jinja2 to not impact the other scenarios.

holmes/plugins/prompts/_runbook_instructions.jinja2
mainred: I can add evals later. Thanks for your suggestions.

holmes/config.py
```diff
@@ -225,6 +229,16 @@ def __get_cluster_name() -> Optional[str]:
 
         return None
 
+    @staticmethod
+    def get_runbook_catalog() -> str:
+        # TODO(mainred): besides the built-in runbooks, we need to allow the user to bring their own runbooks
+        runbook_catalog = load_runbook_catalog()
+        if runbook_catalog is not None:
```
nilo19, line 241: change to if not runbook_catalog to be simpler?

holmes/main.py
arikalon1: Please move the imports after the certificate loading.
We had few times in the past, where this caused a major failure

This is why we added the comment on lines 14-15

holmes/plugins/runbooks/__init__.py
```diff
@@ -48,3 +53,48 @@ def load_builtin_runbooks() -> List[Runbook]:
...
+        logging.error(
+            f"Unexpected error while loading runbook catalog from {catalogPath}: {e}"
+        )
+    return None
+
+
+def get_runbook_by_path(runbook_relative_path: str) -> str:
+    runbook_folder = os.path.dirname(os.path.realpath(__file__))
+    runbook_path = os.path.join(runbook_folder, runbook_relative_path)
+    return runbook_path
```
nilo19, line 100: better check if the path is in the runbook directory. hackers may attack by creating a malicious path e.g., ../../xx

holmes/config.py
mainred: `runbook_catalog is not None` and 'not runbook_catalog` achieves the same result in this case, but the later checks not if it's more than None and for the other falsy values. load_runbook_catalog returns a optional string, which means it can be either None or a valid string. So I think we can keep using `is not None` to be specific.
Here is some clarification of the difference of these two syntax. 
```
Specificity:
variable is not None is specific to checking for the absence of a value (i.e., None). not variable is a more general check for "emptiness" or "falsiness."
Intent Clarity:
variable is not None explicitly communicates the intent to check if a variable holds a value other than None. not variable can be ambiguous if the intent is solely to check for None, as it will also evaluate to True for other falsy values.
```

holmes/plugins/runbooks/__init__.py
mainred: I think it's ok because we user `__file__` to locate the path the runbooks, which is under the same folder as the __init__.py.

nilo19: they may use some prompt to let llm call runbook_folder/../../xx?

mainred: This path is like a constant, and is determined by where this __init__.py is located, the use prompt cannot decide the folder. we use runbook fetcher tool to get the runbook which accepts only the link, relative path, to the file, so the path  is not subject to any user input.

holmes/main.py
mainred: Reverted  the change, since we have wrapped what is required to load the certificate in a function.

arikalon1 (approved):

holmes/config.py
arikalon1: Resolving the comment to be able to merge

PR #551, opened 2025-06-21
If using Gemini - disable structured output

arikalon1:
Fix a bug that swallows exception on bad completion request


holmes/core/investigation.py
```diff
@@ -118,6 +118,9 @@ def get_investigation_context(
     elif ai.llm.model and ai.llm.model.startswith("bedrock"):
         # Structured output does not work well with Bedrock Anthropic Sonnet 3.5 through litellm
         request_structured_output_from_llm = False
+    elif ai.llm.model and ai.llm.model.startswith("gemini"):
```
RoiGlinik: we can merge both ifs, startwith can work with tuple

RoiGlinik (dismissed):

holmes/core/investigation.py
arikalon1: fixed

PR #556, opened 2025-06-23
MCP toolset validation fixes

vrutkovs:
* Don't reset tool status to CUSTOM
* Don't run yaml toolset validation for MCP tools


nherment (approved):
Looks good!

PR #558, opened 2025-06-23
prometheus: fix reading prometheus options

vrutkovs:
`self.toolset.config` is a dict, not an object, so we need to access its items using the `get` method instead of attribute access.

Not sure if this is intended though, perhaps `config` is meant to be an object?

### DIFF
--- a/holmes/plugins/toolsets/prometheus/prometheus.py
+++ b/holmes/plugins/toolsets/prometheus/prometheus.py
@@ -289,14 +289,16 @@ class ListPrometheusRules(BasePrometheusTool):
         self._cache = None
 
     def _invoke(self, params: Any) -> StructuredToolResult:
-        if not self.toolset.config or not self.toolset.config.prometheus_url:
+        if not self.toolset.config or not self.toolset.config.get("prometheus_url"):
             return StructuredToolResult(
                 status=ToolResultStatus.ERROR,
                 error="Prometheus is not configured. Prometheus URL is missing",
                 params=params,
             )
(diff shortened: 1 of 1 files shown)


arikalon1:
hey @vrutkovs 

Thanks for the PR
I think the config object in the prometheus toolset, should be a `PrometheusConfig` object, and not a dict
The fact that it's a dict, is due to another bug, which should be fixed, but was not merged yet

That bug fix should be merged within a few days - I'll update you when it does, and, if that's ok, will ask you to check if this fix is still required

vrutkovs:
Oh, okay, thanks! I'll keep the PR open for now (in case I find some other fixes along the way too)

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -289,14 +289,16 @@ def __init__(self, toolset: "PrometheusToolset"):
         self._cache = None
 
     def _invoke(self, params: Any) -> StructuredToolResult:
-        if not self.toolset.config or not self.toolset.config.prometheus_url:
+        if not self.toolset.config or not self.toolset.config.get("prometheus_url"):
```
mainred, line 292: Hey @vrutkovs I think we want to fix the config issue by https://github.com/robusta-dev/holmesgpt/pull/535. The config object should be initialized during check-prerequisite, but was skipped for status check.

vrutkovs: Thanks, that fixed it for me!

PR #565, opened 2025-06-25
fix: fix several check error introduced by old code

mainred:
These check failures are annoying. They only appear only when there's new check errors in the code, which brings extra effort to distinguish which checks are really brought the new code.

```
holmes/plugins/toolsets/service_discovery.py:27: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
holmes/plugins/toolsets/service_discovery.py:32: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
holmes/plugins/toolsets/prometheus/prometheus.py:548: error: If x = b'abc' then f"{x}" or "{}".format(x) produces "b'abc'", not "abc". If this is desired behavior, use f"{x!r}" or "{!r}".format(x). Otherwise, decode the bytes  [str-bytes-safe]
holmes/plugins/toolsets/prometheus/prometheus.py:694: error: If x = b'abc' then f"{x}" or "{}".format(x) produces "b'abc'", not "abc". If this is desired behavior, use f"{x!r}" or "{!r}".format(x). Otherwise, decode the bytes  [str-bytes-safe]
holmes/plugins/toolsets/grafana/loki_api.py:51: error: Argument "params" to "get" has incompatible type "dict[str, object]"; expected "SupportsItems[str | bytes | int | float, str | bytes | int | float | Iterable[str | bytes | int | float] | None] | tuple[str | bytes | int | float, str | bytes | int | float | Iterable[str | bytes | int | float] | None] | Iterable[tuple[str | bytes | int | float, str | bytes | int | float | Iterable[str | bytes | int | float] | None]] | str | bytes | None"  [arg-type]
holmes/plugins/sources/pagerduty/__init__.py:130: error: Item "None" of "Response | None" has no attribute "text"  [union-attr]
holmes/plugins/sources/opsgenie/__init__.py:32: error: Argument "params" to "get" has incompatible type "dict[str, object]"; expected "SupportsItems[str | bytes | int | float, str | bytes | int | float | Iterable[str | bytes | int | float] | None] | tuple[str | bytes | int | float, str | bytes | int | float | Iterable[str | bytes | int | float] | None] | Iterable[tuple[str | bytes | int | float, str | bytes | int | float | Iterable[str | bytes | int | float] | None]] | str | bytes | None"  [arg-type]
holmes/plugins/sources/prometheus/plugin.py:73: error: Argument 1 to "HTTPBasicAuth" has incompatible type "str | None"; expected "bytes | str"  [arg-type]
holmes/plugins/sources/prometheus/plugin.py:73: error: Argument 2 to "HTTPBasicAuth" has incompatible type "str | None"; expected "bytes | str"  [arg-type]
```


holmes/plugins/sources/prometheus/plugin.py
```diff
@@ -70,7 +70,7 @@ def __fetch_issues_from_api(self) -> List[PrometheusAlert]:
             logging.info(f"Filtering alerts by {self.label_filter}")
 
         if self.username is not None or self.password is not None:
```
mainred, line 72: Can we check both var are not None? so that we don't have to add `type ignore` but not sure if an empty password and a valid name work.

moshemorad (approved):

PR #572, opened 2025-06-26
improve cli output

aantn:


holmes/interactive.py
```diff
@@ -114,8 +117,16 @@ def run_interactive_loop(
                         f"{tool_call.description}. Output=\n{tool_call.result}",
                         markup=False,
                     )
-            console.print("[bold green]AI:[/bold green]", end=" ")
-            console.print(Markdown(response.result))  # type: ignore
+            # console.print(f"\n⏺ [bold {AI_COLOR}]AI:[/bold {AI_COLOR}]", end=" ")
```
mainred: Will these two commented code be  used in the future? otherwise we can consider removing them.

holmes/main.py
```diff
@@ -109,21 +109,45 @@ def init_logging(verbose_flags: Optional[List[bool]] = None):
         logging.basicConfig(
             level=logging.DEBUG,
             format="%(message)s",
-            handlers=[RichHandler(show_level=False, show_time=False)],
+            handlers=[
+                RichHandler(
+                    show_level=False,
+                    markup=True,
+                    show_time=False,
+                    show_path=False,
```
mainred, line 117: We removed the code path in this PR, and this is good.

aantn: Thanks!

holmes/interactive.py
aantn: Nice catch, removed.

PR #574, opened 2025-06-27
Rob 1554 azure sql toolset

nherment:


holmes/plugins/toolsets/azure_sql/apis/storage_analysis_api.py
```diff
@@ -0,0 +1,321 @@
...
+            state_desc as file_state
+        FROM sys.database_files
+        ORDER BY type_desc, file_id;
+        """
+
+        try:
+            return self.sql_api_client._execute_query(server_name, database_name, query)
+        except Exception as e:
+            logging.error(f"Failed to get database size details: {str(e)}")
+            return [{"error": str(e)}]
```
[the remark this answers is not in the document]
nherment: done

holmes/plugins/toolsets/azure_sql/azure_sql_toolset.py
```diff
@@ -0,0 +1,182 @@
...
+
+            except Exception as e:
+                message = f"Failed to set up Azure authentication: {str(e)}"
+                logging.error(message)
+                errors.append(message)
+                return False, message
+
+            # Store single database configuration and create API client
+            self._database_config = azure_sql_config.database
+            self._api_client = AzureSQLAPIClient(
```
moshemorad, line 131: I saw that this could also get username & password but we never pass it and it isn't support in the config. 
It is still needed? can we remove it?

holmes/plugins/toolsets/azure_sql/azure_sql_toolset.py
```diff
@@ -0,0 +1,182 @@
...
+                credential, azure_sql_config.database.subscription_id
+            )
+            logging.info(
+                f"Configured Azure SQL database: {azure_sql_config.database.server_name}/{azure_sql_config.database.database_name}"
+            )
+
+            # Validate each tool's configuration requirements
```
moshemorad, line 138: Just to make sure this is on purpose, we don't validate the tools then? we will add it in the future?

holmes/plugins/toolsets/azure_sql/azure_sql_toolset.py
```diff
@@ -0,0 +1,182 @@
...
+    def __init__(self):
+        # Reduce Azure SDK HTTP logging verbosity
+        logging.getLogger("azure.core.pipeline.policies.http_logging_policy").setLevel(
+            logging.WARNING
+        )
+        logging.getLogger("azure.identity").setLevel(logging.WARNING)
+        logging.getLogger("azure.mgmt").setLevel(logging.WARNING)
+        logging.getLogger("azure.monitor").setLevel(logging.WARNING)
+
+        super().__init__(
```
moshemorad, line 56: Please mark as experimental

holmes/plugins/toolsets/azure_sql/azure_sql_toolset.py
nherment: Removed

holmes/plugins/toolsets/azure_sql/azure_sql_toolset.py
nherment: This is actually a bigger problem.
I think we should validate the tools because each tool may require different permissions.
It makes sense that these tools are part of the same toolset. However toolsets are reported as fully working or not working at all in the SaaS UI. We need a third state which is `error reported but toolset enabled`.

This would allow reporting incorrect permissions, incorrect config for some tools without impacting other tools in the same toolset.

So for now we don't want to validate that all tools are working because if any permission/setup is incorrect it would prevent the whole toolset from working.

moshemorad: @nherment I think that this is something that we need to discuss also with @aantn and product to understand what we expect in those cases. Very good point. (maybe we could just separate it to different toolset but yeah we need to discuss it further.)

aantn: Yes, its a great point. I don't have a strong preference for how to handle. Can you suggest something that makes sense from a technical perspective and create a ticket to track? I'll review and unless there's some blocker with the proposed solution from a product perspective, I can ask keren and frontend to plan the changes on their side.

PR #577, opened 2025-06-28
make pytest output more concise + skip grafana test in more conditions

aantn:
Two improvements to test-skipping in pytest:

1. Instead of spamming logs with a message for every skipped test, log a single message for each skipped test module. e.g. log `SKIPPED [6]` instead of 6 individual SKIPPED messages
2. Skip grafana tests when GRAFANA_URL is defined, but we're not able to actually connect to grafana


aantn:
Please note: I don't have the setup to run any of the skipped tests right now, so whoever reviews this - please verify the tests still run when NOT skipped and that I didn't break anything, especially with grafana

PR #578, opened 2025-06-28
further pytest improvements to improve DX

aantn:


tests/llm/conftest.py
```diff
@@ -23,52 +23,72 @@ def check_llm_api_with_test_call():
...
 
 
-def pytest_collection_modifyitems(config, items):
-    """Show warning when LLM evaluation tests are collected"""
+@pytest.fixture(scope="session", autouse=True)
+def llm_session_setup(request):
+    """Handle LLM test session setup: show warning, check API, and skip if needed"""
+    # Don't show messages during collection-only mode
+    if request.config.getoption("--collect-only"):
+        yield
```
moshemorad: Why we need the yield in that case? yield is used to be able and to have teardown after the tests are finished. e.g. it get back to here and continue the code execution but i see we use always return after yield and there is now "tear down" code. Am i missing something?

moshemorad (dismissed):

tests/llm/conftest.py
aantn: I think you're right and it's equivalent.

aantn: I pushed a fix for this.

aantn:
@moshemorad I pushed a fix + another small improvement.
Can you please re-review?

PR #579, opened 2025-06-29
Run eval before_test/after_test only once even with multiple ITERATIONS

aantn:
Builds upon https://github.com/robusta-dev/holmesgpt/pull/578


aantn:
Replaced by https://github.com/robusta-dev/holmesgpt/pull/702

PR #583, opened 2025-07-01
chore: check toolset prerequisites in parallel

mainred:
This is a follow-up PR for https://github.com/robusta-dev/holmesgpt/pull/535

In this PR, we run toolset prerequisite check in parallel, also we mute the toolset status log when the toolset is loaded from cache. This help ask command output less verbose log, and bring a cleaner UI to the user.


holmes/core/toolset_manager.py
```diff
@@ -113,14 +114,31 @@ def _list_all_toolsets(
...
                 toolset.status = ToolsetStatusEnum.DISABLED
+        self.check_toolset_prerequisites(enabled_toolsets)
 
         return list(toolsets_by_name.values())
 
+    @classmethod
+    def check_toolset_prerequisites(
+        cls, toolsets: list[Toolset], log_toolset_status: bool = True
+    ):
+        with concurrent.futures.ThreadPoolExecutor() as executor:
```
mainred: I don't have in mind what the proper max_workers should be, so leave it empty and use default one decided by cpu count of local machine.
The number of toolset varies between 0 and 34 decided by the local env setup. Considering the majority of the toolset prerequisite check are fast and resource saving, I think it's ok to use the default max_workers.

mainred:
kindly ping @nherment and @aantn to take a look.

holmes/core/toolset_manager.py
aantn: I would consider being even more aggressive, as the prerequisite checks are not CPU bound, so we'll benefit from more parallelization. Maybe go up to 10 by default?

holmes/core/toolset_manager.py
```diff
@@ -242,13 +261,23 @@ def load_toolset_with_status(
...
+            # check prerequisites for only enabled toolset when the toolset is loaded from cache. When the toolset is
+            # not loaded from cache, the prerequisites are checked in the refresh_toolset_status method.
             if (
                 toolset.enabled
                 and toolset.status == ToolsetStatusEnum.ENABLED
                 and using_cached
             ):
-                toolset.check_prerequisites()  # type: ignore
+                enabled_toolsets_from_cache.append(toolset)
+        # we don't log the status of toolsets when it's loaded from cache, considering
```
aantn: Why do we need to check prerequisites at all in this case? Isn't the purpose of the cache to avoid checking prerequisites altogether?

If we are checking them, I think we should continue to log. 

For me, the real solution is to avoid to check them altogether when possible, not silencing the log.

mainred: Yeah, we expect to use cache to avoid checking prerequisites, but an enabled toolset in the cache does not always indicate the toolset is available. Take the ask toolset for example, it requires logging into the azure account. But when the login expires, though the cache status is enabled, it cannot be used.
So I am trying to scope down what toolsets requires to check prerequisites instead of all toolset check every time we run the ask command.
For mute the logging toolset in the ask command, my intention is to make the aks interactive mode as clean as possible, but I agree with what you said, since not all toolset can be enabled, logging the failed status can help the user re-able the toolset.

aantn (approved):
LGTM, thanks

PR #584, opened 2025-07-01
fix: use correct prerequisites to check az cli login status

mainred:
`az account list` does not return error when az cli does not log in to the Azure. Switch it to `az account show` to exit with error when azure login is done.


mainred:
cc @nilo19

PR #585, opened 2025-07-01
ROB-1470 notion toolset precheck

RoiGlinik:


holmes/plugins/toolsets/datadog.py
```diff
@@ -135,16 +135,19 @@ def __init__(self):
...
+            return (
+                False,
+                "Datadog toolset is misconfigured. 'dd_api_key' and 'dd_app_key' are required.",
+            )
 
         try:
             dd_config = DatadogConfig(**config)
             self.dd_api_key = dd_config.dd_api_key
             self.dd_app_key = dd_config.dd_app_key
             return bool(self.dd_api_key and self.dd_app_key), ""
```
moshemorad: Suggestions: make both the fields here required in the config and it will fail. 
In the current implementation if one of them is missing we will not have error message again.
If you have other suggestion like checking it and adding error message im fine with it as well.

holmes/plugins/toolsets/internet/notion.py
```diff
@@ -125,3 +126,12 @@ def __init__(self):
...
             is_default=False,
         )
+
+    def prerequisites_callable(self, config: Dict[str, Any]) -> Tuple[bool, str]:
+        if not config or not config.get("additional_headers", {}):
+            return (
+                False,
+                "Notion toolset is misconfigured. Authorization header is required.",
+            )
+        self.additional_headers = config.get("additional_headers", {})
```
moshemorad:
```suggestion
        self.additional_headers = config["additional_headers"]
```

moshemorad: We need to also the docs?

RoiGlinik: Not sure what you mean

moshemorad: We need to also update the docs on robusta?

moshemorad (approved):

PR #588, opened 2025-07-02
ROB-1516: add kafka lag eval

nherment:


tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/kafka_lag_instructions.md
```diff
@@ -0,0 +1,28 @@
+# Kafka Topic and Application Mapping
```
Sheeproid, line 1: It seems like too much information, the minimal set I'd expect is something like this for each topic:

Topic: finance
Producer(s): orders-app
Consumer(s): invoices-app

And same for payments topic.

Are all of the extra information really required?

tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/test_case.yaml
```diff
@@ -0,0 +1,19 @@
...
+#   kubectl port-forward svc/kafka 9092:9092 -n ask-holmes-namespace-55
+# and make sure the advertised hostname `kafka` is redirected to localhost:
+#   echo "127.0.0.1 kafka" | sudo tee -a /etc/hosts
+before_test: |
+  kubectl apply -f ./app/manifest.yaml && sleep 30
+after_test: kubectl delete -f ./app/manifest.yaml
+include_files:
+  - kafka_lag_instructions.md
+evaluation:
+  correctness: 0
```
Sheeproid, line 19: Why is it expected to fail? (Correctness 0)

1. Please write it on the #backend channel like it says in the issue
2. I think it would be useful to include a comment here so future readers of the test will have a basic understanding of why it's failing without having to run the test.

tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/test_case.yaml
```diff
@@ -0,0 +1,19 @@
+user_prompt: "What's wrong with the `finance` kafka topic?"
+tags:
+  - logs
+  - kafka
+  - runbooks
+expected_output: |
+  `invoices-app` is slow to process messages because email sending is slow
+# Manual port-forward is required:
```
Sheeproid, line 8: Can you include the port forwarding as a part of the before_test?

tests/llm/utils/constants.py
```diff
@@ -12,4 +12,5 @@
     "datetime",  # An eval that tests HolmesGPT's ability to rely on or correctly use date and/or time information
     "network",  # An eval related to networking issues
     "runbooks",  # An eval that relies on runbooks
+    "kafka",  # An eval that relies on runbooks
```
Sheeproid: fix the comment to match 'kafka'

holmes/plugins/toolsets/kafka.py
```diff
@@ -480,6 +561,7 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
 class KafkaToolset(Toolset):
     model_config = ConfigDict(arbitrary_types_allowed=True)
     clients: Dict[str, AdminClient] = {}
```
Sheeproid, line 559: I noticed we are using AdminClient which can do write actions in Kafka. We should probably ask the users to create a new Kafka user for us with limited permissions, like
    - Describe on topics
    - Describe on consumer groups
    - DescribeConfigs on topics/brokers
What do you think? If you agree, please change our docs now and test it or create a new ticket

holmes/plugins/toolsets/kafka.py
```diff
@@ -113,6 +122,27 @@ def get_kafka_client(self, cluster_name: Optional[str]) -> AdminClient:
             f"Failed to resolve Kafka client. No matching cluster: {cluster_name}"
         )
 
+    def get_bootstrap_servers(self, cluster_name: Optional[str]) -> str:
```
Sheeproid: cluster_name should be str, not Optional[str]. also, the if not cluster_name check should be removed

holmes/plugins/toolsets/kafka.py
```diff
@@ -113,6 +122,27 @@ def get_kafka_client(self, cluster_name: Optional[str]) -> AdminClient:
...
+        """
+        Retrieves the bootstrap servers for a given cluster.
+        """
+        if not self.toolset.kafka_config:
```
Sheeproid, line 129: I think this check is redundant, at this point if must be true because of the pre-requisites check

holmes/plugins/toolsets/kafka.py
```diff
@@ -113,6 +122,27 @@ def get_kafka_client(self, cluster_name: Optional[str]) -> AdminClient:
...
+            raise Exception("Kafka configuration not available")
+
+        if len(self.toolset.kafka_config.kafka_clusters) == 1:
```
Sheeproid: should be handled using the for loop below, what if the cluster_name does not match?

holmes/plugins/toolsets/kafka.py
```diff
@@ -352,17 +382,62 @@ def _invoke(self, params: Dict) -> StructuredToolResult:
...
+        consumer_config = {
+            "bootstrap.servers": bootstrap_servers,
+            "group.id": consumer_group_description.group_id,
+            "auto.offset.reset": "earliest",
+            "enable.auto.commit": False,  # Don't auto-commit to avoid side effects
+        }
+        consumer = Consumer(consumer_config)
+
+        # Get topic metadata to know which partitions exist
+        topic_metadata = client.list_topics(topic_name, timeout=10)
```
Sheeproid: I think we can call list_topics and create the topic_partitions only once and get it as a parameter to group_has_topic, saving api calls to client.list_topics.

holmes/plugins/toolsets/kafka.py
```diff
@@ -352,17 +382,62 @@ def _invoke(self, params: Dict) -> StructuredToolResult:
...
-        if len(member.assignment.topic_partitions) > 0:
-            member_topic_name = member.assignment.topic_partitions[0].topic
-            if topic_name == member_topic_name:
+        for topic_partition in member.assignment.topic_partitions:
+            if topic_partition.topic == topic_name:
+                return True
+
+    # Check committed offsets for the topic (handles inactive/empty consumer groups)
+    try:
+        # Try using the Consumer class to check committed offsets for the specific group
```
Sheeproid, line 397: there are a lot of comment here, have you used ClaudeCode? just curios

holmes/plugins/toolsets/kafka.py
nherment: The design of the toolsets does not ensure that tools can be invoked without the config being initialized. Recent changes and issues in tool caching does show that such issues happen due to code changes.
Toolsets should be redesigned so that it's not possible to invoke a tool without the toolset being initialized but without it there are static violations if this check is not added.

holmes/plugins/toolsets/kafka.py
nherment: Yes. In this case I kept the comment because it's hard to understand why this code is there without it.
What happens is that is a consumer is not very active (consuming messages slowly) then it was not displayed at all as part of the consumer groups. 
This code changes still return such consumer groups for any given topic along with their state (active or not)

tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/kafka_lag_instructions.md
nherment: I don't think this will change anything and the current content seems ok? The goal was to make this runbook similar to a company page that would document kafka topics.

Where holmes fail is that it does not see any issues with kafka so it does not investigate further.

tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/test_case.yaml
nherment: I did not find an easy way to do so except with PID files which I thought may be more hassle than worth.

holmes/plugins/toolsets/kafka.py
nherment: Good catch :+1:

tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/test_case.yaml
Sheeproid: Can you explain why a bash script won't work?

I'm not familiar with PID files, can you explain more??

holmes/plugins/toolsets/kafka.py
```diff
@@ -496,11 +490,13 @@ def _invoke(self, params: Dict) -> StructuredToolResult:
                         consumer_group_description_future.result()
                     )
                     bootstrap_servers = self.get_bootstrap_servers(kafka_cluster_name)
+                    topic_metadata = client.list_topics(topic_name, timeout=10)
```
Sheeproid, line 493: according to the docs I see, the return type is ClusterMetadata, thus it can be used here and in the group_has_topic instead of the 'Any' type

Sheeproid (approved):
Nice job. I have one nitpicking comment.

tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/test_case.yaml
nherment: Comment added to the test case file.
The message was sent to the backend channel on Friday.

tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/test_case.yaml
nherment: It would work but I think it will be more hassle than it is worth.

The approach would be to create 2 scripts:
1. Start the port forward and save the process id (PID) in a file
2. End the port forward by reading the pid file and killing the process.

One of the issues is that the port forward could stay hanging and never be terminated. 
In addition the setup requires to redirect `kafka` hostname to localhost which modifies the OS. There may be ways around it but it has to do with kafka's advertising of its own hostname for clients to connect.

holmes/plugins/toolsets/kafka.py
nherment: Good catch

tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/test_case.yaml
Sheeproid: @nherment now I understand what you mean. I still don't like the fact we're adding a test that cannot be easily run by automated test system, but let's not hold the PR for that now.

Issue #443, opened 2025-05-23
Add instructions for using HTTP API and Helm Chart


pavangudiwada:
### What would you like to be added?

It's one of the core features of HolmesGPT and doesn't have any instructions for how to use. 

### Why is this needed?

Provides users instructions on how to use HolmesGPT in-cluster or to integrate it with other tools.

YaoWang-MSFT:
thanks @pavangudiwada for opening this issue. this would really help us onboard HolmesGPT using Http API. any early guidance you can share by chance? it would be much appreciated. thank you.

higp14:
@pavangudiwada  Can you provide guidance on how to install holmesgpt via helm? I'm running into `ErrimagePull` on using image: holmes:latest
registry: robusta-dev

aantn:
@higp14 that should be robustadev without a dash. We're also not tagging latest right now.

So can you please try robustadev/holmes:0.11.5 and confirm that works?

aantn:
@higp14 one more thing please: if you saw robusta-dev registry incorrectly mentioned in our docs please let me know so I can fix it

pavangudiwada:
@aantn If you don't have any ideas for further improvements, we can close this.

aantn:
Closing!

For future reference, we have documented both HTTP API and Helm chart now - see https://robusta-dev.github.io/holmesgpt/

higp14:
Apologies for the delay @aantn . robustadev/holmes:0.11.5 worked. Thank you for your support.

aantn:
My pleasure, happy to help!

PR #592, opened 2025-07-03
Add HTTP API documentation

kristeey:
Add documentation of the Holmes HTTP API including all the endpoints, fields, example usage with curl, and example responses. Nice to have when working with the HTTP API directly, e.g. writing custom integrations.


pavangudiwada:
Partially helps with #443

HTTP_API.md
```diff
@@ -0,0 +1,318 @@
...
+    }
+  ],
+  "instructions": [...]
+}
+```
+
+---
+
+### 4. `/api/workload_health_chat` (POST)
+**Description:** Conversational interface for discussing the health of a workload. Maintains context across messages.
```
aantn: Can you remove 'Maintains context across messages.' - to me this implies the api is stateful but actually it's stateless and relies on the conversation_history passed to the chat for this to work.

HTTP_API.md
```diff
@@ -0,0 +1,318 @@
...
+    {"role": "assistant", "content": "To fix the CrashLoopBackOff, increase the memory limit for the pod and check for memory leaks in the application."}
+  ],
+  "tool_calls": [...],
+  "follow_up_actions": [...]
+}
+```
+
+---
+
+### 6. `/api/chat` (POST)
```
aantn: Could you put this API first please as it's the simplest?

aantn (dismissed):

aantn:
Excellent work @kristeey - thank you!
Left two small comments. Can you please review?

HTTP_API.md
```diff
@@ -0,0 +1,318 @@
+# HolmesGPT API Reference
+
```
aritraghosh, line 2: I would be curious to also know how I can find the Holmes URL for my environment

kristeey: This really depends on your environment and how you install Holmes. If running locally it could be a localhost url. If running in k8s behind a `Service` and `Ingress`, it would be whatever you want to be. For in-cluster communication it would be enough to create a k8s `Service` and use the in-cluster url i.e. `<service-name>.<namespace>.svc.cluster.local:<service-port>`

HTTP_API.md
kristeey: Makes sense, I reordered the endpoints such that the most used (in my opinion) comes first.

HTTP_API.md
aantn: @aritraghosh try running `poetry run python server.py` and it will show you the url. Typically it's localhost:5050

@kristeey worth adding this to the docs?

pavangudiwada:
@kristeey Could you please sign the CLA

CleanShot 2025-07-10 at 16 59 34

kristeey:
> @kristeey Could you please sign the CLA
> 
> 
> 
> CleanShot 2025-07-10 at 16 59 34
> 
> 

Done 👍

aantn (approved):
Lgtm. Thank you very much!

PR #595, opened 2025-07-04
fix: make opensearch env var and config description more generic

nherment:


holmes/plugins/toolsets/opensearch/opensearch_utils.py
```diff
@@ -149,24 +149,22 @@ def build_query(
 class BaseOpenSearchToolset(Toolset):
```
moshemorad, line 148: Please add test that enabled this toolset and make sure it is enabled correctly after init with the relevant config/env vars.

moshemorad (changes requested):

nherment:
@coderabbitai Check the pre-commit results in the github checks and provide fixes for each issue

PR #601, opened 2025-07-07
new evals for logging and more

aantn:


tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/manifest.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: production-55
```
Sheeproid: please prefix all the namespaces with the name of the test like we did in other places.
e.g. ask-holmes-namespace-53

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/manifest.yaml
```diff
@@ -0,0 +1,45 @@
...
+          echo "$(date) INFO: Connecting to database"
+          echo "$(date) WARN: Database connection slow"
+          echo "$(date) ERROR: Failed to connect to database: connection timeout"
+          echo "$(date) ERROR: Retrying database connection..."
+          echo "$(date) ERROR: Database connection failed after 3 retries"
+          echo "$(date) INFO: Switching to backup database"
+          echo "$(date) ERROR: Backup database also unavailable"
+          echo "$(date) ERROR: Application cannot start without database"
+          echo "$(date) FATAL: Shutting down application"
+          sleep 300
```
Sheeproid: Change it to hold/sleep forever so the log fetching will always work.
This will enable us to run the test multiple times with just one setup.

tests/llm/fixtures/test_ask_holmes/56_fetch_error_logs_no_errors/manifest.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: v1
+kind: Namespace
```
Sheeproid: see my comments in test 55

tests/llm/fixtures/test_ask_holmes/57_fetch_logs_no_namespace_confusion/manifest.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: test-57
+---
```
Sheeproid: see my comments in test 55

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case_with_timeout.yaml.example
```diff
@@ -0,0 +1,8 @@
+user_prompt: "post last 10 logs that contain \"ERROR\" from payment-service in production-55 namespace"
+expected_output: "ERROR: Failed to connect to database"
+before_test: 'kubectl apply -f ./manifest.yaml'
+after_test: 'kubectl delete -f ./manifest.yaml'
+timeout: 120  # Override default 70s timeout with 120s for slow commands
```
Sheeproid: is this 'timeout' a new thing? why do we need it here?

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case.yaml
```diff
@@ -0,0 +1,7 @@
+user_prompt: "post last 10 logs that contain \"ERROR\" from payment-service in production-55 namespace"
+expected_output: "ERROR: Failed to connect to database"
```
Sheeproid: how many logs line do you expect here? please state it clearly (you requested 10 but there less than 10 in the actual logs)

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case.yaml
```diff
@@ -0,0 +1,7 @@
+user_prompt: "post last 10 logs that contain \"ERROR\" from payment-service in production-55 namespace"
```
Sheeproid: please run 100 times and add the mock

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case.yaml
```diff
@@ -0,0 +1,7 @@
...
+expected_output: "ERROR: Failed to connect to database"
+before_test: 'kubectl apply -f ./manifest.yaml'
```
Sheeproid: please run the eval 100 times
I'm a bit concerned that the expected output here it too vague.

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case_with_timeout.yaml.example
```diff
@@ -0,0 +1,8 @@
+user_prompt: "post last 10 logs that contain \"ERROR\" from payment-service in production-55 namespace"
+expected_output: "ERROR: Failed to connect to database"
```
Sheeproid: what is this file?

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case.yaml
```diff
@@ -0,0 +1,7 @@
...
+after_test: 'kubectl delete -f ./manifest.yaml'
+evaluation:
+  correctness: 1
+generate_mocks: False
```
Sheeproid: please add the relevant 'tags'. possible 'logs' tag

tests/llm/fixtures/test_ask_holmes/57_fetch_logs_no_namespace_confusion/test_case.yaml
```diff
@@ -0,0 +1,8 @@
+# we ask about app-57 namespace but the deployment is in test-57 namespace
+user_prompt: "post last 10 logs that contain \"ERROR\" from video-streamer in app-57 namespace"
```
Sheeproid: please see comments in test 55

tests/llm/fixtures/test_ask_holmes/57_fetch_logs_no_namespace_confusion/test_case.yaml
```diff
@@ -0,0 +1,8 @@
+# we ask about app-57 namespace but the deployment is in test-57 namespace
```
Sheeproid: probably you'd like to add the 'misleading-history' tag. maybe we should change the tag to 'misleading-context'

because the test here is not about logs at all, right? it's about finding a resource

Sheeproid: the test here is not about logs at all, right? it's about finding the right resource. maybe the name of the test should be change to make it more clear (not sure how though)

Sheeproid (changes requested):

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/manifest.yaml
aantn: My rationale is that `ask-holmes-namespace-53` is leaking information to holmes about itself and what it is testing. Whereas production-55 is still unique but doesn't leak.

It probably doesn't matter for our use cases, but there's increasing evidence that llms understand when they're being evaluated and change behaviour accordingly - https://www.arxiv.org/abs/2505.23836

So if you don't object I would rather keep production-55

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/manifest.yaml
aantn: Fixed

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case.yaml
aantn: done

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case.yaml
aantn: done

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case_with_timeout.yaml.example
aantn: it's a mistake, removing

tests/llm/fixtures/test_ask_holmes/55_fetch_error_logs_with_errors/test_case_with_timeout.yaml.example
aantn: yes, it leaked over from another branch - ignore it please (and removing)

tests/llm/fixtures/test_ask_holmes/57_fetch_logs_no_namespace_confusion/test_case.yaml
aantn: I think it's a little different To me, it's similar to asking holmes about problems with cilium pods, when there are none and the user has calico instead of cilium. Can holmes go beyond giving a technically correct answer and understand the user's real intent. Not sure what to call it, but if you agree please suggest ideas.

aantn: I'll rename the test to `wrong-namespace`. I will also simplify the test to not filter on ERROR logs. I want this test to only check namespace confusion nothing else.

tests/llm/fixtures/test_ask_holmes/64_keda_vs_hpa_confusion/test_case.yaml
```diff
@@ -0,0 +1,13 @@
+user_prompt: 'Why isnt KEDA scaling the invoice-generator deployment?'
+description: Verify Holmes goes beyond 'technically correct' answers and provides a useful answer based on the user's actual intent
```
Sheeproid, line 2: is this new?

tests/llm/utils/constants.py
```diff
@@ -17,6 +17,7 @@
     "misleading-history",  # An eval with misleading historical data
     "k8s-misconfig",  # An eval related to Kubernetes misconfigurations
     "chain-of-causation",  # An eval that tests chain-of-causation analysis
-    "slackbot",  # An eval related to Slack bot functionality
+    "slackbot",  # An eval related to Slack bot functionality,
```
Sheeproid, line 20: I think you also need to add this a marker

Sheeproid (changes requested):
see the comment about the marker

tests/llm/fixtures/test_ask_holmes/64_keda_vs_hpa_confusion/test_case.yaml
aantn: yeah its just me documenting what we're trying to accomplish here in a human readable format. not used for anything

tests/llm/utils/constants.py
aantn: What do you mean? slackbot is an existing market - I added the `understanding-real-intent` marker and it was added in both places.

PR #605, opened 2025-07-07
Improve tools-output and add slash commands for it

aantn:


holmes/interactive.py
```diff
@@ -88,15 +153,43 @@ def run_interactive_loop(
...
+                    )
+                    console.print(
+                        "  [bold]/tools[/bold] - Show available toolsets and their status"
+                    )
+                    console.print(
+                        "  [bold]/toggle-tools[/bold] - Toggle tool output display on/off"
+                    )
+                    console.print(
+                        "  [bold]/dump-tools[/bold] - Dump all tool outputs from last response"
+                    )
```
mainred: I guess these can be achieved by a loop by iterating over the supported commands

holmes/interactive.py
```diff
@@ -88,15 +153,43 @@ def run_interactive_loop(
                 if command == SlashCommands.EXIT.value:
                     return
                 elif command == SlashCommands.HELP.value:
-                    console.print(f"Available commands: {ALL_SLASH_COMMANDS}")
+                    console.print("[bold cyan]Available commands:[/bold cyan]")
```
mainred: I guess it's almost time to group the color we used into one place to comment how each color is used in which cases.
and how about we override the console.print to achieve something like `childconsole.print(color string,  formated_string)`?

holmes/interactive.py
```diff
@@ -12,31 +12,42 @@
...
+from holmes.core.tool_calling_llm import ToolCallingLLM, ToolCallResult
 from holmes.core.tools import pretty_print_toolset_status
 
 
 class SlashCommands(Enum):
     EXIT = "/exit"
     HELP = "/help"
     RESET = "/reset"
-    TOOLS = "/tools"
+    TOOLS_CONFIG = "/config"
```
mainred, line 23: /config or /tools?

holmes/interactive.py
```diff
@@ -46,6 +57,56 @@ def get_completions(self, document, complete_event):
...
+    output_str = result.get_stringified_data()
+
+    color = result.status.to_color()
+    MAX_CHARS = 500
+    if len(output_str) == 0:
+        content = f"[{color}]<empty>[/{color}]"
+    elif len(output_str) > 100:
+        truncated = output_str[:MAX_CHARS].strip()
+        remaining_chars = len(output_str) - MAX_CHARS
+        content = f"[{color}]{truncated}[/{color}]\n\n[dim]... truncated ({remaining_chars:,} more chars)[/dim]"
```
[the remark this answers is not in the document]
mainred: +1

holmes/interactive.py
aantn: I chose `/config` because `/tools` could be confused with showing output from the previously ran tools (`/output` below). If you have better ideas on the names for each feel free to share

holmes/interactive.py
aantn: Could we use constants instead? i.e. TOOL_COLOR="cyan", AI_COLOR="...", USER_COLOR="..."

holmes/interactive.py
aantn: yes, good idea

holmes/interactive.py
mainred: Oh, I thought you replaced TOOLS with TOOLS_CONFIG, because we don't have /config around L159.

holmes/interactive.py
mainred: Sounds good.

holmes/interactive.py
aantn: done

holmes/interactive.py
aantn: Fixed

aantn:
@mainred I've fixed - can you re-review please?

PR #611, opened 2025-07-08
Create runbook for troubleshooting AKS upgrade failures

julia-yin:
Adding a new folder under "runbooks" for upgrade issues + new runbook for troubleshooting AKS upgrade failures.

### DIFF
--- a/tests/plugins/toolsets/azure_sql/test_azure_sql.py
+++ b/tests/plugins/toolsets/azure_sql/test_azure_sql.py
@@ -1,0 +1,253 @@
+from unittest.mock import MagicMock, patch
+
+from holmes.core.tools import ToolsetStatusEnum
+from holmes.plugins.toolsets.azure_sql.azure_sql_toolset import AzureSQLToolset
+
+
+class TestAzureSQLToolset:
+    """Tests for Azure SQL toolset initialization and health checks."""
+
+    def test_nominal_config_successful_initialization(self):
+        """Test that toolset initializes correctly with valid configuration."""
+        # Prepare valid configuration
+        config = {
+            "tenant_id": "test-tenant-id",
+            "client_id": "test-client-id",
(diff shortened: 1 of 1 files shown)


holmes/plugins/runbooks/catalog.json
```diff
@@ -2,8 +2,13 @@
   "catalog": [
     {
       "update_date": "2025-06-17",
-      "description": "Runbook to investigate DNS resolution issue on Kubernetes cluster",
+      "description": "Runbook to investigate DNS resolution issue in Kubernetes clusters",
       "link": "networking/dns_troubleshooting_instructions.md"
+    },
+    {
+      "update_date": "2025-07-08",
+      "description": "Runbook to troubleshoot upgrade issues in Azure Kubernetes Service clusters",
```
arikalon1, line 10: @julia-yin when do we want Holmes to read and use this runbook? What would the user ask?

julia-yin: This runbook is used for troubleshooting failed upgrades / upgrade issues in AKS. If a user's cluster is in a failed state or a recent upgrade operation failed, they could ask something like "why did my AKS cluster upgrade fail?" and get a detailed analysis of the failure and remediation steps.

arikalon1:
@julia-yin I think the runbook file name is missing the `.md` suffix which causes the test to fail

mainred (approved):

holmes/plugins/runbooks/catalog.json
mainred: Yeah, AKS provides API to upgrade the k8s versions of managed cluster and nodepool. This runbook provides steps to investigate these kind of issues.

mainred:
I am going to merge this runbook though we don't have a standard format of runbook yet.
We can format and evaluate if necessary this runbook later.

PR #615, opened 2025-07-09
feat(mcp-server): introduce stdio mcp server

mainred:


mainred:
cc @arikalon1

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -135,3 +216,75 @@ async def _get_server_tools(self):
...
+        return {"command": "python", "args": ["/path/to/mcp_server.py"]}
+
+
+def get_mcp_toolset_from_config(config: dict[str, Any], name: str) -> BaseMCPToolset:
+    if not config:
+        raise ValueError("Config must not be empty")
+    # Prefer explicit mcp type set in nested 'config' for clarity.
+    # Supported values: 'sse' (remote SSE server) and 'stdio' (stdio server).
+    mcp_config = config.get("config", {})
+    mcp_type = mcp_config.get("type")
```
moshemorad, line 272: Shouldn't the default be sse?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -135,3 +216,75 @@ async def _get_server_tools(self):
...
+
+    if mcp_type == "stdio":
+        return StdioMCPToolset(**config, name=name)
+
+    # Backwards compatibility: still support using url or command top-level keys.
+    url = config.pop("url", None)
+    if mcp_type == "sse":
```
moshemorad, line 280: If the type will set to sse and include url we will not add it to the config in the current flow. Is that intended?

holmes/plugins/toolsets/mcp/toolset_mcp.py
mainred: If the user specify the url, it's ok but if the user set no type and no url, we'll raise ValidationError instead of 
`ValueError(
        "MCP Server config must include a transport type ('sse' or 'stdio') "
        "either in 'config.type' or by providing transport-specific keys."
    )`

mainred: I can backfill the type to sse when url is set but type is empty

docs/data-sources/remote-mcp-servers.md
```diff
@@ -143,34 +199,57 @@ Use this config according to your use case.
...
-    holmes ask "Using dynatrace what issues do I have in my cluster?"
+    holmes ask -t custom_toolset.yaml "Using dynatrace what issues do I have in my cluster?"
     ```
 
 ```yaml-helm-values
 mcp_servers:
   mcp_server_1:
+    type: "sse"
     description: "Dynatrace observability platform. Bring real-time observability data directly into your development workflow."
     url: "http://dynatrace-mcp.default.svc.cluster.local:8003"
```
moshemorad: Shouldn't this be passed to config section?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -68,64 +93,122 @@ def create(cls, url: str, tool: MCP_Tool, headers: Optional[Dict[str, str]] = No
...
+    def get_parameterized_one_liner(self, params: Dict) -> str:
+        return f"Call mcp server {self.url} tool {self.name} with params {str(params)}"
 
-        return parameters
+
+class StdioMCPTool(BaseMCPTool):
+    server_params: StdioServerParameters
+
+    async def _invoke_async(self, params: Dict) -> StructuredToolResult:
+        async with stdio_client(self.server_params) as (read_stream, write_stream):
```
moshemorad, line 105: Do we really want to create the process for each tool call?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -68,64 +93,122 @@ def create(cls, url: str, tool: MCP_Tool, headers: Optional[Dict[str, str]] = No
...
+                merged_text = " ".join(
+                    c.text for c in tool_result.content if c.type == "text"
+                )
+                return StructuredToolResult(
+                    status=(
+                        StructuredToolResultStatus.ERROR
+                        if tool_result.isError
+                        else StructuredToolResultStatus.SUCCESS
+                    ),
+                    data=merged_text,
```
moshemorad, line 120: what will merged_text will contain if the tool_result is error? should we put it on the error field in StructuredToolResult?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -68,64 +93,122 @@ def create(cls, url: str, tool: MCP_Tool, headers: Optional[Dict[str, str]] = No
...
+        return f"Call stdio mcp server tool {self.name} with params {str(params)}"
 
 
-class RemoteMCPToolset(Toolset):
-    url: AnyUrl
-    tools: List[RemoteMCPTool] = Field(default_factory=list)  # type: ignore
+class BaseMCPToolset(Toolset):
+    """Base class for MCP toolsets with shared functionality"""
+
+    name: str
```
moshemorad: why we need to add name field? `Toolset` already contains it.

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -68,64 +93,122 @@ def create(cls, url: str, tool: MCP_Tool, headers: Optional[Dict[str, str]] = No
...
+    @property
+    def url(self) -> Optional[str]:
+        raw = self.config and self.config.get("url")
+        if isinstance(raw, str):
+            u = raw.rstrip("/")
+            return u if u.endswith("/sse") else f"{u}/sse"
+        return raw
+
+    def _get_server_type(self) -> str:
+        return "remote"
```
moshemorad: Shouldn't this be sse? what remote means in that case?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -135,3 +218,75 @@ async def _get_server_tools(self):
...
+    mcp_type = mcp_config.get("type")
+
+    if mcp_type == "stdio":
+        return StdioMCPToolset(**config, name=name)
+
+    # Backwards compatibility: still support using url or command top-level keys.
+    url = config.pop("url", None)
+    if mcp_type == "sse":
+        return RemoteMCPToolset(**config, name=name)
+    if url is not None:
```
moshemorad, line 285: Consider add here deprecation message.

holmes/plugins/toolsets/mcp/toolset_mcp.py
moshemorad: I think that keep it backward compatible is a good idea. The thing that im not sure is clear from the flow is what will happen if the user did provided type:sse and also url: xxx not under the config field. Looks like it won't work right?

tests/test_mcp_toolset.py
```diff
@@ -82,20 +89,125 @@ async def mock_get_server_tools():
 
 def test_mcpserver_headers(monkeypatch):
```
moshemorad, line 90: Please add one test for the backward compatibility of url in the root and not part of the config field.

holmes/plugins/toolsets/mcp/toolset_mcp.py
mainred: You mean this kind of setting?  I thought the compatibility is the user either set aks_mcp.url  or use the new way to set the url under config. If the user mix the two setting, the url is not used. Let me update the code the achieve this.
```
mcp_servers:
  aks_mcp:
    description: "AKS MCP server"
    url: "http://127.0.0.1:8000/sse"
    config:
      type: "sse"
```

holmes/plugins/toolsets/mcp/toolset_mcp.py
mainred: The result.content is usually the error, for example.  I can put it on the error field

mcp.NewToolResultError(fmt.Sprintf("%s\n%s", err.Error(), result)), nil
```
func NewToolResultError(text string) *CallToolResult {
	return &CallToolResult{
		Content: []Content{
			TextContent{
				Type: ContentTypeText,
				Text: text,
			},
		},
		IsError: true,
	}
}
```

holmes/plugins/toolsets/mcp/toolset_mcp.py
mainred: Good point, we can reuse the same session.

mainred: Sharing the session keeps raise the following error. I need more time to investigate
```
  File "/home/azureuser/.cache/pypoetry/virtualenvs/holmesgpt-N4JKwMcz-py3.12/lib/python3.12/site-packages/mcp/client/stdio/__init__.py", line 180, in
stdio_client
    async with (
  File "/home/azureuser/.cache/pypoetry/virtualenvs/holmesgpt-N4JKwMcz-py3.12/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 778, in
__aexit__
    if self.cancel_scope.__exit__(type(exc), exc, exc.__traceback__):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/azureuser/.cache/pypoetry/virtualenvs/holmesgpt-N4JKwMcz-py3.12/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 457, in
__exit__
    raise RuntimeError(
RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
``

holmes/plugins/toolsets/mcp/toolset_mcp.py
mainred: Done by L277

Avi-Robusta:
we updated mcp_toolset to support http-streamable and opened a seperate pr for stdio support
https://github.com/HolmesGPT/holmesgpt/pull/1125
Closing this as its redundant now

PR #630, opened 2025-07-14
ROB-1078: add cluster name to system prompt

nherment:


holmes/plugins/prompts/_general_instructions.jinja2
```diff
@@ -1,5 +1,8 @@
 # In general
 
+{% if cluster_name -%}
```
moshemorad, line 3: Thinking here loud, we think it will have also impact when work on issues which aren't k8s? 

If not, do you think that moving it to the `If investigating Kubernetes problems` will be more relevant?

tests/llm/test_ask_holmes.py
```diff
@@ -191,7 +192,15 @@ def ask_holmes(test_case: AskHolmesTestCase, parent_span: Optional[Span]) -> LLM
     )
 
     chat_request = ChatRequest(ask=test_case.user_prompt)
+    config = Config()
```
moshemorad: how does the cluster_name is being loaded in that flow? 
I see you commented the next lines + also didn't used load_from_env which actually loading the cluster name.

maybe it worth to add check to the test that the cluster name was loaded?

tests/llm/fixtures/test_ask_holmes/57_cluster_name_confusion/test_case.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - logs
+before_test: |
+  kubectl apply -f https://raw.githubusercontent.com/robusta-dev/kubernetes-demos/main/image_pull_backoff/no_such_image.yaml
+  sleep 20
+after_test: |
+  kubectl delete -f https://raw.githubusercontent.com/robusta-dev/kubernetes-demos/main/image_pull_backoff/no_such_image.yaml
+expected_output: "failure to pull image `yourcompany/crw:latest`"
+evaluation:
+  correctness: 1
+generate_mocks: True
```
moshemorad: Left true intentionally?

tests/llm/test_ask_holmes.py
nherment: Updated code below. this is how the cluster name is loaded.
This is a test code, I don't think we need to add a check that the cluster name is loaded. Well actually, it would be beneficial to better test the prompt and how it gets built but it should start with refactoring that part. The conversation.py code has a lot of room for improvement and that in turn will make testing it easier.

holmes/plugins/prompts/_general_instructions.jinja2
moshemorad: @nherment i think you might have missing this comment. Can you please take a look and let me know WDYT?

moshemorad (approved):

moshemorad (dismissed):

moshemorad (approved):

PR #633, opened 2025-07-15
ROB-1643: document eval tags

nherment:
image

image


docs/evals-writing.md
```diff
@@ -236,8 +236,55 @@ evaluation:
 
 ## Tagging
 
-Evals are tagged for organisation and reporting purposes.
-The valid tags are defined [here](../tests/llm/utils/constants.py).
+Evals support tags for organization, filtering, and reporting purposes. Tags help categorize tests by their characteristics and enable selective test execution.
+
+### Available Tags
+
+The valid tags are defined in [`tests/llm/utils/constants.py`](../tests/llm/utils/constants.py):
```
Sheeproid: what about the pytest markers? right now tags have to be added to both places which is a bit cumbersome. Either we fix this behavior in thie PR or we should document it here.

nherment: the pytest warnings are not part of this ticket so given the pressure for the POCs and the guidelines to prioritize POCs I've not addressed them. A good solution would be a pre-commit hook that automatically updates the pytest marks based on the tags listed in the code.

Sheeproid (approved):

Sheeproid (approved):

PR #635, opened 2025-07-15
ROB-1546 health check json

RoiGlinik:
Fixes for workload_health endpoint.
Improve and test json result schema.
Add tests flow for workload_health


tests/llm/utils/mock_dal.py
```diff
@@ -72,6 +72,9 @@ def _get_mock_file_path(self, entity_type: str):
     def get_global_instructions_for_account(self) -> Optional[Instructions]:
         return None
 
+    def get_workload_issues(self, *args) -> list:
```
RoiGlinik, line 75: In general we might be better to convert this to a tool in the robusta Toolset

tests/llm/fixtures/test_workload_health/01_crashpod/test_case.yaml
```diff
@@ -0,0 +1,11 @@
+expected_output:
+  - The crashpod or might mention the pod 'crashpod-9688789bc-lgjzj'. Mentioning one of the reasons for a CrashLoopBackOff state. 1. due to command 'wget' exiting with code 125 2. having recursive calls 3. StackOverflowError in the application.
```
moshemorad, line 2: Not sure i understand why it is search from crashpod-XXXX. 

The scenario setup & teardown (berfore_test & after_test) are using yaml which creates pod with name payment-processing-worker no?
also looks like the issue from the logs of the pod are that it is missing env var but here i expects failed wget error. 

What am i missing? it might be that the before and after test aren't aligned to the scenario?

tests/llm/fixtures/test_workload_health/01_crashpod/kubectl_describe.txt
```diff
@@ -0,0 +1,42 @@
...
+  Containers:
+   crashpod:
+    Image:      busybox
+    Port:       <none>
+    Host Port:  <none>
+    Command:
+      sh
+    Args:
+      -c
+      wget -O - https://gist.githubusercontent.com/odyssomay/1078370/raw/35c5981f8c139bc9dc02186f187ebee61f5b9eb9/gistfile1.txt 2>/dev/null; exit 125;
```
moshemorad, line 24: Please make sure to remove this from here. We so couple of times that when holmes use describe and he see the commands as part of the describe it effects the actual result.

In real work use cases those args won't be here so better to remove them.

holmes/core/models.py
```diff
@@ -155,3 +155,27 @@ class WorkloadHealthChatRequest(ChatRequestBaseModel):
     ask: str
     workload_health_result: WorkloadHealthInvestigationResult
     resource: dict
+
+
+workload_health_structured_output = {
+    "type": "json_schema",
+    "json_schema": {
+        "name": "WorkloadHealthResult",
+        "strict": False,
```
moshemorad, line 164: Why we don't prefer to use string in that case?

server.py
```diff
@@ -217,9 +218,11 @@ def workload_health_check(request: WorkloadHealthRequest):
             },
         )
 
-        structured_output = {"type": "json_object"}
         ai_call = ai.prompt_call(
-            system_prompt, request.ask, HOLMES_POST_PROCESSING_PROMPT, structured_output
+            system_prompt,
+            request.ask,
+            HOLMES_POST_PROCESSING_PROMPT,
+            workload_health_structured_output,
```
moshemorad, line 227: Can you please elaborate what you have found as part of the investigation? we assume that the issue is only that it doesn't return the response in the expected format?

holmes/plugins/prompts/kubernetes_workload_ask.jinja2
```diff
@@ -53,9 +48,7 @@ In general:
 Style guide:
 * Be painfully concise.
 * Leave out "the" and filler words when possible.
-* Be terse but not at the expense of leaving out important data like the root cause and how to fix.
-* if asked by Global Instructions or instructions in a triple single quotes section to explicitly include something in the answer, don't leave it out.
-* return a json object with the following schema as a result:
+* your answer should ONLY return a json object with the following schema as a result:
 {
```
moshemorad, line 52: A. We still need to add it here if we are passing it as response_format?
B. Can we load it from the models.py and use it here instead of defining it twice?

tests/llm/fixtures/test_workload_health/01_crashpod/test_case.yaml
RoiGlinik: yea just tought its the same crashpod, because of url . Ill change to the basic crashpod

server.py
RoiGlinik: Yea the issue was result returning json with issues or response not valid for json parse

holmes/core/models.py
RoiGlinik: Related to the way we want to show the response in our UI.

holmes/plugins/prompts/kubernetes_workload_ask.jinja2
RoiGlinik: A. yea it makes a difference according to my tests.
B. will update

RoiGlinik: I left it as is eventually, seems to give the best result otherwise it fails

PR #636, opened 2025-07-15
ROB-1723: datadog logs toolset

nherment:


holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -0,0 +1,351 @@
...
+
+
+DEFAULT_STORAGE_TIERS = [DataDogStorageTier.INDEXES]
+
+
+class DatadogConfig(BaseModel):
+    dd_api_key: str
+    dd_app_key: str
+    indexes: list[str] = ["*"]
+    site_api_url: str  # e.g. https://api.us3.datadoghq.com. c.f. https://docs.datadoghq.com/getting_started/site/
```
moshemorad: can we chagne it from str -> AnyUrl to validate that is indeed a valid url?

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -0,0 +1,351 @@
...
+
+        except Exception as e:
+            logging.exception("Failed during Datadog healthcheck")
+            return False, f"Healthcheck failed with exception: {str(e)}"
+
+    def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
+        if not config:
+            return (
+                False,
+                "Datadog toolset is misconfigured. 'dd_api_key' and 'dd_app_key' are required.",
```
moshemorad: isn't `site_api_url` is also required?

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -0,0 +1,351 @@
...
+            )
+
+        try:
+            dd_config = DatadogConfig(**config)
+            if not dd_config.storage_tiers:
```
moshemorad: this could be handled already by pydantic with min_items check.

https://docs.pydantic.dev/1.10/usage/types/#arguments-to-conlist

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -0,0 +1,351 @@
...
+                params=params.model_dump(),
+            )
+
+    def _perform_healthcheck(self) -> Tuple[bool, str]:
+        """
+        Perform a healthcheck by fetching a single log from Datadog.
+        Returns (success, error_message).
+        """
+        try:
+            logging.info("Performing Datadog configuration healthcheck...")
```
moshemorad, line 234: This won't be good enough? https://docs.datadoghq.com/api/latest/authentication/#validate-api-key
We would like that we are able to indeed fetch logs?

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -0,0 +1,351 @@
...
+            for storage_tier in self.dd_config.storage_tiers:
+                raw_logs = fetch_paginated_logs(
+                    params, self.dd_config, storage_tier=storage_tier
+                )
+
+                if raw_logs:
+                    # no need to try other storage tiers if the current one returned data
+                    break
+
+            if raw_logs:
```
moshemorad: Can be merged with the above check?

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -0,0 +1,351 @@
...
+        response = requests.post(url, headers=headers, json=payload, timeout=timeout)
+
+        if response.status_code == 200:
+            data = response.json()
+            cursor = _extract_cursor_for_next_paginated_api_call(data)
+
+            logs = data.get("data", [])
+            return logs, cursor
+
+        elif response.status_code == 429:
```
moshemorad: Can we please use tenacity for retries?

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -0,0 +1,351 @@
...
+            wait_time = min(
+                BASE_RETRY_DELAY * (2 ** (attempts_count - 1)), MAX_RETRY_DELAY
+            )
+
+            reset_time_header = response.headers.get("X-RateLimit-Reset")
+            if reset_time_header:
+                try:
+                    reset_time = int(reset_time_header)
+                    wait_time = max(0, reset_time) + 0.1
```
moshemorad: Product wise do we want to be able to limit this period? 
what is the max time that throttle can we get here? how it can effect the user expirience?

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -0,0 +1,351 @@
...
+            url=url, headers=headers, payload=payload, timeout=dd_config.request_timeout
+        )
+        logs += new_logs
+        payload["page"]["limit"] = calculate_page_size(params, dd_config, logs)
+
+    # logs are fetched descending order. Unified logging API follows the pattern of kubectl logs where oldest logs are first
+    logs.reverse()
+
+    if len(logs) > limit:
+        logs = logs[-limit:]
```
moshemorad, line 121: Consider add here log since it is indicating some bug. If it will be in error level it will also cause sentry issue and we will be able to monitor it.

moshemorad (changes requested):
Lookg good, left some few comments.

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
nherment: Yes, being able to fetch logs validate the labels mappings.

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
nherment: Rate limit is set by datadog. Unless I misunderstand what you are implying or suggesting we have to wait and can't magically bypass rate limits.
What do you suggest?

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
nherment: This is just defensive coding and I don't think an error log is warranted.

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -278,6 +200,25 @@ def fetch_pod_logs(self, params: FetchPodLogsParams) -> StructuredToolResult:
                 f"Failed to query Datadog logs for params: {params}", exc_info=True
             )
 
+            # Check if this is a RetryError from tenacity wrapping a rate limit error
+            from tenacity import RetryError
```
moshemorad, line 204: Just to make sure i understand correctly, you would like to get here the last exception the raise instead of the RetryError right? In that case i think you can use reraise https://tenacity.readthedocs.io/en/latest/index.html?highlight=reraise#error-handling

tests/plugins/toolsets/datadog/logs/test_utils.py
```diff
@@ -16,7 +17,7 @@
             DatadogConfig(
                 dd_api_key="xyz",
                 dd_app_key="xyz",
-                site_api_url="xyz",
+                site_api_url=AnyUrl("https://example.com"),
```
moshemorad, line 20: Is the AnyUrl is really needed? I expect it to work with even a string other wise it might be broken when setting it in the config using yaml. Can you please double check to make sure we didn't miss anything?

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -0,0 +1,290 @@
...
+    for raw_log_item in raw_logs:
+        message = raw_log_item.get("attributes", {}).get(
+            "message", json.dumps(raw_log_item)
+        )
+        logs.append(message)
+
+    return "\n".join(logs)
+
+
+class DatadogToolset(BasePodLoggingToolset):
```
moshemorad, line 137:
```suggestion
class DatadogLogsToolset(BasePodLoggingToolset):
```

holmes/plugins/toolsets/__init__.py
```diff
@@ -14,7 +14,7 @@
 from holmes.plugins.toolsets.coralogix.toolset_coralogix_logs import (
     CoralogixLogsToolset,
 )
-from holmes.plugins.toolsets.datadog import DatadogToolset
+from holmes.plugins.toolsets.datadog.toolset_datadog_logs import DatadogToolset
```
moshemorad, line 17: Shouldn't this be added to the loading in line 59?

tests/plugins/toolsets/datadog/logs/test_utils.py
nherment: yes it works.

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
nherment: Will implement this from the traces branch

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
nherment: This change has been made in the metrics branch. It's not that relevant in this branch because there is a single datadog toolset

holmes/plugins/toolsets/__init__.py
nherment: it's already there?

Issue #642, opened 2025-07-17
Add /run command to interactive mode


aantn:
Allow users to explicitly specify a command to run themselves and then inject the command and output into the conversation itself.

This gives users the ability to interact with and guide the investigation themselves.

aantn:
Relates to #616 but is different

aantn:
Done :)

PR #644, opened 2025-07-17
Interactive mode: add /run and /clear commands + minor improvements

aantn:
Implements #642 - thanks to @mainred for the fantastic idea

### DIFF
--- a/holmes/interactive.py
+++ b/holmes/interactive.py
@@ -1,4 +1,5 @@
 import logging
+import subprocess
 from enum import Enum
 from typing import Optional, List
 from pathlib import Path
@@ -23,6 +24,8 @@ class SlashCommands(Enum):
     TOOLS_CONFIG = "/config"
     TOGGLE_TOOL_OUTPUT = "/toggle-output"
     SHOW_OUTPUT = "/output"
+    CLEAR = "/clear"
+    RUN = "/run"
 
 
 SLASH_COMMANDS_REFERENCE = {
@@ -32,6 +35,8 @@ SLASH_COMMANDS_REFERENCE = {
     SlashCommands.TOOLS_CONFIG.value: "Show available toolsets and their status",
     SlashCommands.TOGGLE_TOOL_OUTPUT.value: "Toggle tool output display on/off",
     SlashCommands.SHOW_OUTPUT.value: "Show all tool outputs from last response",
+    SlashCommands.CLEAR.value: "Clear the terminal screen",
+    SlashCommands.RUN.value: "Run a bash command and optionally share with LLM",
 }
 
 ALL_SLASH_COMMANDS = [cmd.value for cmd in SlashCommands]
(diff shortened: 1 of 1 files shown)


holmes/interactive.py
```diff
@@ -20,18 +31,28 @@ class SlashCommands(Enum):
...
 
 
 SLASH_COMMANDS_REFERENCE = {
     SlashCommands.EXIT.value: "Exit interactive mode",
     SlashCommands.HELP.value: "Show help message with all commands",
     SlashCommands.RESET.value: "Reset the conversation context",
     SlashCommands.TOOLS_CONFIG.value: "Show available toolsets and their status",
-    SlashCommands.TOGGLE_TOOL_OUTPUT.value: "Toggle tool output display on/off",
-    SlashCommands.SHOW_OUTPUT.value: "Show all tool outputs from last response",
+    SlashCommands.TOGGLE_TOOL_OUTPUT.value: "Toggle auto-display of tool outputs after responses",
```
mainred, line 49: I am thinking about group the Enum with the references. This is what I get from chatgpt

```
from enum import Enum

class Status(Enum):
    SUCCESS = (0, "The operation was successful")
    FAILURE = (1, "The operation failed")
    PENDING = (2, "The operation is still pending")

    def __init__(self, code, explanation):
        self.code = code
        self.explanation = explanation

# Usage
print(Status.SUCCESS)               # Status.SUCCESS
print(Status.SUCCESS.code)          # 0
print(Status.SUCCESS.explanation)   # The operation was successful
```

holmes/interactive.py
```diff
@@ -165,31 +634,116 @@ def run_interactive_loop(
...
+                    ].strip()
+                    if not bash_command:
                         console.print(
-                            f"[bold {ERROR_COLOR}]No tool calls available from the last response.[/bold {ERROR_COLOR}]"
+                            f"[bold {ERROR_COLOR}]Usage: /run <bash_command>[/bold {ERROR_COLOR}]"
                         )
                         continue
 
-                    display_tool_calls(last_response.tool_calls, console)
+                    result = None
```
mainred: why not use one function to handle the /run command?

holmes/interactive.py
```diff
@@ -165,31 +634,116 @@ def run_interactive_loop(
...
+                                Panel(
+                                    output,
+                                    padding=(1, 2),
+                                    border_style="white",
+                                    title="Command Output",
+                                    title_align="left",
+                                )
+                            )
+
+                    except KeyboardInterrupt:
```
mainred: This is great. I tried to run some command that hangs. Catching KeyboardInterrupt helps end the command without breaking the interactive mode.

holmes/interactive.py
```diff
@@ -165,31 +634,116 @@ def run_interactive_loop(
...
+                        error_message = "Command interrupted by user"
+                        console.print(
```
mainred: Should we ignore the prompt to share the command with LLM for cases ends unexpectedly like this?
```
User: /run az login
Running: az login
^CCommand interrupted by user
Share ran a command with LLM? (Y/n): n
User:
```

holmes/interactive.py
```diff
@@ -156,7 +612,20 @@ def run_interactive_loop(
                 user_input = session.prompt(input_prompt, style=style)  # type: ignore
 
             if user_input.startswith("/"):
-                command = user_input.strip().lower()
+                original_input = user_input.strip()
+                command = original_input.lower()
+
+                # Handle prefix matching for slash commands
+                matches = [cmd for cmd in ALL_SLASH_COMMANDS if cmd.startswith(command)]
```
mainred, line 692: It's not possible to have an ambiguous command so far considering we don't have command sharing a prefix. but if in the future we have commands with one prefix, for example, we have /run /runcommand, and the user input /run, this will be an ambiguous command unexpectedly.
So I think we'd use exact match instead of fuzzy matching to get the matched command.

holmes/interactive.py
aantn: Thanks :)

holmes/interactive.py
aantn: I think it's OK to leave as is, because you might interrupt a command that is taking forever than ask the LLM why the command was stuck

holmes/interactive.py
aantn: I really like the fuzzy because you can type /r and it does /run - so it's shorter. Lets leave for now and revisit in the future if we need?

holmes/interactive.py
aantn: good idea, fixed.

holmes/interactive.py
aantn: Fixed, good idea!

aantn:
@mainred I addressed all comments, if its good to merge please approve!

mainred (approved):

PR #645, opened 2025-07-17
ROB-1740: datadog metrics toolset

nherment:
Should be merged after https://github.com/robusta-dev/holmesgpt/pull/636


holmes/plugins/toolsets/datadog/datadog_api.py
```diff
@@ -92,20 +118,34 @@ def __call__(self, retry_state):
...
+        )
 
     if response.status_code == 200:
         data = response.json()
-        cursor = extract_logs_cursor(data)
 
-        logs = data.get("data", [])
-        return logs, cursor
+        if method == "POST" and "data" in data:
+            cursor = extract_logs_cursor(data)
```
moshemorad, line 140: This util is general api so i think that a long as this method is general function we need to remove logs mention in the names etc.

holmes/plugins/toolsets/datadog/toolset_datadog_logs.py
```diff
@@ -37,23 +40,19 @@ class DataDogStorageTier(str, Enum):
 DEFAULT_STORAGE_TIERS = [DataDogStorageTier.INDEXES]
 
 
-class DatadogConfig(BaseModel):
-    dd_api_key: str
-    dd_app_key: str
+class DatadogLogsConfig(DatadogBaseConfig):
     indexes: list[str] = ["*"]
```
moshemorad, line 44: I think you need to rebase it over the latest changes

holmes/plugins/toolsets/datadog/datadog_api.py
nherment: Will do that in the traces branch

moshemorad (approved):

PR #646, opened 2025-07-17
add github action for publishing to pypi + a script to do so manually

aantn:
the manual script was also run one-off to push the current version to pypi


.github/workflows/publish-pypi.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Update package version
+      run: |
+        sed -i 's/__version__ = .*/__version__ = "${{ github.ref_name }}"/g' holmes/__init__.py
+        sed -i 's/version = "0.0.0"/version = "${{ github.ref_name }}"/g' pyproject.toml
+
+    - name: Install dependencies
+      run: poetry install --no-root
+
+    - name: Publish to PyPI
+      run: poetry publish --build -u __token__ -p ${{ secrets.PYPI_TOKEN }}
```
aantn: I generated a token and added it as a secret for the repository

mainred (commented):
How about github action https://github.com/marketplace/actions/publish-python-poetry-package? 2.4k projects are using this github action to publish poetry packages.

aantn:
@mainred could we consider changing that in a later PR? I'm comfortable with this and did my best to test, so I'd like to merge before our next release. I'm happy to improve more later!

arikalon1 (approved):
nice work

PR #650, opened 2025-07-17
Try load robusta ai

moshemorad:
This PR added logic to load Robusta AI when the env var isn't provided. 
To support backward compatibility we check first if any mode / model list is define in env var and if not we try load robusta ai.

### DIFF
--- a/holmes/core/tool_calling_llm.py
+++ b/holmes/core/tool_calling_llm.py
@@ -347,15 +347,16 @@ class ToolCallingLLM:
                 )
 
             perf_timing.measure("pre-tool-calls")
-            with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
+            with concurrent.futures.ThreadPoolExecutor(max_workers=16) as executor:
                 futures = []
-                for t in tools_to_call:
+                for tool_index, t in enumerate(tools_to_call, 1):
                     logging.debug(f"Tool to call: {t}")
                     futures.append(
                         executor.submit(
                             self._invoke_tool,
                             tool_to_call=t,
                             previous_tool_calls=tool_calls,
+                            tool_number=tool_index,
                         )
                     )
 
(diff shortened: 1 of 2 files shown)


holmes/config.py
```diff
@@ -148,11 +149,31 @@ def model_post_init(self, __context: Any) -> None:
...
+        if ROBUSTA_AI is not None:
+            return ROBUSTA_AI
+
+        # MODEL is set in the env vars, e.g. the user is using a custom model
+        # so we don't need to load the robusta AI model and keep the behavior backward compatible
+        if "MODEL" in os.environ:
+            return False
+
+        # if the user has provided a model list, we don't need to load the robusta AI model
+        if self._model_list:
```
arikalon1, line 172: I think, in fact, we can load both model list and robusta AI
Both before and after this change, if the user provides both the ROBUSTA_AI env var, and the model list, both will be loaded.
I think it works ok, but the comment and the check seems redundant

holmes/config.py
```diff
@@ -220,6 +241,7 @@ def load_from_env(cls):
             if val is not None:
                 kwargs[field_name] = val
         kwargs["cluster_name"] = Config.__get_cluster_name()
+        # kwargs["should_try_robusta_ai"] = True
```
arikalon1: should we remove this?

arikalon1 (approved):
looks good, left a couple of minor comments

holmes/config.py
moshemorad: Yeah this will still work. The thing i didn't want to impact already exists customer. e.g. they don't have ROBUSTA_AI env var and i would like to check If i should load it or not. On old customers that doesn't working with ROBUSTA_AI it is either set to False or not exist. In the case it didn't exist i check if other env var exists which means they already configured their own model and if not only than i load ROBUSTA_AI

holmes/config.py
moshemorad: No, It is actually part of the solution i just plays with it (This is why it is still in draft)

arikalon1 (approved):
LGTM

Issue #566, opened 2025-06-25
Improve version check


aantn:
Right now the version check runs on every `holmes ask` invocation and has two issues:

1. it can be slow! to mitigate this, we cap the version check at 300ms in `fetch_holmes_info()` (checking online for the latest holmes version is what takes the time)
2. it isn't reliable - because we cap the version check at 300ms, sometimes fetching the latest version times out and then we skip the version check

To fix this, I suggest the following:

1. On startup, fetch the latest version from the sever **in a background thread** and save it to ~/.holmes/version 
2. On subsequent runs, if ~/.holmes/version already exists then do the actual comparison and print a message to the user if relevant. Then run the version check again in the background so we have an up to date ~/.holmes/version for next time.

PR #659, opened 2025-07-20
Do version check in background to avoid slowing down startup

aantn:
In the future we can take a similar approach with toolset reloading

### DIFF
--- a/holmes/version.py
+++ b/holmes/version.py
@@ -1,0 +1,179 @@
+"""
+Centralized version management for Holmes.
+Handles current version detection, latest version fetching, and comparison logic.
+"""
+
+import json
+import os
+import subprocess
+import sys
+import threading
+from typing import Optional, NamedTuple
+from functools import cache
+from cachetools import cached  # type: ignore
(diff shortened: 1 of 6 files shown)


aantn:
Fixes #566

holmes/version.py
```diff
@@ -0,0 +1,178 @@
...
+            current_version=current_version,
+            latest_version=holmes_info.latest_version,
+        )
+
+    # Check if current version starts with latest version
+    is_latest = current_version.startswith(holmes_info.latest_version)
+
+    update_message = None
+    if not is_latest:
+        update_message = f"Update available: v{holmes_info.latest_version} (current: {current_version})"
```
mainred, line 152: nit: should we remove the prefix v to let the latest and current version pattern match?

mainred (approved):
Left one nit, you can decide whether or not fix it in this or next PR, or ignore it.

holmes/version.py
aantn: I'm merging, but will also fix later

PR #663, opened 2025-07-21
Added the Prometheus alert investigation workflow

pavangudiwada:


docs/walkthrough/investigating-prometheus-alerts.md
```diff
@@ -0,0 +1,70 @@
...
+- HolmesGPT CLI installed ([installation guide](../installation/cli-installation.md))
+- An AI provider API key configured ([setup guide](../ai-providers/index.md))
+- Access to your AlertManager instance
+
+## Investigating a Prometheus Alert Using HolmesGPT
+
+
+### Step 1: Create a Test Alert
+
+Let's create a dummy alert in Prometheus for our investigation. Run the following command to create a`KubePodCrashLooping` alert:
```
aantn: This isn't quite right - this doesn't necessary create a Prometheus alert. It deploys the kubernetes workload which is crashing but whether there will be an alert or not depends on your alerts + it will take time. Can you deploy this workload + send an alert to alertmanager manually?

aantn (approved):
Looks great!

PR #665, opened 2025-07-21
Fixes a keyerror when URL parameter is missing in the internet toolset

### LINKED ISSUE #664: Holmes crashes with KeyError in Internet toolset when URL parameter is missing
### What happened?

Sometimes FetchWebpage tool in the internet toolset crashes with a KeyError when the get_parameterized_one_liner method is called with parameters that don't contain the expected "url" key. 

```
Analyzing issue 1/15: TargetDown...
No section received from the client. Default sections will be used.                                    
Structured output is disabled for this request                                                         
No runbooks found for this issue. Using default behaviour. (Add runbooks to guide the investigation.)
Tool call to fetch_webpage failed with an Exception                                                    
  File "/Users/pavan/Documents/repos/holmesgpt/holmes/core/tool_calling_llm.py", line 415, in          
_invoke_tool                                                                                           
    tool_response = tool.invoke(tool_params, tool_number=tool_number)                                  
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                  
  File "/Users/pavan/Documents/repos/holmesgpt/holmes/core/tools.py", line 147, in invoke              
    f"Running tool {tool_number_str}{self.name}: {self.get_parameterized_one_liner(params)}"           
                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pavan/Documents/repos/holmesgpt/holmes/plugins/toolsets/internet/internet.py", line 217,
in get_parameterized_one_liner                                                                         
    url: str = params["url"]                                                                           
               ~~~~~~^^^^^^^                                                                           
KeyError: 'url'                                                                                        
```

HolmesGPT to fetch a run book if it's available and continue the investigation. Not to crash, of course. 

Run a Holmes investigation with the `TargetDown` alert or any alert. This happened a few times, not sure of a specific way to reproduce. 

_No response_

pavangudiwada:
Possibly fixes #664


aantn (approved):

aantn:
I took a closer look and I'm not entirely sure what's happening here - it looks like the url is not being provided by the LLM despite it being required. We have an open task to implement strict mode for tool calling which will fix that, so I'm merging your fix for now. Long term it wont be necessary but lets get a fix immediately.

PR #666, opened 2025-07-21
Updated K9s plugin docs and fixed to work with contexts

pavangudiwada:
The new plugin 
1. Switches context automatically
2. Makes sure the default KubeConfig is not affected by the K9s context switching. 

CleanShot 2025-07-21 at 16 35 03


docs/installation/ui-installation.md
```diff
@@ -10,25 +10,104 @@ Integrate HolmesGPT into your [K9s](https://github.com/derailed/k9s){:target="_b
...
+        description: Ask HolmesGPT
+        scopes:
+          - all
+        command: bash
+        background: false
+        confirm: false
+        args:
+          - -c
+          - |
+            TEMP_KUBECONFIG=$(mktemp)
```
aantn: Can you add a comment explaining why we do this? (I know why we do it myself, but its not intuitive or self-explanatory so we should have a comment explaining to readers.)

docs/installation/ui-installation.md
```diff
@@ -10,25 +10,104 @@ Integrate HolmesGPT into your [K9s](https://github.com/derailed/k9s){:target="_b
...
+            cp ${KUBECONFIG:-~/.kube/config} $TEMP_KUBECONFIG
+            KUBECONFIG=$TEMP_KUBECONFIG kubectl config use-context $CONTEXT
```
aantn: KUBECONFIG here points at a directory not a file, no? How does this work?

docs/installation/ui-installation.md
aantn: Is it possible to check the current context and skip all this when its not necessary?

docs/installation/ui-installation.md
pavangudiwada: Yea you're right. It will fail if there are multiple files in the config. I'll switch to getting the config file from kubectl

docs/installation/ui-installation.md
pavangudiwada: I add a conditional statement.

docs/installation/ui-installation.md
aantn: I think your approach here was right, because it avoids switching the kubectl context in other terminal windows. It only impacts the curent window. I think I misunderstood something with regards to mktemp.

Please go back to the approach with TEMP_KUBECONFIG (but only use it when kubectx is actually different.) Its my bad, but your original approach was better.

pavangudiwada: Fixed and updated with some comments

aantn:
Looks good.

docs/installation/ui-installation.md
```diff
@@ -10,25 +10,126 @@ Integrate HolmesGPT into your [K9s](https://github.com/derailed/k9s){:target="_b
...
+        background: false
+        confirm: false
+        args:
+          - -c
+          - |
+            INSTRUCTIONS="# Edit the line below. Lines starting with '#' will be ignored."
+            DEFAULT_ASK_COMMAND="why is $NAME of $RESOURCE_NAME in -n $NAMESPACE not working as expected"
+            QUESTION_FILE=$(mktemp)
+
+            echo "$INSTRUCTIONS" > "$QUESTION_FILE"
```
aantn, line 94: @pavangudiwada can we just drop the user in interactive mode now?

pavangudiwada: After we release a new version, I think we can just run holmes ask and remove a plugin and some config. Right now we would have to change things too. A default question because Holmes won't have enough context about the app(I could be wrong). 
```
holmes ask QUESTION -i 
```

PR #672, opened 2025-07-22
chore: move utils functions out of main.go

mainred:
This PR moves the utils functions out of main.go to make utils function importal.


aantn:
@mainred can you explain? Why wasn't it importable before?

mainred:
Hey @aantn, what we want to achieve in az cli is like in the following PR
https://github.com/Azure/azure-cli-extensions/pull/8990
We want to re-implement the ask function, while keep at most utils function inside the holmes library as possible.

aantn:
> Hey @aantn, what we want to achieve in az cli is like in the following PR [Azure/azure-cli-extensions#8990](https://github.com/Azure/azure-cli-extensions/pull/8990) We want to re-implement the ask function, while keep at most utils function inside the holmes library as possible.

Got it, thanks! Approved.

PR #675, opened 2025-07-22
Fixes reset command and updates it for better UX

pavangudiwada:
When the context is reset, it's also nice to reset (clear) what's in the terminal to start fresh. (Claude code's /clear does this, however their default /clear is HolmesGPT's /reset + /clear)

Old behavior

https://github.com/user-attachments/assets/8d27e888-853a-4555-b82a-b33866d32836

New behavior

https://github.com/user-attachments/assets/69819834-998e-4cbb-b0e2-8b5eca0bd304

### DIFF
--- a/holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
+++ b/holmes/plugins/toolsets/coralogix/toolset_coralogix_logs.py
@@ -62,6 +62,9 @@ class CoralogixLogsToolset(BasePodLoggingToolset):
     def coralogix_config(self) -> Optional[CoralogixConfig]:
         return self.config
 
+    def logger_name(self) -> str:
+        return "Coralogix"
+
     def fetch_pod_logs(self, params: FetchPodLogsParams) -> StructuredToolResult:
         if not self.coralogix_config:
             return StructuredToolResult(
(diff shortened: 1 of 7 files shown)


holmes/interactive.py
```diff
@@ -833,7 +833,8 @@ def get_bottom_toolbar():
                     for cmd, description in SLASH_COMMANDS_REFERENCE.items():
                         console.print(f"  [bold]{cmd}[/bold] - {description}")
                     continue
-                elif command == SlashCommands.RESET.value:
+                elif command == SlashCommands.RESET.command:
+                    console.clear()
```
mainred, line 837: This PR looks good to fix the UX, but it seems our reset did not clear the previous llm context? cc @aantn

aantn: @mainred its cleared a few lines underneath inside the same elif. (Or did you experience a bug where it didn't happen?)

@pavangudiwada maybe print the message only after we actually reset the messages so the code is more clear.

mainred: Sorry, my bad. llm context should have been cleared when /clear is issued.

aantn: @mainred would you prefer we remove /reset and only have /clear which both clears the window and resets the context? Maybe its simpler that way

aantn: I'm implementing that in a new PR.

aantn:
Closing, as this is included in https://github.com/robusta-dev/holmesgpt/pull/680

PR #680, opened 2025-07-22
show correct tool numbers for /show and remove /reset

aantn:


holmes/core/tool_calling_llm.py
```diff
@@ -249,6 +249,7 @@ def call(  # type: ignore
         user_prompt: Optional[str] = None,
         sections: Optional[InputSectionsDataType] = None,
         trace_span=DummySpan(),
+        tool_number_offset: int = 0,
```
mainred, line 252: Thanks for the fix

holmes/core/tools.py
```diff
@@ -154,8 +154,9 @@ def invoke(
             if hasattr(result, "get_stringified_data")
             else str(result)
         )
+        show_hint = f"/show {tool_number}" if tool_number else "/show"
         logging.info(
-            f"  [dim]Finished {tool_number_str}in {elapsed:.2f}s, output length: {len(output_str):,} characters - /show to view contents[/dim]"
+            f"  [dim]Finished {tool_number_str}in {elapsed:.2f}s, output length: {len(output_str):,} characters - {show_hint} to view contents[/dim]"
```
mainred, line 159: `/show <number>` is good, but `/show <toolet_name>` is not, and I cannot get the tool result from the tool name command.
```
User: /show
Available tool outputs:
  1. kubectl get -A --show-labels -o wide pod
  2. kubectl top pods -A
  3. kubectl top nodes
  4. kubectl get -A --show-labels -o wide networkpolicy
Usage: /show <number> or /show <tool_name>
User: /show kubectl_top_nodes
Tool not found: kubectl_top_nodes
```

mainred: Is it enough to use index?

aantn: Can you try `/show kubectl get -A --show-labels -o wide networkpolicy` or `kubectl get`

aantn: index can work too but I don't think there is a downside to keeping the `/show <tool_name>` - maybe we can call it `/show <substring>`?

mainred: Sorry for the late reply.
I think the <tool_name> is a bit misleading since we usually call `kubectl_top_nodes` instead of  `tool command` as the name. and we seem to return the first tool result when it matches the prefix of the tool description when multiple tool matches.
What about /show <tool_command> not sure if it's better to give an example.

aantn: You're right, what do you think of `/show <substring>`

aantn:
Due to a mistake on my end (I pushed some unrelated commits by accident, closing and replacing with https://github.com/robusta-dev/holmesgpt/pull/696/files)

PR #683, opened 2025-07-22
fix: use dummy tracer

mainred:


holmes/interactive.py
```diff
@@ -722,9 +721,9 @@ def run_interactive_loop(
     show_tool_output: bool,
     tracer=None,
```
aantn, line 722: Feel free to also make this parameter non-optional. The tracer is always being picked and passed by the caller in main.py anyway

aantn (approved):

PR #698, opened 2025-07-23
ROB-1797 fix bug of losing temp arg

RoiGlinik:
fixes a bug where temperature is poped and then lost for 2+ complete iterations.

seen on bedrock with thinking arguments .


arikalon1 (approved):
nice work

PR #701, opened 2025-07-24
ROB-1778 support openshift token for prometheus toolset

RoiGlinik:


holmes/common/openshift.py
```diff
@@ -0,0 +1,12 @@
+from typing import Optional
+
+# NOTE: This one will be mounted if openshift is enabled in values.yaml
+TOKEN_LOCATION = "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
Avi-Robusta: maybe set this as an env var so it can be configured incase it is in a different location on different cluster types?

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -55,6 +57,20 @@ def ensure_trailing_slash(cls, v: Optional[str]) -> Optional[str]:
             return v + "/"
         return v
 
+    @model_validator(mode="after")
+    def validate_openshift_token(self):
+        # If openshift is enabled, and the user didn't configure auth headers, we will try to load the token from the service account.
+        if IS_OPENSHIFT:
+            if not self.headers.get("Authorization"):
```
Avi-Robusta, line 64:
```suggestion
            if self.headers.get("Authorization"):
              return self
            openshift_token = load_openshift_token()...
```

Avi-Robusta: just cleaner to understand

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -55,6 +57,20 @@ def ensure_trailing_slash(cls, v: Optional[str]) -> Optional[str]:
             return v + "/"
         return v
 
+    @model_validator(mode="after")
+    def validate_openshift_token(self):
+        # If openshift is enabled, and the user didn't configure auth headers, we will try to load the token from the service account.
+        if IS_OPENSHIFT:
```
Avi-Robusta, line 63: the nested ifs make this a bit hard to read,
it would be easier to read for something like 
```
        if not IS_OPENSHIFT:
          return self
          
         if self.healthcheck == "-/healthy":
             self.healthcheck = "api/v1/query?query=up"
             
         if self.headers.get("Authorization"):
           return self
           
          openshift_token = load_openshift_token()
          if openshift_token:
              logging.info("Using openshift token for prometheus toolset auth")
              self.headers["Authorization"] = f"Bearer {openshift_token}"
           return self
```

wdyt?

helm/holmes/templates/holmes.yaml
```diff
@@ -56,6 +56,10 @@ spec:
           - name: HOLMES_POST_PROCESSING_PROMPT
             value: {{ .Values.postProcessingPrompt }}
           {{- end }}
+          {{ if .Values.openshift }}
```
Avi-Robusta, line 59: add to the values file
```
openshift: False # For prometheus Openshift support
```

Avi-Robusta (approved):

helm/holmes/templates/holmes.yaml
RoiGlinik: value exists, this is just new

holmes/plugins/toolsets/prometheus/prometheus.py
RoiGlinik: OK changed to almost like that. Because this should be a generic check  I left the openshfit check as is becuase we might have other configs here

Avi-Robusta (approved):

PR #727, opened 2025-07-27
Share more logic for system prompt templating

aantn:


holmes/main.py
```diff
@@ -295,20 +285,21 @@ def ask(
         run_interactive_loop(
             ai,
             console,
-            system_prompt_rendered,
             prompt,
             include_file,
             post_processing_prompt,
             show_tool_output,
             tracer,
+            config.get_runbook_catalog(),
```
mainred, line 295: Hiding the implementation of system prompt make user impossible to extend the system prompt.
How about adding a additional variable like extended-system-prompt to allow the user to extend the system prompt to add their context?

mainred (approved):

PR #729, opened 2025-07-27
Improve HolmesGPT accuracy on questions about itself and configuring tools

aantn:


mainred (approved):
Thanks for the improvment.

holmes/plugins/toolsets/runbook/runbook_fetcher.py
```diff
@@ -33,15 +36,75 @@ def __init__(self, toolset: "RunbookToolset"):
...
+
+        # Add default runbook folder
+        default_folder = Path(__file__).parent.parent.parent / "runbooks"
+        search_paths.append(str(default_folder))
+
+        # Find the runbook
+        runbook_path = get_runbook_by_path(link, search_paths)
 
-        runbook_path = get_runbook_by_path(path)
+        if runbook_path is None:
```
Sheeproid, line 55: you've changed how holmes load the runbooks so it works in the tests with the default folder. Please verify the regular functionality is unharmed. If you've already done so, please resolve the comment

holmes/plugins/toolsets/runbook/runbook_fetcher.py
```diff
@@ -33,15 +36,75 @@ def __init__(self, toolset: "RunbookToolset"):
...
+                status=ToolResultStatus.ERROR,
+                error=err_msg,
+                params=params,
+            )
+
+        # Read and return the runbook content
         try:
             with open(runbook_path, "r") as file:
                 content = file.read()
+                wrapped_content = textwrap.dedent(f"""\
```
Sheeproid, line 70: do you do the textwarp with the dedent and the inded because the use of triple-quotes adds spaces? is this a common thing to do? It looks error prone becuase the use of the constant '20' relies on the structure of the python function

holmes/plugins/runbooks/__init__.py
```diff
@@ -94,7 +94,27 @@ def load_runbook_catalog() -> Optional[RunbookCatalog]:
     return None
 
 
-def get_runbook_by_path(runbook_relative_path: str) -> str:
-    runbook_folder = os.path.dirname(os.path.realpath(__file__))
-    runbook_path = os.path.join(runbook_folder, runbook_relative_path)
-    return runbook_path
+def get_runbook_by_path(
```
Sheeproid, line 97: see comment about verification in runbooks_fetcher

holmes/plugins/prompts/generic_ask.jinja2
```diff
@@ -3,6 +3,7 @@ Whenever possible you MUST first use tools to investigate then answer the questi
 Do not say 'based on the tool output' or explicitly refer to tools at all.
 If you output an answer and then realize you need to call more tools or there are possible next steps, you may do so by calling tools at that point in time.
 If you have a good and concrete suggestion for how the user can fix something, tell them even if not asked explicitly
+{% include '_current_date_time.jinja2' %}
 
 Use conversation history to maintain continuity when appropriate, ensuring efficiency in your responses.
```
Sheeproid, line 10: @aantn we only have the '_runbook_instructions' in the generic_ask and not in server (conversation?) ask or even the RCA. Note sure if this should be fixed as a part of the PR but maybe open a ticket.

holmes/plugins/prompts/_runbook_instructions.jinja2
```diff
@@ -1,13 +1,21 @@
 {% if runbooks and runbooks.catalog|length > 0 %}
 # Runbook Selection
 
-## Available Runbooks
+You (HolmesGPT) have access to a set of runbooks that provide step-by-step troubleshooting instructions for various known issues.
```
Sheeproid, line 4: It's the first time I see HolmesGPT in a prompt. It is always 'You are a tool-calling AI assist'. 
I think that either:
1. Change "You (HolmesGPT) have access to".. to "You have access to..."
2. Replace "You are a tool-calling AI..." with "You are HolmesGPT, a tool-calling AI..." across the board.

holmes/plugins/prompts/_runbook_instructions.jinja2
```diff
@@ -1,13 +1,21 @@
...
+If one of the following runbooks relates to the user's issue, you MUST fetch it with the fetch_runbook tool.
```
Sheeproid, line 5: are both needed, or can we only keep one of them? (probably not critical to change and test, just seems redundant)
1. If one of the following runbooks relates to the user's issue, you MUST fetch it with the fetch_runbook tool.
2. If there is a runbook that MIGHT match the user's issue, you MUST:

holmes/plugins/prompts/_general_instructions.jinja2
```diff
@@ -9,7 +9,8 @@
 * in this case, try to find substrings or search for the correct spellings
 * always provide detailed information like exact resource names, versions, labels, etc
 * even if you found the root cause, keep investigating to find other possible root causes and to gather data for the answer like exact names
-* if a runbook url is present as well as tool that can fetch it, you MUST fetch the runbook before beginning your investigation.
+* if a runbook url is present you MUST fetch the runbook before beginning your investigation
```
Sheeproid, line 12: I'm a bit confused, we have both this instructions and _runbook_instructions, and they behave differently. see my other comment below

holmes/plugins/prompts/_permission_errors.jinja2
```diff
@@ -0,0 +1,6 @@
+# Handling Permission Errors
```
Sheeproid, line 1: it looks shorter now, perhaps we can put directly in general_instructions?

holmes/plugins/prompts/generic_ask.jinja2
aantn: Right, that's correct. Will fix separately.

holmes/plugins/prompts/_runbook_instructions.jinja2
aantn: Not fixing now, but makes sense

holmes/plugins/prompts/_general_instructions.jinja2
aantn: I thought of leaving it in here in case there is a runbook url referenced in the alert (not a holmes runbook, just plain url) but no builtin runbooks. In that case the _runbook_instructions are never included in the template. We could possibly handle better, but resolving for now.

holmes/plugins/prompts/_runbook_instructions.jinja2
aantn: yeah, probably redundant. will experiment later with simplifying, but keeping for now as its what I tested.

aantn:
Thanks @Sheeproid - good review. As discussed, I will merge as is and open a follow up PR improving further.

PR #734, opened 2025-07-28
ask for multiple tool calls for reasoning models

Sheeproid:
claude sonnet 4 tends to call tools one by one, generating a lot of LLM calls and slowing down the process for the user


aantn (approved):
great work

PR #743, opened 2025-07-29
WIP: Advanced logging capabilities

aantn:
# Overview
This fixes several known failures where holmes was not seeing appropriate logs due to incorrect filtering / data overflow.

# Concept
The fix includes 4 main parts:
1. Allowing holmes to use regex filters (where supported) and encouraging it to do so
2. Allowing holmes to use an exclude filter and encouraging it to do so (technically this is doable with negative lookahead regexes, but having separate include and exclude filters gave the best results)
3. When holmes queries logs, including hints at the end of the results on how to use filters in a follow up query (if appropriate)
4. Adding a 'capabilities' mechanism to the logging framework so that we can incrementally add new capabilities like this (i.e. regex support) on a per-log-provider basis without needing to support it all at once for every provider.

# Example
Here is a logging query before this fix, where holmes fails to find logs related to a login issue:
Screenshot 2025-07-29 at 11 32 55

Here is the same eval after this fix:
Screenshot 2025-07-29 at 11 32 40

# Testing
 
I tested with appropriate evals and verified the fix works. Holmes now succeeds at evals where it previously failed. Two of these are now 100% reliable and were promoted to the 'easy' category to function as regression tests going forward.

I also ran all existing 'easy' evals 10 times to verify there is no off-target regression.


holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -14,12 +15,20 @@
 from holmes.plugins.toolsets.utils import get_param_or_raise
 
 # Default values for log fetching
-DEFAULT_LOG_LIMIT = 2000
-DEFAULT_TIME_SPAN_SECONDS = 3600
+DEFAULT_LOG_LIMIT = 100
+SECONDS_PER_DAY = 24 * 60 * 60
+DEFAULT_TIME_SPAN_SECONDS = 7 * SECONDS_PER_DAY  # 1 week in seconds
```
moshemorad, line 20: Consider make it configurable

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -14,12 +15,20 @@
 from holmes.plugins.toolsets.utils import get_param_or_raise
 
 # Default values for log fetching
-DEFAULT_LOG_LIMIT = 2000
-DEFAULT_TIME_SPAN_SECONDS = 3600
+DEFAULT_LOG_LIMIT = 100
```
moshemorad, line 18: Are we sure that this is good enough for logging systems which aren't k8s and doesn't support the advances logging capabilities?

moshemorad (approved):

holmes/plugins/toolsets/logging_utils/logging_api.py
aantn: Yes, I'm comfortable with it. The alternative is filling up the context window.
In evals, I've seen that stronger models like Sonnet will issue follow up queries if the initial logging data doesn't contain what they want. So better to limit and let the LLM requery when necessary.

PR #747, opened 2025-07-29
bump azure-mgmt-sql to the latest 4.0.0b21

mainred:


pyproject.toml
```diff
@@ -31,7 +31,7 @@ markdown = "^3.6"
 certifi = "^2024.7.4"
 urllib3 = "^1.26.19"
 boto3 = "^1.34.145"
-setuptools = "^80.9.0"
+setuptools = "70.0.0"
```
mainred, line 34: Ping setuptools to 70.0.0 since there's a hard requirement of 70.0.0 for azure-cli to build python wheels, and otherwise the wheel build will break for the following issue as a result of incompatibility of setuptools and wheels.
```
    from wheel.wheelfile import WheelFile
ModuleNotFoundError: No module named 'wheel.wheelfile'
```
https://dev.azure.com/azclitools/public/_build/results?buildId=263962&view=logs&j=2e66ce57-bcb2-51b3-abdd-698c078f35ff&t=16b5c26c-808a-5e55-7cf9-03cc11587706&l=3573
https://github.com/Azure/azure-cli/issues/29467

mainred: setuptools are used in the our check-in. As long as the check-in are green, I think it's good to pin the setuptools versions.
Please bring your concerns if we have.

pyproject.toml
```diff
@@ -52,12 +52,12 @@ pytest-shared-session-scope = "^0.4.0"
...
 requests = "^2.32.4"
-azure-mgmt-sql = "^3.0.1"
+azure-mgmt-sql = "^4.0.0b21"
 pyodbc = "^5.0.1"
 azure-monitor-query = "^1.2.0"
-azure-mgmt-monitor = "^6.0.2"
+azure-mgmt-monitor = "^7.0.0b1"
 azure-mgmt-alertsmanagement = "^1.0.0"
-azure-mgmt-resource = "^24.0.0"
+azure-mgmt-resource = "^23.3.0"
```
mainred, line 60: This is not a bump, but downgrade to match the version used in azure-cli.
I know downgrading a package without a specific reason is not good, but I think it's ok since azure-cli officially use 23.3.0 as well.

arikalon1 (approved):
Nice work

arikalon1 (approved):

PR #750, opened 2025-07-29
make config file dir configurable

mainred:
export CONFIG_PATH_DIR=/home/azureuser/testholmes
running holmes aks gives the following files correctly
```
ls ~/testholmes 
config.yaml  history  toolsets_status.json
```

### DIFF
--- a/holmes/core/config.py
+++ b/holmes/core/config.py
@@ -1,0 +1,5 @@
+import os
+
+config_path_dir: str = os.environ.get(
+    "CONFIG_PATH_DIR", os.path.expanduser("~/.holmes")
+)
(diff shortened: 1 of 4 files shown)


holmes/core/config.py
```diff
@@ -0,0 +1,5 @@
+import os
+
+config_path_dir: str = os.environ.get(
+    "CONFIG_PATH_DIR", os.path.expanduser("~/.holmes")
```
aantn: Can you make it HOLMES_CONFIGPATH_DIR?

The current value is,too generic and could conflict with some other cli tool

aantn (approved):

PR #757, opened 2025-07-29
add line count to tool output

aantn:


pavangudiwada (approved):
Works as expected 
CleanShot 2025-07-30 at 08 06 00

PR #759, opened 2025-07-30
ROB-1689 call_stream support for api chat

RoiGlinik:
improve and simplify stream_call function
make stream message more flexible for other return formats.


moshemorad (approved):
LGTM

PR #765, opened 2025-07-31
[ROB-1799] fixed output formatting issues - better sonnet-4 support

Avi-Robusta:
We would recieve a non jsoned output with a json in it like so
```
Now I have all the information needed to provide the analysis. The issue is clear from the evidence.

```json
{
  "Alert Explanation": "The pod `analytics-exporter-598c8b7d5-gjn4p` has been stuck in Pending state for over 15 minutes due to node affinity mismatch - it requires amd64 architecture but all cluster nodes are arm64.",
...
  "Related Logs": null,
  "External Links": "[Kubernetes Node Selection](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/) - Official documentation on node selectors and affinity rules."
}
`'` 

```

it wouldnt parse the results properly, now it does


holmes/core/investigation_structured_output.py
```diff
@@ -177,6 +177,18 @@ def pre_format_sections(response: Any) -> Any:
         # In that case it gets parsed once to get rid of the first level of marshalling
         with suppress(Exception):
             response = json.loads(response)
+
+    # Try to find any embedded code block with or without "json" label and parse it
```
moshemorad, line 181: please add test for it in the relevant test function

moshemorad (dismissed):

PR #766, opened 2025-07-31
make agent name in welcome banner configurable

mainred:
This change was included in https://github.com/robusta-dev/holmesgpt/pull/730/files, but was reverted after resolving a conflict.


arikalon1 (approved):
nice work

PR #775, opened 2025-08-03
Rob 1873 validate datadog logs query working as expected

moshemorad:


tests/llm/fixtures/test_ask_holmes/91_calling_datadog/test_case.yaml
```diff
@@ -0,0 +1,15 @@
+description: |
+  A test case that simulates an issue real customer bumped into.
+  When holmes were asked to fetch logs, they were not able to do so.
+  The main reason behind this are do to the conversation history and the lack of explicity saying that we enabled datadog logs toolset as part of the system prompt logging section.
```
aantn, line 4: Typo in do (should be due)

aantn (approved):
Looks good. What's the behaviour if environment variables referenced in toolsets.yaml are missing in the test environment? I.e. if I try to run these evals and I don't have the Datadog environment variables defined?

It would be nice if the test was marked as skipped then. (pytest.skip)

Please merge regardless. If the above doesn't work yet, we can improve in the future.

PR #780, opened 2025-08-04
ROB-1873-validate-datadog-logs-query-working-as-expected

moshemorad:


tests/llm/utils/mock_toolset.py
```diff
@@ -413,42 +404,46 @@ def __init__(
...
 
-        elif self._mode == MockMode.GENERATE:
-            # Generate mode: call real tool and save result
-            logging.info(f"Generating mock for tool {self.name} with params: {params}")
-            result = self._tool.invoke(params)
+    def _invoke(self, params: Dict) -> StructuredToolResult:
+        """Execute the tool based on the current mode."""
+        if self._mode == MockMode.GENERATE:
+            try:
+                return self._call_mock_invoke(params)
```
aantn, line 442: I'd expect to always generate the mock with --generate and not to reuse from the cache. Is there a good reason to use the cached file when --generate was passed? (The user explicitly wants to generate.)

holmes/core/tools_utils/toolset_utils.py
```diff
@@ -16,12 +16,17 @@ def filter_out_default_logging_toolset(toolsets: list[Toolset]) -> list[Toolset]
     All other types of toolsets are included as is.
     """
 
-    logging_toolsets: list[BasePodLoggingToolset] = []
+    logging_toolsets: list[Toolset] = []
```
aantn, line 19: Just curious - what's the motivation for this change?

aantn (approved):

holmes/core/tools_utils/toolset_utils.py
moshemorad: @aantn Since when we are working on mock only we are wrapping the toolsets with `SimplifiedMockToolset` which isn't actually BasePodLoggingToolset. I saw that we are printing that logging toolset wasn't enabled and it was in mock mode due to the reason above so i just make sure it will work as well.

tests/llm/utils/mock_toolset.py
moshemorad: Good point im not sure about the generate and regenerate flow. Let me think about it littlebit more and get back to you.

holmes/core/tools_utils/toolset_utils.py
aantn: Thanks, makes sense.

PR #797, opened 2025-08-06
return errors on empty queries. Prompy update

arikalon1:


holmes/plugins/toolsets/datadog/toolset_datadog_metrics.py
```diff
@@ -114,6 +114,12 @@ def _invoke(self, params: Any) -> StructuredToolResult:
             )
 
             metrics = data.get("metrics", [])
+            if not metrics:
+                return StructuredToolResult(
+                    status=ToolResultStatus.ERROR,
```
moshemorad, line 119: Do we sure we want to return here error and no_data?

moshemorad (approved):

holmes/plugins/toolsets/datadog/toolset_datadog_metrics.py
arikalon1: I think so. This is not an error returned from server, just no results
Until now, this would happen when you use the wrong filter tags.
The response would be success, but without metrics.
Now we fail it, and instruct the llm to try in a different way

PR #802, opened 2025-08-07
feat: adding option for custom annotations

DMSwimTech:
Hey, 
Can we add option for custom annotations? This is something that can be used in environments where istio or akeyless is used and annotations are required


helm/holmes/values.yaml
```diff
@@ -5,6 +5,7 @@ logLevel: INFO
 additionalEnvVars: []
 additional_env_vars: []
 imagePullSecrets: []
+customDeploymentAnnotations: {}
```
arikalon1: Can we rename it to `annotations` ?
This would be more inline with our values naming convention

arikalon1 (approved):
thanks for the PR @DMSwimTech 

nice work

arikalon1 (approved):

PR #806, opened 2025-08-10
Faster cli startup - refresh toolsets in background

aantn:
TODO:
* what happens when enabled toolsets changes in the refresh? do we need to regenerate the system prompt? update something in toolcallingllm? (It's also OK by me if we just tell the user to restart to apply changes, if easier)

### DIFF
--- a/holmes/config.py
+++ b/holmes/config.py
@@ -254,21 +254,19 @@ class Config(RobustaBaseConfig):
         )
         return ToolExecutor(cli_toolsets)
 
-    def create_agui_tool_executor(self, dal: Optional["SupabaseDal"]) -> ToolExecutor:
+    def create_agui_tool_executor(
+        self, dal: Optional["SupabaseDal"] = None
+    ) -> ToolExecutor:
         """
         Creates ToolExecutor for the AG-UI server endpoints
         """
-
         if self._agui_tool_executor:
             return self._agui_tool_executor
 
         # Use same toolset as CLI for AG-UI front-end.
-        agui_toolsets = self.toolset_manager.list_console_toolsets(
-            dal=dal, refresh_status=True
-        )
-
+        agui_manager = ToolsetManager.for_cli(self, dal=dal)
+        agui_toolsets = agui_manager.load(use_cache=False)
         self._agui_tool_executor = ToolExecutor(agui_toolsets)
-
         return self._agui_tool_executor
 
     def create_tool_executor(self, dal: Optional["SupabaseDal"]) -> ToolExecutor:
--- a/holmes/core/toolset_manager.py
+++ b/holmes/core/toolset_manager.py
@@ -570,13 +570,22 @@ class ToolsetManager:
         )
 
     @classmethod
-    def for_cli(cls, config: Optional["Config"]):
+    def for_cli(cls, config: Optional["Config"], dal: Optional[SupabaseDal] = None):
         """Create CLI-configured manager from Config object"""
-        return cls._create_from_config(config, CLI_TOOL_TAGS, default_enabled=True)
+        return cls._create_from_config(
+            config, CLI_TOOL_TAGS, dal=dal, default_enabled=True
+        )
 
     @classmethod
     def for_server(cls, config: "Config", dal: SupabaseDal):
         """Create server-configured manager from Config object"""
+        return cls._create_from_config(
+            config, SERVER_TOOL_TAGS, dal=dal, default_enabled=False
+        )
+
+    @classmethod
+    def for_agui(cls, config: "Config", dal: SupabaseDal):
+        """Create AG-UI-configured manager from Config object"""
         return cls._create_from_config(
             config, SERVER_TOOL_TAGS, dal=dal, default_enabled=False
         )


holmes/main.py
```diff
@@ -248,10 +249,18 @@ def ask(
     tracer = TracingFactory.create_tracer(trace, project="HolmesGPT-CLI")
     tracer.start_experiment()
 
+    # Check if toolsets will be loaded from cache
+    from holmes.core.toolset_manager import cache_exists
+
+    # In non-interactive mode, always refresh toolsets for fresh results
+    # In interactive mode, use cache unless explicitly told to refresh
+    should_refresh = refresh_toolsets or not interactive
```
mainred, line 266: For non-interactive mode, the tooset loading will also be printed out, that''s not user friendly. Some of the tools I don't even ever enabled by meeting their prerequisites, still we try to load during start up.

mainred: I think what we have before this PR is pretty good.
What we cache is only the toolsets that user once enabled, and the next the user run cli, it only tries to load these enabled toolsets. And we may not able to ignore the cached enabled toolsets, since these cached enabled toolsets may not always available. For example, some of the toolsets relies on environment like confluence, we set the env in current session, and the confluence toolset is enabled, but the next time I start a new session, it will not be avaialbe since env is not set. Same case for aks/core, aks/core relies on az login, the credential established from which will expire some time later.

holmes/config.py
```diff
@@ -346,7 +354,8 @@ def create_console_issue_investigator(
         from holmes.core.runbooks import RunbookManager
 
         runbook_manager = RunbookManager(all_runbooks)
-        tool_executor = self.create_console_tool_executor(dal=dal)
+        # Always refresh toolsets for investigations (non-interactive)
+        tool_executor = self.create_console_tool_executor(refresh_status=True)
```
arikalon1: Why did we change the default of refresh toolsets for the non-interactive?
I'm mostly not using interactive, and would like to cache toolsets.

holmes/config.py
```diff
@@ -276,7 +276,13 @@ def create_tool_executor(self, dal: Optional["SupabaseDal"]) -> ToolExecutor:
         if self._server_tool_executor:
             return self._server_tool_executor
 
-        toolsets = self.toolset_manager.list_server_toolsets(dal=dal)
+        # Create server-specific manager
+        if dal is None:
```
arikalon1, line 281: do we want to somehow allow running the server without the saas?

holmes/core/toolset_manager.py
```diff
@@ -15,430 +21,447 @@
...
+                # Override existing toolset
+                override_toolsets = load_toolsets_from_config(
+                    {name: toolset_config}, strict_check=False
+                )
+                if not override_toolsets:
+                    logging.warning(
+                        f"Skipping invalid override config for toolset '{name}'"
+                    )
+                    continue
+                self.toolsets[name].override_with(override_toolsets[0])
```
arikalon1: What is the reason to use `override_with` on existing?
Why not just put `self.toolsets[name] = new_toolsets[0]` on both cases?
Would probably make this code a little simpler

holmes/core/toolset_manager.py
```diff
@@ -15,430 +21,447 @@
...
+CLI_TOOL_TAGS = [ToolsetTag.CORE, ToolsetTag.CLI]
+SERVER_TOOL_TAGS = [ToolsetTag.CORE, ToolsetTag.CLUSTER]
+
+
+def cache_exists() -> bool:
+    """Check if the toolset cache file exists."""
+    return os.path.exists(DEFAULT_TOOLSET_STATUS_LOCATION)
+
+
+class ToolsetCache:
```
arikalon1: My personal preference, is to put the class in it's own file.
Makes it's easier to review in smaller files

holmes/core/toolset_manager.py
```diff
@@ -15,430 +21,447 @@
...
+                # Create new toolset
+                new_toolsets = load_toolsets_from_config(
+                    {name: toolset_config}, strict_check=True
+                )
+                if not new_toolsets:
+                    logging.warning(f"Skipping invalid config for new toolset '{name}'")
+                    continue
+                self.toolsets[name] = new_toolsets[0]
+
+    def load_from_yaml_file(self, file_path: Path) -> set:
```
arikalon1: This is actually called only for custom toolsets. Maybe we should name it accordingly 
Regardless, I think we're over complicating it.
We allow overriding builtin toolset with a custom one - is this really required?
We can remove it - and if a user wants to override, he can disable the builtin, and add a custom one with a different name.
We also support custom yaml toolset with mcp_servers - isn't that just config?

holmes/core/toolset_manager.py
```diff
@@ -15,430 +21,447 @@
...
+            return {}
+
+    def read_toolsets(self) -> Dict[str, Dict]:
+        """Read just the toolset data from cache"""
+        data = self._read()
+        return data.get("toolsets", {})
+
+    def write(self, toolsets: List[Toolset], content_hash: str):
+        """Write cache with content hash"""
+        cache_data = {
```
arikalon1: How does this work?
What are we caching here? Is it the whole toolset, like the name suggests, or only the toolset status?
If it's the whole toolset, why aren't we caching other important fields, like the config, llm_instructions etc?

holmes/config.py
aantn: @mainred any objection on your side? I recall you bringing this up once.

holmes/config.py
aantn: I'm inclined to allow it, but for now I just reserved the existing behaviour. Lets merge this without changing it and we can change later?

holmes/core/toolset_manager.py
aantn: Will clarify by changing name to ToolsetStatusCache

holmes/interactive_mode.py
```diff
@@ -1098,8 +1069,19 @@ def get_bottom_toolbar():
...
+
     while True:
         try:
+            # Start background refresh after first prompt is shown
+            # Only if we didn't just refresh (either via flag or because cache was missing)
+            if not background_refresh_started and not refresh_toolsets:
+                background_refresh_started = True
+                if toolset_refresh_manager.should_refresh():
+                    # Start background refresh immediately
+                    start_background_toolset_refresh()
```
mainred, line 1039: The toolset status cache is written by toolset_refresh_manager in background, and read by tool executor, how do we make sure the tool executor read the expected toolset? As we build system prompt in tool executor based on the toolset status and the system prompt will be used across the whole interactive session, If the toolset status in the system prompt is not the latest, the tools might not be called expectedly.

holmes/core/toolset_manager.py
```diff
@@ -1,460 +1,405 @@
...
         )
-        loaded_custom_toolsets.extend(custom_toolsets)
 
-        return loaded_custom_toolsets
+        # Either use cache or check prerequisites
+        if use_cache and not self.status_cache.is_stale_for_content(content_hash):
+            cached_data = self.status_cache.read_toolsets()
+            checked = self._apply_cached_data(relevant, cached_data)
+        else:
+            checked = self.checker.check_all(
```
mainred, line 363: For CLI, even we use cache, since the cache content stales after 3600s(1h), all toolset will be checked?

mainred: And prints these logs
```
❌ Toolset argocd/core: `argocd version` returned 127
❌ Toolset aws/rds: `aws sts get-caller-identity` returned 127
❌ Toolset aws/security: `aws sts get-caller-identity` returned 127
✅ Toolset kubernetes/kube-prometheus-stack
✅ Toolset internet
❌ Toolset robusta: The data access layer is not available
❌ Toolset opensearch/status: The toolset is missing its configuration
❌ Toolset grafana/loki: Missing Loki configuration. Check your config.
❌ Toolset datadog/logs: The toolset is missing its configuration
❌ Toolset newrelic: No configuration provided
❌ Toolset confluence: Environment variable CONFLUENCE_BASE_URL was not set
❌ Toolset slab: Environment variable SLAB_API_KEY was not set
❌ Toolset coralogix/logs: The toolset is missing its configuration
❌ Toolset git: Missing one or more required Git configuration values.
❌ Toolset MongoDBAtlas: Missing config credentials.
❌ Toolset grafana/tempo: The toolset is missing its configuration
❌ Toolset datadog/metrics: The toolset is missing its configuration
❌ Toolset datadog/traces: No configuration provided for Datadog Traces toolset
✅ Toolset datadog/rds
❌ Toolset opensearch/logs: Missing OpenSearch configuration. Check your config.
❌ Toolset opensearch/traces: Missing opensearch traces URL. Check your config
❌ Toolset grafana/grafana: The toolset is missing its configuration
❌ Toolset rabbitmq/core: RabbitMQ toolset is misconfigured. 'management_url' is required.
❌ Toolset notion: Notion toolset is misconfigured. Authorization header is required.
✅ Toolset bash
❌ Toolset kafka/admin: The toolset is missing its configuration
✅ Toolset runbook
❌ Toolset azure/sql: The toolset is missing its configuration
❌ Toolset ServiceNow: Missing config credentials.
✅ Toolset kubernetes/logs
✅ Toolset docker/core
✅ Toolset json_processor
✅ Toolset kubernetes/core
discovered service with label-selector: `app=kube-prometheus-stack-prometheus` at url:
`http://prometheus-kube-prometheus-prometheus.default.svc.cluster.local:9090`
Prometheus auto discovered at url http://prometheus-kube-prometheus-prometheus.default.svc.cluster.local:9090
❌ Toolset helm/core: `helm version` returned 1
❌ Toolset prometheus/metrics: Failed to initialize using url=http://prometheus-kube-prometheus-prometheus.default.svc.cluster.local:9090/-/healthy
❌ Toolset kubernetes/krew-extras: `kubectl version --client && kubectl lineage --version` returned 1
✅ Toolset kubernetes/live-metrics
✅ Toolset aks/core
✅ Toolset aks/node-health
```

docs/data-sources/configuring-toolsets.md
```diff
@@ -0,0 +1,274 @@
...
+toolsets:
+  <toolset-name>:
+    description: string        # Required: Human-readable description
+    enabled: boolean          # Optional: Default false
+    tools: list              # Required: List of tools
+    config: object          # Optional: Configuration values
+    prerequisites: list     # Optional: Prerequisite checks
+    tags: list             # Optional: Categorization tags
+    additional_instructions: string  # Optional: Extra LLM instructions
+```
```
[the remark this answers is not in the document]
moshemorad: why? toolset class indeed support tags. no?

PR #807, opened 2025-08-10
Speed up cli startup - lazy imports

aantn:


holmes/plugins/toolsets/lazy_imports.py
```diff
@@ -0,0 +1,78 @@
...
+    spec.loader = loader
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[name] = module
+    loader.exec_module(module)
+
+    return module
+
+
+# Pre-configured lazy imports for common expensive modules
+def get_azure_identity():
```
mainred, line 39: How we detect these expensive loading modules?

mainred (commented):
Will we continue working on this speed up?

holmes/plugins/toolsets/lazy_imports.py
aantn: I added a bunch of prints and tested the hard way :/

mainred (approved):

PR #809, opened 2025-08-10
add support for REASONING_EFFORT environment variable

aantn:


docs/ai-providers/openai.md
```diff
@@ -27,13 +27,43 @@ holmes ask "what pods are failing?" --api-key="your-api-key"
...
+
+When using GPT-5 models, you can control the reasoning effort level by setting the `REASONING_EFFORT` environment variable. This allows you to balance between response quality and processing time/cost.
+
+```bash
+# Use minimal reasoning effort for faster responses
+export REASONING_EFFORT="minimal"
+holmes ask "what pods are failing?" --model="gpt-5"
+
+# Use default reasoning effort
+export REASONING_EFFORT="default"
```
moshemorad: Are we sure the default is a valid value? https://github.com/BerriAI/litellm/pull/13401/files#diff-ff7eaee5a729fba18ac829136f92fdbadbb3b27243818ff8b63aef768b4a05b6

is it supported by gpt5?

holmes/core/llm.py
```diff
@@ -217,6 +220,10 @@ def completion(
         if self.args.get("thinking", None):
             litellm.modify_params = True
 
+        if REASONING_EFFORT:
```
moshemorad, line 223: Consider add comment that when upgrading litellm we can remove that.

moshemorad (dismissed):

holmes/core/llm.py
aantn: thanks, it automerged after your approval so I'll fix separately

docs/ai-providers/openai.md
aantn: good catch, thanks

moshemorad (approved):

PR #814, opened 2025-08-11
replace pwd with os compatible getpass

mainred:
1. pwd module (Unix/Linux only) in holmes/core/tracing.py

import pwd
"username": pwd.getpwuid(os.getuid()).pw_name,

2. signal in tests/llm/utils/port_forward.py
```
import signal
os.killpg(os.getpgid(self.process.pid), signal.SIGTERM)
os.killpg(os.getpgid(self.process.pid), signal.SIGKILL)
```
We can ignore this as it's test

3. hard-coded unix paths in holmes/common/env_vars.py

"ROBUSTA_CONFIG_PATH", "/etc/robusta/config/active_playbooks.yaml"

Currently it's used in server side, and can be fixed by overwriting the path

### DIFF
--- a/holmes/core/tracing.py
+++ b/holmes/core/tracing.py
@@ -1,12 +1,12 @@
-import os
+import getpass
 import logging
+import os
 import platform
-import pwd
 import socket
 from datetime import datetime
-from typing import Optional, Any, Union, Dict
-from pathlib import Path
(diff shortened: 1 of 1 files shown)


arikalon1 (approved):
nice work

PR #818, opened 2025-08-11
Strict tools

nherment:


holmes/common/env_vars.py
```diff
@@ -52,6 +52,13 @@ def load_bool(env_var, default: Optional[bool]) -> Optional[bool]:
 TOOL_CALL_SAFEGUARDS_ENABLED = load_bool("TOOL_CALL_SAFEGUARDS_ENABLED", True)
 IS_OPENSHIFT = load_bool("IS_OPENSHIFT", False)
 
+LLMS_WITH_STRICT_TOOL_CALLS = os.environ.get(
```
moshemorad, line 55: A. So sonnet isn't supporting that?
B. Instead of doing that this way can we maybe add it as arg to model list entry and use it from there?
C. Isn't litellm already handling those cases? i bump into somethink like https://github.com/BerriAI/litellm/blob/0288ed35da38910debd2b9e823aaf32e4dd4605d/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py#L281 when was working on structured output.

moshemorad: About A i just saw that indeed they aren't supporting it 😞

PR #823, opened 2025-08-12
chore: Add ai safety prompt to system prompt

nilo19:
chore: Add ai safety prompt to system prompt


holmes/plugins/prompts/generic_ask.jinja2
```diff
@@ -6,6 +6,8 @@ If you output an answer and then realize you need to call more tools or there ar
 If you have a good and concrete suggestion for how the user can fix something, tell them even if not asked explicitly
 {% include '_current_date_time.jinja2' %}
 
+{% include '_ai_safety.jinja2' %}
```
mainred: As a system prompt used for all cases, can _global_instructions.jinja2 cover all case? cc @aantn

nilo19: I didn't find a clear path how this template is applied to holmes ask.

mainred: Sorry, not _global_instructions.jinja2 but https://github.com/robusta-dev/holmesgpt/blob/master/holmes/plugins/prompts/_general_instructions.jinja2, which is shared across multiple system prompts for other entrypoints besides ask as well.

mainred: will be better to add a _safe_ai.jinja2 and include it in _general_instructions.jinja2

nilo19: done.

PR #839, opened 2025-08-13
Fix case sensitivity in model cost-lookup

aantn:
Possible fix for https://github.com/robusta-dev/holmesgpt/issues/246#issuecomment-3177467822


aantn:
Closing, replaced by https://github.com/robusta-dev/holmesgpt/pull/878/

PR #845, opened 2025-08-14
ROB-1824: do not use robusta ai api key if target model is not robusta ai

nherment:


holmes/config.py
```diff
@@ -477,8 +479,8 @@ def _get_llm(self, model_key: Optional[str] = None, tracer=None) -> "LLM":
             )
             api_key = model_params.pop("api_key", api_key)
             model = model_params.pop("model", model)
-
-        from holmes.core.llm import DefaultLLM
```
moshemorad, line 481: I think that this is intentionally done by @aantn to speed up the holmes cli startup.

holmes/config.py
```diff
@@ -477,8 +479,8 @@ def _get_llm(self, model_key: Optional[str] = None, tracer=None) -> "LLM":
...
+        if not api_key and "robusta.dev" in model_params.get("base_url", "") and self.api_key:
```
moshemorad: Thinking here loud, can't we just add it to the robusta model in the model_list when loading it and keep this logic generic? I think it will be good idea going forward. WDYT?

Also just making sure did you validate the cli and make sure it isn't breaking anything?

moshemorad (dismissed):

holmes/config.py
aantn: Not in this case! I've done lazy loading in general but haven't been able to avoid litellm imports. It can't really be deferred much as its too central. So fine by me.

holmes/config.py
nherment: @moshemorad do you mean adding the robusta ai specific code to load_model_from_file ?
It seems the same to me

moshemorad: What i meant is instead of having this logic that loads the secret if it is robusta we can load the secret as part of model list logic and just set it there once and for all. But if you feel it is the same than can ignore.

nherment: I think how you suggest it is a bit cleaner but the current way feels easier to find for future maintenance.

PR #848, opened 2025-08-14
Fixing loading configs from env vars and Prometheus Amp support

Avi-Robusta:


holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -76,6 +78,32 @@ def validate_prom_config(self):
...
+    aws_region: str
+    aws_service_name: str = "aps"
+    healthcheck: str = "api/v1/query?query=up"  # Override for AMP
+    prometheus_ssl_enabled: bool = False
+
+    def is_amp(self) -> bool:
+        return True
+
+
+def get_aws_auth(config: Union[PrometheusConfig, AMPConfig]):
```
moshemorad: please move it to config level e.g. implement it differently for each config instead of passing and checking if is_amp.

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -312,6 +378,12 @@ def _invoke(self, params: Any) -> StructuredToolResult:
                 error="Prometheus is not configured. Prometheus URL is missing",
                 params=params,
             )
+        if self.toolset.config.is_amp():
```
moshemorad, line 381: Instead of raising here and error could we not register this tool if it is amp?

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -759,28 +844,46 @@ def _reload_llm_instructions(self):
...
+    def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
+        try:
+            if config:
+                config_cls = self.determine_prometheus_class(config)
+                self.config = config_cls(**config)  # type: ignore
+
+                self._reload_llm_instructions()
+                return self._is_healthy()
+        except Exception:
+            logging.exception(
```
moshemorad: Why we done this change? if the user provided not working config we just try to resolve it our selves? i don't think it is a good idea.

moshemorad (approved):

PR #851, opened 2025-08-17
Subtasks

arikalon1:


holmes/core/openai_formatting.py
```diff
@@ -24,6 +24,26 @@ def type_to_open_ai_schema(param_attributes: Any, strict_mode: bool) -> dict[str
...
+        if hasattr(param_attributes, "items") and param_attributes.items:
+            items_schema = type_to_open_ai_schema(param_attributes.items, strict_mode)
+            type_obj = {"type": "array", "items": items_schema}
+        else:
+            # Fallback for arrays without explicit item schema
+            type_obj = {"type": "array", "items": {"type": "object"}}
+            if strict_mode:
+                type_obj["items"]["additionalProperties"] = False
     else:
         match = re.match(pattern, param_type)
```
nherment, line 48: The code below was handling type="array[...]" including "array[object]". It can be either deleted or merged with your recursive approach for objects.

```python
if match.group("inner_type"):
            inner_type = match.group("inner_type")
            if inner_type == "object":
                items_obj: dict[str, Any] = {"type": "object"}
                if strict_mode:
                    items_obj["additionalProperties"] = False
                type_obj = {"type": "array", "items": items_obj}
            else:
                type_obj = {"type": "array", "items": {"type": inner_type}}
```

holmes/core/todo_manager.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        Format tasks for injection into system prompt.
+        Returns empty string if no tasks exist.
+        """
+        tasks = self.get_session_tasks(session_id)
+
+        if not tasks:
+            return ""
+
+        status_order = {"pending": 0, "in_progress": 1, "completed": 2}
```
moshemorad: Please const it please use the enum.

holmes/core/todo_manager.py
```diff
@@ -0,0 +1,84 @@
...
+        progress_count = sum(1 for t in tasks if t.status == TaskStatus.IN_PROGRESS)
+        completed_count = sum(1 for t in tasks if t.status == TaskStatus.COMPLETED)
+
+        lines.append(
+            f"**Task Status**: {completed_count} completed, {progress_count} in progress, {pending_count} pending"
+        )
+        lines.append("")
+
+        for task in sorted_tasks:
+            status_indicator = {
```
moshemorad, line 68: Pleas const as well.

holmes/core/todo_manager.py
```diff
@@ -0,0 +1,84 @@
...
+                "pending": "[ ]",
+                "in_progress": "[~]",
+                "completed": "[✓]",
+            }.get(task.status.value, "[?]")
```
moshemorad: If not found do we want to raise sentry log error for that case?

holmes/core/todo_manager.py
```diff
@@ -0,0 +1,84 @@
...
+        self._sessions: Dict[str, List[Task]] = {}
+        self._lock: Lock = Lock()
+
+    def get_session_tasks(self, session_id: str) -> List[Task]:
+        with self._lock:
+            return self._sessions.get(session_id, []).copy()
+
+    def update_session_tasks(self, session_id: str, tasks: List[Task]) -> None:
+        with self._lock:
+            self._sessions[session_id] = tasks.copy()
```
moshemorad, line 23: out of curiosity, why do we need copy here? 
Also if we need copy since it is list we sure we don't need deepcopy?

holmes/config.py
```diff
@@ -74,7 +74,7 @@ class Config(RobustaBaseConfig):
         None  # if None, read from OPENAI_API_KEY or AZURE_OPENAI_ENDPOINT env var
     )
     model: Optional[str] = "gpt-4o"
-    max_steps: int = 10
+    max_steps: int = 40
```
mainred, line 77: why we bump the default max steps to 40?

holmes/core/tool_calling_llm.py
```diff
@@ -207,6 +211,7 @@ def __init__(
         self.max_steps = max_steps
         self.tracer = tracer
         self.llm = llm
+        self.investigation_id = str(uuid.uuid4())
```
moshemorad, line 214: it look alright in most of our cases since we creating new toolcalling llm in each request. but look like in cli interactive it might be an issue and the investigation id will be shared in all the session. Is that intended?

checked it and looks like when asking multiple question the tasks ids are continue instead of reset. ignore if it is intended.

holmes/plugins/toolsets/investigator/core_investigation.py
```diff
@@ -0,0 +1,159 @@
...
+    def __init__(self):
+        super().__init__(
+            name="core_investigation",
+            description="Core investigation tools for task management and planning",
+            enabled=True,
+            tools=[TodoWriteTool()],
+            tags=[ToolsetTag.CORE],
+            is_default=True,
+        )
+        self.status = ToolsetStatusEnum.ENABLED
```
moshemorad: Not sure i understand why we need it here?

holmes/plugins/toolsets/investigator/core_investigation.py
```diff
@@ -0,0 +1,159 @@
...
+
+            response_data = f"✅ Investigation plan updated with {len(tasks)} tasks. Tasks are now stored in session and will appear in subsequent prompts.\n\n"
+            if formatted_tasks:
+                response_data += formatted_tasks
+            else:
+                response_data += "No tasks currently in the investigation plan."
+
+            return StructuredToolResult(
+                status=ToolResultStatus.SUCCESS,
+                data=response_data,
```
moshemorad, line 119: Do we think sending the task list to the AI as part of the tool call + prompt isn't redundant?

holmes/config.py
arikalon1: the todo tools add more steps. So we needed to increase the max

holmes/plugins/toolsets/investigator/core_investigation.py
arikalon1: might be, but this will be used by the ui for presentation
we might change it afterwards

holmes/plugins/toolsets/investigator/core_investigation.py
arikalon1: you're right, it's not required.
Removing it

holmes/core/tool_calling_llm.py
arikalon1: yeah, that's correct

But I think it works ok even in interactive sessions

holmes/core/todo_manager.py
arikalon1: just being extra defensive

holmes/plugins/toolsets/investigator/core_investigation.py
```diff
@@ -0,0 +1,159 @@
...
+    Tool,
+    StructuredToolResult,
+    ToolResultStatus,
+)
+from holmes.plugins.toolsets.investigator.model import Task, TaskStatus
+
+
+class TodoWriteTool(Tool):
+    name: str = "TodoWrite"
+    description: str = "Save investigation tasks to break down complex problems into manageable sub-tasks. ALWAYS provide the COMPLETE list of all tasks, not just the ones being updated."
```
mainred, line 23: The list of calls should have no dependencies, so it's ok to run them in parallel.

holmes/core/todo_manager.py
arikalon1: good point, fixed

holmes/core/todo_manager.py
arikalon1: It's not possible, so not adding a log line here

holmes/core/todo_manager.py
arikalon1: fixed

holmes/plugins/toolsets/investigator/core_investigation.py
```diff
@@ -0,0 +1,157 @@
...
+    ToolsetTag,
+    ToolParameter,
+    Tool,
+    StructuredToolResult,
+    ToolResultStatus,
+)
+from holmes.plugins.toolsets.investigator.model import Task, TaskStatus
+
+
+class TodoWriteTool(Tool):
```
mainred, line 21: why not move it to the tools.py?

holmes/plugins/prompts/investigation_procedure.jinja2
```diff
@@ -0,0 +1,210 @@
+{% if investigation_id %}
```
mainred, line 1: I am curious, why investigation_id is so important. It's an uuid across one interactive or non-interactive mode and marking on session.

mainred (commented):
Sorry, my comments are late.

holmes/plugins/prompts/investigation_procedure.jinja2
mainred: Yes, it's important to track one tasks running in parallel. These can be multiple session in one interactive mode ask.

PR #855, opened 2025-08-19
Rob 1931 bash toolset replaces cli toolsets

nherment:


holmes/plugins/toolsets/bash/parse_command.py
```diff
@@ -1,39 +1,108 @@
...
-    create_kubectl_parser(commands_parser)
-    create_grep_parser(commands_parser)
+    # Add all BashCommand classes
+    for command in commands:
+        command.add_parser(commands_parser)
+
     return parser
 
 
+def validate_command(
```
moshemorad, line 86: Is that intentional? Not sure i understand what is going on here (I do see that it is used)
Can you explain the usage since the function only use pass?

holmes/plugins/toolsets/bash/parse_command.py
```diff
@@ -1,39 +1,108 @@
...
+from holmes.plugins.toolsets.bash.utilities.tail import TailCommand
+from holmes.plugins.toolsets.bash.utilities.tr import TrCommand
+from holmes.plugins.toolsets.bash.utilities.base64_util import Base64Command
+from holmes.plugins.toolsets.bash.utilities.jq import JqCommand
+from holmes.plugins.toolsets.bash.utilities.sed import SedCommand
+from holmes.plugins.toolsets.bash.utilities.grep import GrepCommand
+
+
+# All commands now use BashCommand classes
+commands: list[BashCommand] = [
```
moshemorad: A. consider to rename to AVAILABLE_COMMANDS, in some flow it is confusing littlebit to keep up with the terminology.
B. consider to use dict comp and build it that way when you need to get specific command you will be able to get it by the key instead of searching the list. Will make code little bit cleaner.

holmes/plugins/toolsets/bash/utilities/base64_util.py
```diff
@@ -0,0 +1,12 @@
+from holmes.plugins.toolsets.bash.common.bash_command import (
+    SimpleBashCommand,
+)
+
+
+class Base64Command(SimpleBashCommand):
```
moshemorad, line 6: @nherment out of curiousity in which cases we saw that the llm used base64 command?

nherment: I've only seen it used when decoding secrets.
However the approach here was to add all the utilities I could think of as long as they are safe and easy to add support for. This is to increase the chances for LLM commands to be already in the whitelist.

holmes/plugins/toolsets/bash/parse_command.py
nherment: I separated it from stringify but never moved the code there. It's done now.

PR #863, opened 2025-08-19
Docs: update mkdocs port

aantn:
I'm choosing a port less likely to conflict with other tooling commonly deployed


pavangudiwada (approved):
LGTM

PR #864, opened 2025-08-19
Update default models

aantn:


holmes/core/llm.py
```diff
@@ -466,6 +466,7 @@ def _load_default_robusta_config(self):
                 "name": ROBUSTA_AI_MODEL_NAME,
                 "base_url": ROBUSTA_API_ENDPOINT,
                 "is_robusta_model": True,
+                # This seems like a mistake?
```
aantn: @moshemorad I'm not familiar with the code, but what's the implication of setting this to gpt-4o? Does that lead to an incorrect max token count? 👇

I think @nherment was working on fixes for this, but not sure if its in this code or somewhere else

holmes/core/llm.py
```diff
@@ -535,6 +536,7 @@ def _create_robusta_model_entry(
         self, model_name: str, args: Optional[dict[str, Any]] = None
     ) -> dict[str, Any]:
         entry = self._create_model_entry(
+            # this also seems like a mistake? (wont token counts be wrong etc)
```
aantn: @moshemorad and @nherment - same comment here.

holmes/core/llm.py
moshemorad: @aantn this is some tech debt we have. It can cause max_token & token_count to behave incorrectly AFAIK. 

The current way to handle the context window size is either pass env var to override it or args which are used in `update_custom_args` in the DefaultLLM.

The reason we are setting the model here to gpt-4o is to support the relay endpoints and communication way. we need to maybe consider passing the actual model that being used in another propetry and use it here. But we should tackle it as part of other PR.

holmes/core/llm.py
moshemorad: Same as above.

moshemorad (dismissed):

holmes/core/llm.py
aantn: @moshemorad thanks, so just to clarify I'll leave a `TODO: tech debt` here and in the other place, and not change either as part of this PR. Works?

Issue #873, opened 2025-08-19
argocd version check takes infinite time and HolmesGPT doesn't run


pavangudiwada:
### What happened?

Sometimes, Holmes's pre-requisite test of argocd keeps going without any result. 

### What did you expect to happen?

Holmes to complete pre-requisite checks and work fine. 

### How can we reproduce it (as minimally and precisely as possible)?

Run 

```
ITERATIONS=1 RUN_LIVE=true poetry run pytest --no-cov tests/llm/test_ask_holmes.py -s -k '28_'
```

PR #874, opened 2025-08-19
Added tool timeout for when it takes too long to run pre-check

pavangudiwada:
Helps in cases like #873

### DIFF
--- a/holmes/core/tools.py
+++ b/holmes/core/tools.py
@@ -429,6 +429,7 @@ class Toolset(BaseModel):
                         text=True,
                         stdout=subprocess.PIPE,
                         stderr=subprocess.PIPE,
+                        timeout=60,
                     )
                     if (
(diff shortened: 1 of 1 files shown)


moshemorad (approved):

holmes/core/tools.py
```diff
@@ -453,7 +471,8 @@ def check_prerequisites(self):
 
             elif isinstance(prereq, CallablePrerequisite):
                 try:
-                    (enabled, error_message) = prereq.callable(self.config)
+                    config = self.config if self.config is not None else {}
```
moshemorad: Why we did this change?

pavangudiwada: ~This prevents Holmes from crashing when checking if a tool is available, especially when no configuration has been set up yet.~ You're right, it doesn't seem to be necessary. Fixing it.

PR #876, opened 2025-08-19
Improve description handling in JiraSource: extract from issue if not…

greg-savage:
… provided and handle various formats


aantn:
Hi @greg-savage I noticed you closed the PR right away? Should we review?

PR #885, opened 2025-08-21
Fix tool call numbering to maintain sequential order across iterations

pavangudiwada:
### Before 
image

### After
CleanShot 2025-08-21 at 10 15 59@2x


aantn (approved):
LGTM. Thanks.

PR #891, opened 2025-08-22
Docs: Add instructions on choosing model when using HTTP API

aantn:
@pavangudiwada can you please review + test with an Azure model? I believe there was a bug in the previous version of these docs that @aritraghosh checked.


pavangudiwada:
@aantn I believe the config.model is not being passed as an environment variable. 

We'll have to add it as an env, like so
```
additionalEnvVars:
  - name: MODEL
    value: "azure/gpt-4.1"
```

After passing the model as an env, I got the end point to work. 

CleanShot 2025-08-24 at 07 01 39@2x

Re: Testing with the model name in the query, I'm not able to verify it works. HolmesGPT logs show that the model used is the same model mentioned in the values file and not the one used in the query. 
CleanShot 2025-08-24 at 07 09 08@2x

In short, the docs changes don't work

docs/installation/kubernetes-installation.md
```diff
@@ -118,6 +137,11 @@ kubectl port-forward svc/holmesgpt-holmes 8080:80
 # kubectl port-forward svc/{your-release-name}-holmes 8080:80 -n {your-namespace}
 
 # Test with a basic question
+curl -X POST http://localhost:8080/api/chat \
+  -H "Content-Type: application/json" \
+  -d '{"ask": "list pods in namespace default?", "model": "gpt-4o-mini"}'
```
aantn, line 150: @pavangudiwada I checked with @arikalon1 and this needs to be the name of a model configured in the modelList for helm - see https://holmesgpt.dev/ai-providers/using-multiple-providers/

Do you mind testing this and verifying, then updating the docs?

note that if modelList is:

```
modelList:
  foo:
    model: openai/gpt-4o
    temperature: 0
```

then in the curl command you would use `"model": "foo"` not `"model": "openai/gpt-4o"`

aantn:
@pavangudiwada after speaking to Arik, I think we can document the method to use `modelList` only (and tell users they must send the name of a model configured in the modelList to the HTTP API)

docs/installation/kubernetes-installation.md
pavangudiwada: Updated with the proper fix

pavangudiwada (approved):
LGTM

PR #894, opened 2025-08-22
evals: add support for model-comparison and cost calculation

aantn:
This PR includes 3 main changes:

1. It allows running evals with multiple models at once and comparing them

Screenshot 2025-08-23 at 10 57 33

2. It adds cost information to evals in the cli report

Screenshot 2025-08-23 at 10 57 16

3. To support (2) it modifies some files in holmes itself (not just evals framework) to track costs. Exposing this to users was not the primary goal of this PR, but I added a cli-flag to allow it.

### DIFF
--- a/tests/llm/test_investigate.py
+++ b/tests/llm/test_investigate.py
@@ -200,7 +200,7 @@ def test_investigate(
         )
         raise
 
-    tools_called = [t.tool_name for t in result.tool_calls]
+    tools_called = [t.tool_name for t in result.tool_calls] if result.tool_calls else []
     update_test_results(request, output, tools_called, scores, result)
 
     assert result.sections, "Missing sections"
--- a/tests/llm/test_workload_health.py
+++ b/tests/llm/test_workload_health.py
@@ -165,7 +165,9 @@ def test_health_check(
                     tags=tags,
                 )
 
-            tools_called = [t.tool_name for t in result.tool_calls]
+            tools_called = (
+                [t.tool_name for t in result.tool_calls] if result.tool_calls else []
+            )
             print(f"\n** TOOLS CALLED **\n{tools_called}")
             print(f"\n** OUTPUT **\n{output}")
             print(f"\n** SCORES **\n{scores}")


holmes/core/tool_calling_llm.py
```diff
@@ -299,6 +312,36 @@ def call(  # type: ignore
                 )
                 logging.debug(f"got response {full_response.to_json()}")  # type: ignore
 
+                # Log and accumulate cost information if available
+                try:
+                    cost_value = (
+                        full_response._hidden_params.get("response_cost", 0)
```
arikalon1: Is the response cost a field returned by the llms?
Couldn't find the documentation of it. Tried to understand what it includes

tests/llm/test_ask_holmes.py
```diff
@@ -54,9 +55,17 @@ def get_ask_holmes_test_cases():
     return get_test_cases(TEST_CASES_FOLDER)
 
 
+def get_models():
```
arikalon1: I would name the function, `get_tests_models`
This function defined twice, might be better to just reuse it

I think it's not backward compatible, because the default here is `gpt-4o` and not what the user define in the old param, `MODEL`

tests/llm/test_investigate.py
```diff
@@ -72,9 +73,17 @@ def get_investigate_test_cases():
     return get_test_cases(TEST_CASES_FOLDER)
 
 
+def get_models():
```
arikalon1: same

docs/development/evals/index.md
```diff
@@ -64,6 +64,34 @@ RUN_LIVE=true MODEL=azure/your-deployment-name CLASSIFIER_MODEL=azure/your-deplo
...
 - The model specified here is passed directly to LiteLLM, so any model supported by LiteLLM can be used
 
+### Multi-Model Benchmarking
+
+HolmesGPT supports running evaluations across multiple models simultaneously to compare their performance:
+
+```bash
+# Test multiple models in a single run
+# Models are specified as comma-separated list
+RUN_LIVE=true MODELS=gpt-4o,anthropic/claude-3-5-sonnet-20241022,gpt-4o-mini \
```
arikalon1: I think I would rename the new env var, to `TESTS_MODELS` or `EVALS_MODELS`
To make it clear, it's only used for testing

tests/llm/test_ask_holmes.py
aantn: This was a mistake. I suggest we just use the old parameter MODEL (no `s`) and it can be a single model or multiple. Does that work?

docs/development/evals/index.md
aantn: As asked above, this was a mistake. I meant to reuse MODEL. Are you ok with that approach?

PR #905, opened 2025-08-25
Improve historical data querying for logs and metrics

aantn:
Needs further testing


arikalon1 (approved):
looks good, nice work

PR #906, opened 2025-08-25
ROB-1932 bash tool cli approval

nherment:


holmes/core/tool_calling_llm.py
```diff
@@ -562,6 +565,42 @@ def _invoke_tool(
...
+                    error_message = tool_response.error or "Command requires approval"
+
+                    approved, feedback = self.approval_callback(command, error_message)
+
+                    if approved:
+                        # Re-execute the tool with approval
+                        logging.debug(f"User approved command: {command}")
+                        # We need to modify the tool to bypass validation - this requires updating the bash toolset
+                        # For now, let's execute the command directly through the bash execution
+                        from holmes.plugins.toolsets.bash.common.bash import (
```
moshemorad: Isn't a better way to do that? maybe calling invoke with extra param that pass the info that this call was approved? 

I understand that it is faster to do it that way but feels like we are keeping complicating our code.

Not sure about it, so up to you.

holmes/core/tool_calling_llm.py
```diff
@@ -562,6 +565,42 @@ def _invoke_tool(
...
 
+            # Handle approval required status
+            if tool_response.status == ToolResultStatus.APPROVAL_REQUIRED:
+                if self.approval_callback is not None:
+                    command = tool_response.invocation or str(tool_params)
+                    error_message = tool_response.error or "Command requires approval"
+
+                    approved, feedback = self.approval_callback(command, error_message)
+
+                    if approved:
```
moshemorad: I think that we might have issue here. What will happen if multiple tool request will result in approval request? Each one of them is from different thread and both of them updating the cli. 

Were you able to bump into this kind of situation? if not worth to try and simulate it and see how it beheaves.

holmes/core/tool_calling_llm.py
nherment: Yes I thought of it but went the hackish way.
This is fixed now. All tool can implement required approval.

holmes/core/tool_calling_llm.py
nherment: Good catch. This is now fixed. All tools are executed and then there is a synchronous handling for any tool that requires approval

holmes/core/tool_calling_llm.py
```diff
@@ -524,92 +586,97 @@ def _invoke_tool(
...
+                tool_params=tool_call_result.result.params or {},
+                user_approved=True,
+                trace_span=DummySpan(),
+                tool_number=None,  # Could be extracted if needed
+            )
+            tool_call_result.result = new_response
+        else:
+            # User denied - update to error
+            feedback_text = f" User feedback: {feedback}" if feedback else ""
+            tool_call_result.result.status = ToolResultStatus.ERROR
```
moshemorad, line 685: Thinking here loud, maybe we could have another status for this case? WDYT?

holmes/core/tool_calling_llm.py
```diff
@@ -524,92 +586,97 @@ def _invoke_tool(
...
+            logging.debug(
+                f"User approved command: {tool_call_result.result.invocation}"
+            )
+
+            new_response = self._directly_invoke_tool(
+                tool_name=tool_call_result.tool_name,
+                tool_params=tool_call_result.result.params or {},
+                user_approved=True,
+                trace_span=DummySpan(),
+                tool_number=None,  # Could be extracted if needed
```
moshemorad: I think we need this, we use the tool number for the prints in the cli also for showing the tool output by the tool number in interactive mode.

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -179,23 +185,28 @@ def _invoke(self, params: Dict[str, Any]) -> StructuredToolResult:
...
+                command_to_execute = make_command_safe(command_str, self.toolset.config)
+
+            except (argparse.ArgumentError, ValueError) as e:
+                with sentry_sdk.configure_scope() as scope:
+                    scope.set_extra("command", command_str)
+                    scope.set_extra("error", str(e))
+                    scope.set_extra("unsafe_allow_all", BASH_TOOL_UNSAFE_ALLOW_ALL)
+                    sentry_sdk.capture_exception(e)
+
+                if not BASH_TOOL_UNSAFE_ALLOW_ALL:
```
moshemorad, line 201: Should we adjust this logic as well? e.g. if allow all it just runs the command and if `ENABLE_CLI_TOOL_APPROVAL` we need to run the approval flow? WDYT?

holmes/plugins/toolsets/newrelic.py
```diff
@@ -130,7 +134,7 @@ def error(msg: str) -> StructuredToolResult:
                 params=params,
             )
 
-        duration = params.get("duration")
+        duration = get_param_or_raise(params, "duration")
```
moshemorad, line 137: Nice catch :)

moshemorad (changes requested):
Great work, few comments.

holmes/core/tool_calling_llm.py
nherment: I think we should have UI specs first and only then implement what's needed on the backend. Then if we need to differentiate a denied command in the UI we can add a new status

holmes/plugins/toolsets/bash/bash_toolset.py
nherment: unless I'm misunderstanding this is already the case.
- `BASH_TOOL_UNSAFE_ALLOW_ALL` makes it so that all commands are executed, even unsafe ones, without any approval required.
- `ENABLE_CLI_TOOL_APPROVAL` enables the approval workflow for the CLI.

The decision to enable or not the approval workflow is done by the CLI code. Tools should not know whether approval is enabled or not but only need to decide if a command is safe or not by setting the status to `APPROVAL_REQUIRED`.

moshemorad: You are correct, missed it.

moshemorad (approved):

PR #907, opened 2025-08-25
Add a test for traces due to db slowness

Sheeproid:


tests/llm/utils/mock_toolset.py
```diff
@@ -565,6 +566,35 @@ def _configure_toolsets(
...
                 toolset = new_runbook_toolset
+            elif toolset.name == "kubernetes/core":
+                if not isinstance(toolset, YAMLToolset):
+                    raise ValueError(
+                        f"Expected kubernetes/core to be YAMLToolset, got {type(toolset)}"
+                    )
+                yaml_toolset: YAMLToolset = toolset
+
+                # Block secret access to prevent LLM from reading code hints in secrets
+                security_check = """if [ "{{ kind }}" = "secret" ] || [ "{{ kind }}" = "secrets" ]; then
```
aantn, line 577: Nice

tests/llm/test_ask_holmes.py
```diff
@@ -62,6 +62,13 @@ def test_ask_holmes(
     # Set initial properties early so they're available even if test fails
     set_initial_properties(request, test_case, model)
 
+    # Check if --only-setup is set
```
aantn, line 65: @Sheeproid I think this is a mistake and shouldn't be here. I moved it in a recent pr to `check_and_skip_test(test_case)` below which also reports the Skip to braintrust properly

aantn: Ah i see its the new option you added. Can you move it to the same location please? Its used there for investigate tests too

aantn (changes requested):

aantn (approved):

PR #914, opened 2025-08-26
[ROB-1923] - added amp holmes irsa support

Avi-Robusta:
Verified backwards compatibility with keys also
```
holmes:
  serviceAccount:
    annotations:
      eks.amazonaws.com/role-arn: arn:aws:iam::...:role/amp-iamproxy-ingest-role_name_here
  toolsets:
    prometheus/metrics:
      config:
        prometheus_url: ...
        aws_region: us-east-2
        aws_service_name: aps
        prometheus_ssl_enabled: true
        additional_labels:
          cluster: my-cluster-label
      enabled: true
```

### DIFF
--- a/holmes/plugins/toolsets/prometheus/prometheus.py
+++ b/holmes/plugins/toolsets/prometheus/prometheus.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import boto3
 import os
 import re
 import time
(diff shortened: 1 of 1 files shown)


holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -86,22 +87,53 @@ def get_auth(self) -> Any:
...
-    aws_access_key: str
-    aws_secret_access_key: str
+    aws_access_key: Optional[str] = None
+    aws_secret_access_key: Optional[str] = None
     aws_region: str
     aws_service_name: str = "aps"
-    healthcheck: str = "api/v1/query?query=up"  # Override for AMP
-    prometheus_ssl_enabled: bool = False
+    healthcheck: str = "api/v1/query?query=up"
+    prometheus_ssl_enabled: bool = True
```
arikalon1: Why did you change the default?
Wouldn't think brake the auto discovery, and connecting to prometheus running in the cluster?

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -86,22 +87,53 @@ def get_auth(self) -> Any:
...
+        if self.aws_access_key and self.aws_secret_access_key:
+            return AWS4Auth(
+                self.aws_access_key,
+                self.aws_secret_access_key,
+                self.aws_region,
+                self.aws_service_name,
+            )
+        return None
+
+    def get_auth(self):
```
arikalon1, line 126: I would rename it to `get_aws_auth`, since it's only for aws

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -86,22 +87,53 @@ def get_auth(self) -> Any:
...
-            self.aws_region,  # type: ignore
-            self.aws_service_name,  # type: ignore
+            frozen.access_key,
+            frozen.secret_key,
+            self.aws_region,
+            self.aws_service_name,
+            session_token=frozen.token,
+        )
+
+    def _build_static_auth(self) -> Optional[AWS4Auth]:
```
arikalon1: I would rename it to `_build_static_aws_auth`, since it's only for aws

arikalon1 (approved):
nice work, left a few minor comments

holmes/plugins/toolsets/prometheus/prometheus.py
Avi-Robusta: `get_auth` is an inherited function in the base PrometheusConfig they both have the so we can run 
`self.toolset.config.get_auth() `in the tools without caring which type of prometheus config it is. So renaming would break it

PR #916, opened 2025-08-28
Enhance prompt caching

arikalon1:


holmes/core/llm.py
```diff
@@ -266,3 +268,60 @@ def get_maximum_output_token(self) -> int:
...
+                target_msg = msg
+                break
+
+        if not target_msg:
+            logging.debug("No non-user message found for cache_control")
+            return
+
+        content = target_msg.get("content")
+
+        if content is None:
```
moshemorad, line 305: @arikalon1 in which cases this could happen?

holmes/core/llm.py
```diff
@@ -229,9 +229,11 @@ def completion(
             ]  # can be removed after next litelm version
 
         self.args.setdefault("temperature", temperature)
+
+        self._add_cache_control_to_last_message(messages)
```
moshemorad, line 233: I saw the litellm will ignore it on unsupported models. But maybe it is a good idea anyway to check if the model is supporting it before or not?

https://docs.litellm.ai/docs/completion/prompt_caching#check-model-support

moshemorad (approved):

holmes/core/llm.py
arikalon1: Their docs said that unsupported models just ignores it.
I also verified it works with gpt-4o
So I think it's better not to add an extra validation here

holmes/core/llm.py
arikalon1: not sure it can happen, just to be on the safe side

PR #920, opened 2025-08-29
ROB-1807: add eval for disk space PVC issue

nherment:


tests/llm/fixtures/test_ask_holmes/157_disk_full_statefulset/manifest.yaml
```diff
@@ -0,0 +1,286 @@
...
+                    # Progress reporting
+                    if batch_id % 25 == 0:
+                        print(f"[{datetime.now()}] Progress update: Processed {batch_id * self.batch_size} user interactions")
+                        print(f"[{datetime.now()}] Quality metrics: Data validation passed, feature extraction complete")
+
+                    # Small delay between batches
+                    time.sleep(0.01)
+
+                except OSError as e:
+                    if "No space left on device" in str(e):
```
Sheeproid, line 195: @nherment the test looks good. The only note is that the errors are currently being explicitly written in the Python application code (e.g., CRITICAL ERROR: Failed to write batch {batch_id} - storage capacity exceeded). This might bias the AI toward that specific problem.

Instead, it’s preferred that no explicit code/log handles the OS error. The OS error should be allowed to crash the app naturally, and the exception should be written out as-is.

nherment: Yes I tried to avoid that but by default kind does not enforce disk space limits for PVCs. 
What I've done now is use longhorn which does enforce the allocated space limit.

I do think it's a good approach for the eval code to artificially generate errors as it makes these tests easier to build and maintain.

PR #922, opened 2025-08-31
Add tempo comparison test tool and pass test 114

Sheeproid:


holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -64,6 +65,52 @@ def get_example_config(self):
...
         return cast(GrafanaTempoConfig, self._grafana_config)
 
+    def build_k8s_filters(
+        self, params: Dict[str, Any], use_exact_match: bool = True
+    ) -> List[str]:
+        """Build TraceQL filters for k8s parameters.
+
+        Args:
+            params: Dictionary containing k8s parameters
+            use_exact_match: If True, uses exact match (=), if False uses regex match (=~)
```
aantn, line 77: @Sheeproid what's the reason for defaulting to `use_exact_match: true` instead of `false`? I would think that a default of fuzzy-matching can lead to better results

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -291,14 +319,245 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                    type="string",
+                    required=False,
+                ),
+                "sample_count": ToolParameter(
+                    description="Number of traces to fetch from each category (fastest/slowest). Default 3",
+                    type="integer",
+                    required=False,
+                ),
+                "start_datetime": ToolParameter(
+                    description="Start time for analysis (RFC3339 or relative)",
```
aantn: Should we mention in the description the default if not passed so that the LLM knows? I can only see an advantage to giving the LLM the extra info - I wouldn't expect there to be a downside.

tests/llm/fixtures/shared/tempo.yaml
```diff
@@ -2,7 +2,19 @@
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: tempo-storage
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 500Mi
```
aantn, line 15: This seems a bit big. Do we really need that much space for the evals?

tests/llm/fixtures/shared/tempo.yaml
```diff
@@ -80,9 +97,10 @@ data:
               endpoint: 0.0.0.0:4318
 
     ingester:
-      max_block_duration: 5m
-      trace_idle_period: 30s
-      max_block_bytes: 10_000_000
+      max_block_duration: 24h
```
aantn, line 100: Perhaps make it longer, even 1 week. Don't see a downside given that we're not sending a ton of trace data

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -154,25 +200,7 @@ def _invoke(self, params: Dict) -> StructuredToolResult:
...
-            filters.append(
-                f'{prefix}{labels.namespace}="{params.get("namespace_name")}"'
-            )
-        if params.get("deployment_name"):
-            filters.append(
-                f'{prefix}{labels.deployment}="{params.get("deployment_name")}"'
-            )
-        if params.get("node_name"):
-            filters.append(f'{prefix}{labels.node}="{params.get("node_name")}"')
+        filters = self._toolset.build_k8s_filters(params, use_exact_match=True)
```
aantn, line 205: Shouldn't use_exact_match be false?

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -291,14 +319,245 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+- For service latency: service_name="payment" (matches "payment-service" too)
+- For namespace issues: namespace_name="production"
+- Combined: service_name="auth", namespace_name="staging\"""",
+            parameters={
+                "service_name": ToolParameter(
+                    description="Service to analyze (partial match supported)",
+                    type="string",
+                    required=False,
+                ),
+                "pod_name": ToolParameter(
```
aantn, line 342: note that the comment on `partial match supported` is currently wrong as the code is not using that parameter internally. same for other parameters here.

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -291,14 +319,245 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
         return f"{toolset_name_for_one_liner(self._toolset.name)}: Fetched Tempo Trace (trace_id={params.get('trace_id')})"
 
 
+class FetchTracesSimpleComparison(Tool):
+    def __init__(self, toolset: BaseGrafanaTempoToolset):
+        super().__init__(
+            name="fetch_tempo_traces_comparative_sample",
+            description="""Fetches statistics and representative samples of fast, slow, and typical traces for performance analysis.
```
aantn: Should we tell the LLM in the description that you need to provide ONE OF pod_name, service_name, etc? (I think that's what the code actually expects right?)

tests/llm/fixtures/shared/tempo.yaml
Sheeproid: I had issues traces disappearing so I made everything related to storage and blocks bigger until it worked.
I prefer not investing more time in testing at this point.

tests/llm/fixtures/shared/tempo.yaml
Sheeproid: I had issues traces disappearing so I made everything related to storage and blocks bigger until it worked.
I prefer not investing more time in testing at this point.

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
Sheeproid: The default here is not relevant, it is always chosen in another place as pass as parameter. I"ll just remove it.

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
Sheeproid: Yes. Also, I think the default is wrong in GetTempoTraces. I"ll fix it in all places

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
Sheeproid: the old tool GetTempoTraces does not support regex, thus we should use use_exact_match=True.

aantn: any reason not to add it?

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
Sheeproid: it uses 'use_exact_match=False', it uses a regex and places it inside a .* and .*

aantn: no, this is fine. (although if a full regex is supported we could expose it too)
I thought we were setting `use_exact_match=True` on that flow so the comment was wrong. I might have gotten confused with the double negative, in which case ignore my previous comment

PR #927, opened 2025-09-02
New api for tempo used by current tempo toolset

Sheeproid:


aantn:
@Sheeproid I think this is an old PR and does not need review so I am closing it. Please confirm this is correct.

PR #938, opened 2025-09-04
don't trigger build on push non-master branch

mainred:
Pull requests triggers duplicated builds for both push and pull_request. Limit the push event to master only to not trigger a redundant build.

image

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -1,4 +1,4 @@
-# this runs on every commit/PR to test that we are properly building binaries that work
+# This runs on every commit/PR to test that we are properly building binaries that work
 # we test this on each commit/PR to catch build problems early
 name: Build and test HolmesGPT
 
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -2,7 +2,12 @@
 # we test this on each commit/PR to catch build problems early
 name: Build and test HolmesGPT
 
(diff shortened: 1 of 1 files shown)


mainred:
cc @moshemorad

.github/workflows/build-and-test.yaml
```diff
@@ -2,7 +2,12 @@
 # we test this on each commit/PR to catch build problems early
 name: Build and test HolmesGPT
 
-on: [push, pull_request, workflow_dispatch]
+on:
```
moshemorad, line 5: This will cause issue when pushing commit on open pr since this workflow won't be triggered. 

Just to double check it i saw this pr need update branch and i triggered it and indeed this is the case and the actions are stuck on pending status.

mainred: > This will cause issue when pushing commit on open pr since this workflow won't be triggered.

Here we have 3 triggers.
the first one is by my opening the PR, and the second one is you helped merge the master, the last one is when I pushed a commit to this PR.
image

mainred: The explanation of the pull_request from the [official document](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)
_if no activity types are specified, the workflow runs when a pull request is opened or reopened or **when the head branch of the pull request is updated**._

moshemorad (approved):

PR #941, opened 2025-09-04
Add Freeform tempo toolset

Sheeproid:


holmes/plugins/toolsets/grafana/grafana_tempo_api.py
```diff
@@ -0,0 +1,473 @@
...
+    def __init__(self, status_code: int, response_text: str, url: str):
+        self.status_code = status_code
+        self.response_text = response_text
+        self.url = url
+
+        # Try to extract error message from JSON response
+        try:
+            import json
+
+            error_data = json.loads(response_text)
```
aantn, line 32: Any specific reason to parse this and not just always return response_text? Is there something big we are trying to filter out of the error message in the other fields?

holmes/plugins/toolsets/grafana/grafana_tempo_api.py
```diff
@@ -0,0 +1,473 @@
...
+            error_message = response_text
+
+        super().__init__(f"Tempo API error {status_code}: {error_message}")
+
+
+class GrafanaTempoAPI:
+    """Python wrapper for Grafana Tempo REST API.
+
+    This class provides a clean interface to all Tempo API endpoints,
+    supporting both GET and POST methods based on configuration.
```
aantn, line 50: Could we remove the POST support? It seems like a mistake. (There might be specific endpoints that require POST but having this as a global flag doesn't make sense.)

holmes/plugins/toolsets/grafana/grafana_tempo_api.py
```diff
@@ -0,0 +1,473 @@
...
+        self.base_url = get_base_url(config)
+        self.headers = build_headers(config.api_key, config.headers)
+        self.use_post = use_post
+
+    def _make_request(
+        self,
+        endpoint: str,
+        params: Optional[Dict[str, Any]] = None,
+        path_params: Optional[Dict[str, str]] = None,
+        timeout: int = 30,
```
aantn, line 70: I would make this dramatically shorter or remove it altogether. (Why do we need retries at all?) With 30s per timeout * 3 retries, a single call to tempo could take 90s seconds.

holmes/plugins/toolsets/grafana/grafana_tempo_api.py
```diff
@@ -0,0 +1,473 @@
...
+                    timeout=timeout,
+                )
+            response.raise_for_status()
+            return response.json()
+
+        try:
+            return make_request()
+        except requests.exceptions.HTTPError as e:
+            # Extract detailed error message from response
+            response = e.response
```
aantn, line 127: See comments in the exception class. Could all this code could be much simpler. Something like raise Exception(f"{e.response}")?

holmes/plugins/toolsets/grafana/grafana_tempo_api.py
```diff
@@ -0,0 +1,473 @@
...
+
+        @backoff.on_exception(
+            backoff.expo,
+            requests.exceptions.RequestException,
+            max_tries=retries,
+            giveup=lambda e: isinstance(e, requests.exceptions.HTTPError)
+            and getattr(e, "response", None) is not None
+            and e.response.status_code < 500,
+        )
+        def make_request():
```
aantn, line 103: Use of the inner function here seems uncessary. We could just post the code directly in try/except - no need to define a function.

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -426,20 +413,31 @@ def _invoke(
...
-                base_url=base_url,
-                api_key=self._toolset.grafana_config.api_key,
-                headers=self._toolset.grafana_config.headers,
-                query=stats_query,
+
+            # Debug log the query (useful for troubleshooting)
+            import logging
+
+            logger = logging.getLogger(__name__)
+            logger.info(f"Tempo query: {stats_query}")
```
aantn, line 245: We should probably remove these. They'll spam CLI output

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                    description="Start time in Unix epoch seconds",
+                    type="integer",
+                    required=False,
+                ),
+                "end": ToolParameter(
+                    description="End time in Unix epoch seconds",
+                    type="integer",
+                    required=False,
+                ),
+                "spss": ToolParameter(
```
aantn, line 407: Should we set a default for this?

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                "Uses the Tempo API endpoint: GET /api/search with 'q' parameter. "
+                'TraceQL allows complex filtering like: {resource.service.name="api"} && {span.http.status_code=500}'
+            ),
+            parameters={
+                "q": ToolParameter(
+                    description="TraceQL query (e.g., '{resource.service.name=\"api\" && span.http.status_code=500}')",
+                    type="string",
+                    required=True,
+                ),
+                "limit": ToolParameter(
```
aantn, line 390: Should we set a default for this?

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                    description="Maximum number of traces to return",
+                    type="integer",
+                    required=False,
+                ),
+                "start": ToolParameter(
```
aantn, line 395: We only accept start/end time in timestamp format here, so this might not work when fetching traces from the time of a historical alert. (Fetch me all traces from the time of this alert.)

Could we add an eval for that please?

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+        try:
+            result = api.search_traces_by_query(
+                q=params["q"],
+                limit=params.get("limit"),
+                start=params.get("start"),
+                end=params.get("end"),
+                spss=params.get("spss"),
+            )
+            return StructuredToolResult(
+                status=ToolResultStatus.SUCCESS,
```
aantn, line 432: Please check for empty results and return ToolResultStatus.NO_DATA in that case. Its important so that the UI knows those results are empty.

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                params=params,
+            )
+        except Exception as e:
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=str(e),
+                params=params,
+            )
+
+    def get_parameterized_one_liner(self, params: Dict) -> str:
```
aantn, line 443: Please update to follow the standard convention:

Something like

```
    def get_parameterized_one_liner(self, params) -> str:
        return f"{toolset_name_for_one_liner(self.toolset.name)}: Search Traces ({q})"
```

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+        return f"{toolset_name_for_one_liner(self._toolset.name)}: Searched traces with TraceQL"
+
+
+class SearchTracesByTags(Tool):
```
aantn, line 447: Same comments on this toolset as the previous one. Not commenting 1:1 for concisness sake.

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+        try:
+            trace_data = api.query_trace_by_id_v2(
+                trace_id=params["trace_id"],
+                start=params.get("start"),
+                end=params.get("end"),
+            )
+
+            # Return raw trace data as YAML for readability
+            return StructuredToolResult(
+                status=ToolResultStatus.SUCCESS,
```
aantn, line 578: Same comment here on returning NO_DATA when appropriate

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            )
+        except Exception as e:
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=str(e),
+                params=params,
+            )
+
+    def get_parameterized_one_liner(self, params: Dict) -> str:
+        return f"{toolset_name_for_one_liner(self._toolset.name)}: Retrieved trace {params.get('trace_id')}"
```
aantn, line 590: To match the standard convention, please change from `<toolset name> Retrieved trace <trace id>` to `<toolset name> Fetch Trace (trace_id=...<trace id>)`

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+    def __init__(self, toolset: BaseGrafanaTempoToolset):
+        super().__init__(
+            name="search_tag_names",
+            description=(
+                "Discover available tag names across traces. "
+                "Uses the Tempo API endpoint: GET /api/v2/search/tags. "
+                "Returns tags organized by scope (resource, span, intrinsic)."
+            ),
+            parameters={
+                "scope": ToolParameter(
```
aantn, line 603: Consider converting this from a string parameter to an enum parameter, to make illegal values unrepresentable

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                    description="Start time in Unix epoch seconds",
+                    type="integer",
+                    required=False,
+                ),
+                "end": ToolParameter(
+                    description="End time in Unix epoch seconds",
+                    type="integer",
+                    required=False,
+                ),
+                "limit": ToolParameter(
```
aantn, line 625: Consider setting a default and communicating it in the description

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                    description="Maximum number of tag names to return",
+                    type="integer",
+                    required=False,
+                ),
+                "max_stale_values": ToolParameter(
```
aantn, line 630: Also consider setting a default

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            result = api.search_tag_names_v2(
+                scope=params.get("scope"),
+                q=params.get("q"),
+                start=params.get("start"),
+                end=params.get("end"),
+                limit=params.get("limit"),
+                max_stale_values=params.get("max_stale_values"),
+            )
+            return StructuredToolResult(
+                status=ToolResultStatus.SUCCESS,
```
aantn, line 656: Please set NO_DATA when appropriate

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            )
+        except Exception as e:
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=str(e),
+                params=params,
+            )
+
+    def get_parameterized_one_liner(self, params: Dict) -> str:
+        return f"{toolset_name_for_one_liner(self._toolset.name)}: Discovered tag names"
```
aantn, line 668: Here too, please change format to `<toolset name>: <2-3 word verb> (<params>)`

Meaning, please add the interesting params in parenthesis at the end and please change 'Discovered tag names' (past tense) to 'Search Tag Names'

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                    description="Start time in Unix epoch seconds",
+                    type="integer",
+                    required=False,
+                ),
+                "end": ToolParameter(
+                    description="End time in Unix epoch seconds",
+                    type="integer",
+                    required=False,
+                ),
+                "limit": ToolParameter(
```
aantn, line 703: Same comment on defaults here and for max_stale_values

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            )
+        except Exception as e:
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=str(e),
+                params=params,
+            )
+
+    def get_parameterized_one_liner(self, params: Dict) -> str:
+        return f"{toolset_name_for_one_liner(self._toolset.name)}: Retrieved values for tag '{params.get('tag')}'"
```
aantn, line 746: Here too, please update the format

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            result = api.search_tag_values_v2(
+                tag=params["tag"],
+                q=params.get("q"),
+                start=params.get("start"),
+                end=params.get("end"),
+                limit=params.get("limit"),
+                max_stale_values=params.get("max_stale_values"),
+            )
+            return StructuredToolResult(
+                status=ToolResultStatus.SUCCESS,
```
aantn, line 734: Please add NO_DATA when relevant

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+
+        try:
+            result = api.query_metrics_instant(
+                q=params["q"],
+                start=params.get("start"),
+                end=params.get("end"),
+                since=params.get("since"),
+            )
+            return StructuredToolResult(
+                status=ToolResultStatus.SUCCESS,
```
aantn, line 811: Could be NO_DATA

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+        except Exception as e:
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=str(e),
+                params=params,
+            )
+
+    def get_parameterized_one_liner(self, params: Dict) -> str:
+        return (
+            f"{toolset_name_for_one_liner(self._toolset.name)}: Computed TraceQL metric"
```
aantn, line 824: Please update

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                        "sum_over_time, max_over_time, min_over_time, avg_over_time, "
+                        "quantile_over_time, histogram_over_time, compare. "
+                        "Can use topk or bottomk modifiers. "
+                        "Syntax: {selector} | function(attribute) [by (grouping)]. "
+                        'Example: {resource.service.name="api"} | avg_over_time(duration)'
+                    ),
+                    type="string",
+                    required=True,
+                ),
+                "step": ToolParameter(
```
aantn, line 857: Please derive step from time range instead of letting LLM decide it - large values will have a catastrophic effect on the context window

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                    type="string",
+                    required=False,
+                ),
+                "end": ToolParameter(
+                    description="End time (Unix seconds/nanoseconds/RFC3339)",
+                    type="string",
+                    required=False,
+                ),
+                "since": ToolParameter(
+                    description="Duration string (e.g., '3h', '1d')",
```
aantn: The relationship between start/end and since is not clearly explained to the LLM. I believe they are mutually exclusive, but please confirm.

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+                    description="Start time (Unix seconds/nanoseconds/RFC3339)",
+                    type="string",
+                    required=False,
+                ),
+                "end": ToolParameter(
+                    description="End time (Unix seconds/nanoseconds/RFC3339)",
+                    type="string",
+                    required=False,
+                ),
+                "since": ToolParameter(
```
aantn: The relationship between start/end and since is not clearly explained to the LLM. I believe they are mutually exclusive, but please confirm.

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            result = api.query_metrics_range(
+                q=params["q"],
+                step=params.get("step"),
+                start=params.get("start"),
+                end=params.get("end"),
+                since=params.get("since"),
+                exemplars=params.get("exemplars"),
+            )
+            return StructuredToolResult(
+                status=ToolResultStatus.SUCCESS,
```
aantn, line 909: NO_DATA

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -553,6 +534,532 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            )
+        except Exception as e:
+            return StructuredToolResult(
+                status=ToolResultStatus.ERROR,
+                error=str(e),
+                params=params,
+            )
+
+    def get_parameterized_one_liner(self, params: Dict) -> str:
+        return f"{toolset_name_for_one_liner(self._toolset.name)}: Retrieved TraceQL metrics time series"
```
aantn, line 921: please update

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -562,9 +1069,16 @@ def __init__(self):
             docs_url="https://holmesgpt.dev/data-sources/builtin-toolsets/grafanatempo/",
             tools=[
                 FetchTracesSimpleComparison(self),
-                GetTempoTraces(self),
-                GetTempoTraceById(self),
-                GetTempoTags(self),
+                # GetTempoTraces(self),
```
aantn: Lets remove the old ones from the code altogether?

tempo_cli.py
```diff
@@ -0,0 +1,419 @@
+#!/usr/bin/env python3
+"""
```
aantn, line 2: Please move to scripts/

tempo_cli_README.md
```diff
@@ -0,0 +1,162 @@
+# Tempo CLI - Grafana Tempo Command Line Interface
```
aantn, line 1: Please move to scripts

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -15,43 +12,25 @@
...
     DEFAULT_TIME_SPAN_SECONDS,
 )
 from holmes.plugins.toolsets.utils import (
-    get_param_or_raise,
-    process_timestamps_to_int,
     toolset_name_for_one_liner,
+    process_timestamps_to_int,
 )
 
 TEMPO_LABELS_ADD_PREFIX = load_bool("TEMPO_LABELS_ADD_PREFIX", True)
```
aantn, line 31: When would this not be set?

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -67,6 +46,23 @@ def get_example_config(self):
...
+        try:
+            api = GrafanaTempoAPI(self.grafana_config, use_post=TEMPO_API_USE_POST)
+            if api.query_echo_endpoint():
+                return True, "Successfully connected to Tempo"
+            else:
+                return False, "Failed to connect to Tempo echo endpoint"
+        except Exception as e:
+            return False, f"Failed to connect to Tempo: {str(e)}"
+
     def build_k8s_filters(
```
aantn, line 67: Consider removing this altogether! Its currently only used by the comparative tool. I wonder if we really need it there or could get comparative to work without it.

aantn (dismissed):

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
Sheeproid: I"ve removed the 'since'

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
Sheeproid: I've removed the since

PR #943, opened 2025-09-04
add docs on robusta ai

aantn:


aantn:
Make sure we're also included everything from https://docs.robusta.dev/master/configuration/holmesgpt/getting-started.html - e.g. about when robusta token is in a secret

PR #946, opened 2025-09-07
ROB-2058

moshemorad:


holmes/config.py
```diff
@@ -170,20 +170,28 @@ def configure_robusta_ai_model(self) -> None:
...
                 return
 
-            for model in models:
+            for model in robusta_models.models:
                 logging.info(f"Loading Robusta AI model: {model}")
                 self._model_list[model] = {
+                    "name": model,
                     "base_url": f"{ROBUSTA_API_ENDPOINT}/llm/{model}",
                     "is_robusta_model": True,
+                    "model": "gpt-4o",  # Robusta AI model is using openai like API.
```
Avi-Robusta, line 186: I want to put a thumbs up here, this should solve the issue we had when he set the main model as sonnet in env var.
Github needs a like button for code.

holmes/config.py
```diff
@@ -193,6 +201,7 @@ def _load_default_robusta_config(self):
         if self._should_load_robusta_ai() and self.api_key:
             logging.info("Loading default Robusta AI model")
             self._model_list[ROBUSTA_AI_MODEL_NAME] = {
+                "name": ROBUSTA_AI_MODEL_NAME,
                 "base_url": ROBUSTA_API_ENDPOINT,
                 "is_robusta_model": True,
             }
```
Avi-Robusta, line 208: add `"model": "gpt-4o",` here too just to be safe

holmes/core/llm.py
```diff
@@ -72,14 +72,16 @@ def __init__(
...
+        tracer: Optional[Any] = None,
+        name: Optional[str] = None,
     ):
         self.model = model
         self.api_key = api_key
         self.api_base = api_base
         self.api_version = api_version
         self.args = args or {}
         self.tracer = tracer
+        self.name = name
```
Avi-Robusta, line 84: now that we have `name` param i would add it to the exception for completion so we know which llm had issues

tests/plugins/toolsets/test_internet.py
```diff
@@ -112,7 +112,12 @@ def test_html_to_markdown(fixture: Fixture):
     assert match, f"Values mismatch. Run the following command to compare expected with actual: `diff {fixture.expected_output_file_path} {actual_file_path_for_debugging}`"
 
 
-def test_fetch_webpage():
+def test_fetch_webpage(responses):
+    responses.get(
```
Avi-Robusta, line 116: whats this added for? testing web connectivity?

holmes/config.py
```diff
@@ -170,20 +170,28 @@ def configure_robusta_ai_model(self) -> None:
                 self._load_default_robusta_config()
                 return
 
-            models = fetch_robusta_models(
+            robusta_models = fetch_robusta_models(
                 self.account_id, self.session_token.get_secret_value()
             )
-            if not models:
+            if not robusta_models:
```
Avi-Robusta:
```suggestion
            if not robusta_models or not robusta_models.models:
```

holmes/config.py
```diff
@@ -193,6 +201,7 @@ def _load_default_robusta_config(self):
         if self._should_load_robusta_ai() and self.api_key:
```
Avi-Robusta, line 201: we need also `self._load_default_robusta_config = ROBUSTA_AI_MODEL_NAME` or something

Avi-Robusta (changes requested):
Looks good, two small changes

holmes/config.py
moshemorad: Good catch!

holmes/config.py
moshemorad: Not sure i understand can you please explain?

holmes/core/llm.py
moshemorad: 100%! 
Im goinig to refactor the shole llm model loading and usage on next pr so this will be done there.

tests/plugins/toolsets/test_internet.py
moshemorad: Added pytest-responses it is patching all requests by default and raise an error.
e.g. this test was actually doing real network call so i patched it. No reason for tests to do actual call unless there is a real reason behing it.

holmes/config.py
Avi-Robusta: sorry i ment 
self._default_robusta_model = ROBUSTA_AI_MODEL_NAME

Avi-Robusta (approved):

PR #955, opened 2025-09-10
Improve prometheus truncation

aantn:


holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -743,8 +825,31 @@ def _invoke(
...
                 if self.toolset.config.tool_calls_return_data:
-                    response_data["data"] = data.get("data")
+                    result_data = data.get("data", {})
+
+                    # Estimate the size of the data
+                    data_str_preview = json.dumps(result_data)
+                    data_size_chars = len(data_str_preview)
+
+                    # If data is too large (>80K chars), provide summary instead
+                    if data_size_chars > 80000:
```
aantn: Given the improvements in https://github.com/robusta-dev/holmesgpt/pulls?q=is%3Apr+is%3Aclosed this should take TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT into account. For this to work, the Prometheus truncation threshold must **always** be less than the general threshold

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -906,8 +1012,29 @@ def _invoke(
...
```
aantn: Same comment re. threshold

holmes/plugins/toolsets/prometheus/prometheus_instructions.jinja2
```diff
@@ -19,6 +19,31 @@
 * Only generate and execute a prometheus query after checking what metrics are available with the `list_available_metrics` tool
 * Check that any node, service, pod, container, app, namespace, etc. mentioned in the query exist in the kubernetes cluster before making a query. Use any appropriate kubectl tool(s) for this
 * The toolcall will return no data to you. That is expected. You MUST however ensure that the query is successful.
+
+## Handling High-Cardinality Metrics
+* CRITICAL: When querying metrics that may return many time series (>10), ALWAYS use aggregation to limit results
+* ALWAYS use `topk()` to limit the number of series returned
```
aantn: ALWAYS use `topk()` (or `bottomk()`)

holmes/plugins/toolsets/prometheus/prometheus.py
nherment: Ideally we use the same token counting mechanism as we use to truncate. This requires passing more context to the tool execution. It's not in place yet. What I've done instead is make the max chars configurable through the toolset config. It is set to 80'000 by default.

PR #956, opened 2025-09-10
feat: prevent tool calls responses that are too big

nherment:


holmes/common/env_vars.py
```diff
@@ -74,3 +74,10 @@ def load_bool(env_var, default: Optional[bool]) -> Optional[bool]:
 ENABLE_CLI_TOOL_APPROVAL = load_bool("ENABLE_CLI_TOOL_APPROVAL", True)
 
 MAX_GRAPH_POINTS = float(os.environ.get("MAX_GRAPH_POINTS", 300))
+
+# Limit each tool response to N% of the total context window.
+# Number between 0 and 100
+# Setting to either 0 or any number above 100 disables the logic that limits tool response size
+TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT = float(
```
moshemorad, line 81: @nherment can we move it to config instead of env_var? I think it will make more sense + will have built in validations using pydantic. Your call.

holmes/core/tools_utils/tool_context_window_limiter.py
```diff
@@ -0,0 +1,27 @@
+from holmes.common.env_vars import TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT
+from holmes.core.llm import LLM
+from holmes.core.tools import StructuredToolResultStatus
+from holmes.core.tools_utils.data_types import ToolCallResult
+
+
+def prevent_overly_big_tool_response(tool_call_result: ToolCallResult, llm: LLM):
+    if (
```
moshemorad, line 8: I think we should first make sure that the tool call result status is success right? no reason to do that if it is no_data/approval_request

Not sure about error thought WDYT?

holmes/core/tools_utils/tool_context_window_limiter.py
```diff
@@ -0,0 +1,27 @@
...
+        0 < TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT
+        and TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT <= 100
+    ):
+        message = tool_call_result.as_tool_call_message()
```
moshemorad, line 13: Thinking loud: Are we sure we want to count all the text and not only the content?

holmes/common/env_vars.py
```diff
@@ -74,3 +74,10 @@ def load_bool(env_var, default: Optional[bool]) -> Optional[bool]:
...
+    os.environ.get("TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT", 0)
```
moshemorad: I think that according to the ticket it should be 10% by default no?

holmes/core/tool_calling_llm.py
```diff
@@ -952,7 +890,9 @@ def call_stream(
 
                 for future in concurrent.futures.as_completed(futures):
                     tool_call_result: ToolCallResult = future.result()
-
+                    prevent_overly_big_tool_response(
```
moshemorad: Consider: move it into `_invoke_llm_tool_call` function.

holmes/common/env_vars.py
nherment: I agree, it should be in the config. However if I am to do this now I'll be adding a field to ToolCallingLLM and passing it this config value. It' would be just adding tech debt. We should instead look at a unified solution once we refactor the config.
This applies to all env vars and it's likely we want to consolidated all/most of them into the config.

holmes/core/tools_utils/tool_context_window_limiter.py
nherment: Yes that'd be better :+1:

holmes/core/tools_utils/tool_context_window_limiter.py
nherment: I think so. What matters is the space taken by the whole message and this is how we count tokens. Note that the token counting only takes the message.content if that's what you were referring to.

PR #962, opened 2025-09-12
ROB-2085: bash enabled by default, replaces most yaml toolsets

nherment:


helm/holmes/values.yaml
```diff
@@ -42,17 +42,7 @@ serviceAccount:
   imagePullSecrets: []
   annotations: {}
 
-toolsets:
-  kubernetes/core:
-    enabled: true
-  kubernetes/logs:
-    enabled: true
-  robusta:
```
moshemorad, line 50: I saw that robusta indeed enabled by default if dal was able to init.
Im not sure that this is the case with prometheus & internet toolset. am i missing something?

holmes/plugins/toolsets/bash/bash_toolset.py
```diff
@@ -22,120 +18,18 @@
...
 )
 from holmes.plugins.toolsets.bash.common.bash import execute_bash_command
-from holmes.plugins.toolsets.bash.common.config import BashExecutorConfig
-from holmes.plugins.toolsets.bash.kubectl.constants import SAFE_NAMESPACE_PATTERN
-from holmes.plugins.toolsets.bash.kubectl.kubectl_run import validate_image_and_commands
 from holmes.plugins.toolsets.bash.parse_command import make_command_safe
-from holmes.plugins.toolsets.utils import get_param_or_raise
 
 
 class BaseBashExecutorToolset(Toolset):
```
moshemorad, line 24: I see that there are 4 evals which failed and marked as easy. Looks like it is due to the bash toolset change. How we gonna fix it?

helm/holmes/values.yaml
nherment: fixed

holmes/plugins/toolsets/bash/bash_toolset.py
nherment: Are you referring to the LLM report in github?
A lot of these tests are now failing frquently. For the 'easy' llm failures, they have the same behaviour on this branch as on master.

I have pushed some of the mock files which should help in getting green tests.

PR #963, opened 2025-09-12
ensure token_count is added to the final message

nherment:


holmes/core/llm.py
```diff
@@ -530,3 +530,23 @@ def _create_model_entry(
             "is_robusta_model": is_robusta_model,
             "model": model,
         }
+
+
+def get_llm_usage(llm_response: Union[ModelResponse, CustomStreamWrapper]) -> dict:
```
aantn: Should this be `Union[ModelResponse, CustomStreamWrapper, TextCompletionResponse]`?

If I understand correctly, we're passing in `Union[ModelResponse, CustomStreamWrapper]` from tool_calling_llm.py, and when handling CustomStreamWrapper we do a recursive call that can also involve TextCompletionResponse.

Is that correct?

nherment: Yes that's correct. The linting error you caught is silenced by all the `# type: ignore` which are unfortunately needed. Updated the code.

PR #964, opened 2025-09-12
Build docker images on every commit

aantn:


.github/workflows/docker-dev-images.yaml
```diff
@@ -0,0 +1,56 @@
...
+      contents: 'read'
+      id-token: 'write'
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - uses: google-github-actions/auth@v2
+      with:
+        project_id: 'robusta-development'
+        workload_identity_provider: 'projects/429189597230/locations/global/workloadIdentityPools/github/providers/robusta-repos'
```
aantn: This still needs to be setup

PR #967, opened 2025-09-14
Improve Prometheus integration

aantn:
1. add support for querying metrics with regexes
2. add truncation warning for metadata apis (series/labels)

### DIFF
--- a/holmes/plugins/toolsets/prometheus/prometheus.py
+++ b/holmes/plugins/toolsets/prometheus/prometheus.py
@@ -46,7 +46,7 @@ class PrometheusConfig(BaseModel):
     prometheus_url: Optional[str]
     healthcheck: str = "-/healthy"
     # Setting to None will remove the time window from the request for labels
-    metrics_labels_time_window_hrs: Union[int, None] = 48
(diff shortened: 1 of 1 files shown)


holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -245,23 +245,41 @@ def fetch_metadata_with_series_api(
...
+
+    Returns:
+        Tuple[Dict, bool]: (metadata dict, is_truncated flag)
+    """
     url = urljoin(prometheus_url, "api/v1/series")
-    params: Dict = {"match[]": f'{{__name__=~".*{metric_name}.*"}}', "limit": "10000"}
+    limit = 100
+    # If pattern doesn't contain regex metacharacters, wrap with .* for backward compatibility
+    # Otherwise use as-is for full regex support
+    if not any(c in metric_name for c in r"^$.*+?{}[]|()\\"):
```
moshemorad: Consider: extract the check to func and reuse or even better all the pattern logic in the if else.

moshemorad (approved):

PR #974, opened 2025-09-14
ROB-2117 allow override context size with model list

RoiGlinik:


holmes/core/llm.py
```diff
@@ -523,12 +531,15 @@ def _create_model_entry(
...
         is_robusta_model: Optional[bool] = None,
+        args: Optional[dict[str, Any]] = None,
     ) -> dict[str, Any]:
+        args = args or {}
         return {
             "name": model_name,
             "base_url": base_url,
             "is_robusta_model": is_robusta_model,
             "model": model,
+            **args,
```
moshemorad: I think that this might be an issue, we are duplicating this dict on the `_get_llm` and than we need to pop all irrelevant keys since this might cause key error on some models like azure. I think that we need to keep it in single key and make sure to pop it before it is being used.

holmes/core/llm.py
```diff
@@ -424,7 +428,8 @@ def configure_robusta_ai_model(self) -> None:
 
             for model in robusta_models.models:
                 logging.info(f"Loading Robusta AI model: {model}")
-                self._llms[model] = self._create_robusta_model_entry(model)
+                args = robusta_models.models_args.get(model)
+                self._llms[model] = self._create_robusta_model_entry(model, args)
```
moshemorad, line 435: Since it is urgent, please open sub task on the ticket to add tests. Let's do them after the release. Thanks.

moshemorad (approved):

moshemorad (approved):

PR #976, opened 2025-09-14
Minor fixes - mostly to cli output

aantn:
Reduce spammy cli output + other minor fixes

### DIFF
--- a/holmes/core/llm.py
+++ b/holmes/core/llm.py
@@ -492,7 +492,7 @@ class LLMModelRegistry:
             )
 
         model_key, first_model_params = next(iter(self._llms.items()))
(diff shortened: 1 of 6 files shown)


holmes/core/llm.py
```diff
@@ -492,7 +492,7 @@ def get_model_params(self, model_key: Optional[str] = None) -> dict:
             )
 
         model_key, first_model_params = next(iter(self._llms.items()))
-        logging.info(f"Using first available model: {model_key}")
```
aantn, line 495: Note: this is a duplicate log (at least in cli path). We still print model elsewhere in cli

PR #978, opened 2025-09-16
ROB-2024: ignore text files when fetching issue data in RCA

nherment:


moshemorad:
Looks good.
Did we checked it with some exists finding and we are sure it isn't impacting the result?

holmes/common/env_vars.py
moshemorad, line 1: Looks good.
Did we checked it with some exists finding and we are sure it isn't impacting the result?

nherment: For investigation it's not for sure b/c we were not extracting zipped content. For the others, I don't think so. The limits are set higher than most evidence data.

PR #981, opened 2025-09-17
Improve cli experience by making output less spammy

aantn:


holmes/plugins/toolsets/investigator/core_investigation.py
```diff
@@ -57,22 +57,20 @@ def print_tasks_table(self, tasks):
...
-        logging.info(separator)
-        logging.info(header)
-        logging.info(separator)
+        tasks_to_display = []
 
         for task in tasks:
             status_display = f"{status_icons[task.status.value]} {task.status.value}"
             row = f"| {task.id:<{id_width}} | {task.content:<{content_width}} | {status_display:<{status_width}} |"
-            logging.info(row)
+            tasks_to_display.append(row)
```
aantn, line 72: Note: we're logging from multiple threads so sometimes I saw other logs being printed in the middle of the task list. To prevent this from happening I moved the logging to a single print

PR #984, opened 2025-09-19
ROB-2161 improved truncation methodology

nherment:


holmes/plugins/toolsets/newrelic/newrelic.py
```diff
@@ -76,14 +80,49 @@ def __init__(self, toolset: "NewRelicToolset"):
         )
         self._toolset = toolset
 
-    def format_metrics(
+    def compress_metrics_data(self, response: PromResponse) -> Optional[str]:
+        llm_data: Optional[str] = None
+        try:
+            if self._toolset.config and self._toolset.compress_metrics:
```
aantn: Why do we need to check the config here at all, if you're reading the setting from the toolset (and not the config)?

holmes/plugins/toolsets/newrelic/newrelic.py
```diff
@@ -76,14 +80,49 @@ def __init__(self, toolset: "NewRelicToolset"):
...
+                compression_ratio = (
+                    (1 - compressed_size / original_size) * 100
+                    if original_size > 0
+                    else 0
+                )
+
+                if compression_ratio > self._toolset.compress_metrics_minimum_ratio:
+                    # below this amount it's likely not worth mutating the response
+                    llm_data = compressed_data
+                    logging.info(
```
aantn: Please don't log at info level - it spams CLI output

holmes/plugins/toolsets/newrelic/newrelic.py
```diff
@@ -76,14 +80,49 @@ def __init__(self, toolset: "NewRelicToolset"):
...
+                        f"Compressed Newrelic metrics: {original_size:,} → {compressed_size:,} chars "
+                        f"({compression_ratio:.1f}% reduction)"
+                    )
+                else:
+                    logging.info(
```
aantn: Same here

holmes/core/tools_utils/tool_context_window_limiter.py
```diff
@@ -11,23 +11,24 @@ def prevent_overly_big_tool_response(tool_call_result: ToolCallResult, llm: LLM)
         and 0 < TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT
         and TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT <= 100
     ):
-        message = tool_call_result.as_tool_call_message()
-
-        messages_token = llm.count_tokens_for_message(messages=[message])
+        if not tool_call_result.size:
+            message = tool_call_result.as_tool_call_message()
+            messages_token = llm.count_tokens_for_message(messages=[message])
+            tool_call_result.size = messages_token
```
aantn, line 17: @nherment it looks from the code here that we're returning the original size to the UI but in fact I think we're returning the modified size. Is that correct? (Ideally we would return the modified size in the size field like today and have another field for original_size)

holmes/core/tool_calling_llm.py
```diff
@@ -688,6 +688,12 @@ def _get_tool_call_result(
             result=tool_response,
         )
 
+        message = tool_call_result.as_tool_call_message()
+
+        token_count = self.llm.count_tokens_for_message(messages=[message])
+        tool_call_result.size = token_count
```
aantn, line 694: Are calls to count tokens expensive? We seem to be doing it in several places. E.g. here and in prevent_overly_big_tool_response (or is this always called first in which case it never happens in prevent_overly_big_tool_response?)

holmes/plugins/toolsets/newrelic/newrelic.py
```diff
@@ -76,14 +80,49 @@ def __init__(self, toolset: "NewRelicToolset"):
         )
         self._toolset = toolset
 
-    def format_metrics(
+    def compress_metrics_data(self, response: PromResponse) -> Optional[str]:
```
aantn: The naming here is a big confusing. We have `raw_metric_to_compressed_metric`, `compress_metrics_data`, and `summarize_metrics`

holmes/plugins/toolsets/newrelic/newrelic.py
```diff
@@ -76,14 +80,49 @@ def __init__(self, toolset: "NewRelicToolset"):
...
+                    logging.info(
+                        f"Compressed Newrelic metrics: {original_size:,} → {compressed_size:,} chars "
+                        f"({compression_ratio:.1f}% reduction). Original data will be used instead."
+                    )
+        except Exception:
+            logging.warning("Failed to compress newrelic data", exc_info=True)
+
+        return llm_data
+
+    def to_new_relic_records(
```
aantn: Name is confusing here too. Isn't this to Prometheus format, not to NR format?

holmes/plugins/toolsets/prometheus/data_compression.py
```diff
@@ -0,0 +1,348 @@
...
+
+def format_labels(
+    labels: set[tuple[str, Any]], section_name: str, line_prefix: str = ""
+) -> list[str]:
+    lines = []
+    if labels:
+        sorted_labels = sorted(
+            labels, key=lambda t: t[0]
+        )  # keep label list stable in the outpout by sorting them by key
+        if len(sorted_labels) <= 10:
```
aantn, line 30: Why do we need the two formats here?

holmes/plugins/toolsets/prometheus/data_compression.py
```diff
@@ -0,0 +1,348 @@
...
+        else:
+            lines.append(line_prefix + section_name)
+            for k, v in sorted_labels:
+                lines.append(f"{line_prefix}{INDENT_SPACES}- {str(k)}: {str(v)}")
+    else:
+        raise Exception("No label")
+    return lines
+
+
+INDENT_SPACES = "  "
```
aantn: Please put this before the function where it is used

holmes/core/models.py
```diff
@@ -68,6 +68,25 @@ def as_streaming_tool_result_response(self):
 
 def format_tool_result_data(tool_result: StructuredToolResult) -> str:
     tool_response = tool_result.data
+    if tool_result.llm_data:
```
aantn, line 71: Are there more elegant ways we could handle this? i.e. why can't the implementation on this just be `return tool_result.llm_data if tool_result.llm_data else tool_result.data`?

I'm having a little trouble tracking what goes where (as_tool_call_message go to the LLM I assume? do as_tool_result_response and as_streaming_tool_result_response go to the UI?)

holmes/plugins/toolsets/prometheus/data_compression.py
```diff
@@ -0,0 +1,348 @@
...
+
+
+def find_most_common_label(
+    metrics: list[CompressedMetric], ignore_label_set: Set[tuple[str, Any]]
+) -> tuple[Optional[tuple[str, Any]], int]:
+    """
+    Find label keys and values that are most common across all label sets.
+    Returns labels that appear in ALL label sets with the same value.
+
+    Args:
```
aantn, line 306: Args is wrong

holmes/plugins/toolsets/prometheus/data_compression.py
```diff
@@ -0,0 +1,348 @@
...
+        summarized_text += format_compressed_metrics(metrics)
+
+    return summarized_text
+
+
+def find_most_common_label(
+    metrics: list[CompressedMetric], ignore_label_set: Set[tuple[str, Any]]
+) -> tuple[Optional[tuple[str, Any]], int]:
+    """
+    Find label keys and values that are most common across all label sets.
```
aantn: Docs seem to be wrong. They imply you return multiple labels, but in fact only a single label is returned. Same applies to comments in the function

holmes/plugins/toolsets/prometheus/data_compression.py
```diff
@@ -0,0 +1,348 @@
...
+        return list(obj)
+    raise TypeError
+
+
+def summarize_metrics(metrics_to_process: list[CompressedMetric]) -> str:
+    summarized_text = ""
+    filtered_metrics = pre_filter_metrics(metrics=metrics_to_process)
+    if (
+        len(filtered_metrics.metrics_with_only_zero_values)
+        >= len(filtered_metrics.other_metrics) * 0.1
```
aantn: what if the len condition is false? It seems like you drop the time series entirely from the output!

holmes/plugins/toolsets/prometheus/data_compression.py
```diff
@@ -0,0 +1,348 @@
...
+                    lines.append(metric_lines)
+        txt = "\n".join([line for line in lines])
+        return txt
+    else:
+        raise Exception("Data has no metrics and is not a CompressedMetric")
+
+
+def format_zero_values_metrics(metrics: list[Union[Group, CompressedMetric]]) -> str:
+    formatted_string = (
+        "# Metrics with the following hierarchised labels have NO DATA:\n"
```
aantn: Comment could be misleading for LLM. They have data, the data is zero.

holmes/core/models.py
nherment: We can't just return llm_data because it breaks the graphs in the UI.

`tool_result.data` contains the `random_key` field which is necessary to link the embedded tags `<< ... random_key: .. >>` to the tool call data.

holmes/core/tool_calling_llm.py
nherment: This should be called first

holmes/core/tools_utils/tool_context_window_limiter.py
nherment: We're counting the tokens the way the LLM sees it. `tool_call_result.as_tool_call_message()` formats the message for the LLM and we're counting the same message.

holmes/plugins/toolsets/newrelic/newrelic.py
nherment: I'll rename them

holmes/plugins/toolsets/newrelic/newrelic.py
nherment: Indeed this is not necessary.

holmes/core/models.py
aantn: @nherment would it be simpler if we updated the ui code?

holmes/plugins/toolsets/prometheus/data_compression.py
nherment: The intention was to make it easier to read

holmes/plugins/toolsets/prometheus/data_compression.py
nherment: If there are no zero values metrics then the zero value text is not added to the summary. The metrics with values are still added to the summary later in this function

aantn: @nherment can you explain? The conditional doesn't check if there are any zero valued metrics. It checks if they are more than 10%. What if they are only 5%?

nherment: Ah indeed. Fixed.

holmes/core/models.py
nherment: A good solution is to improve the protocol for LLM messages and add the necessary fields to `StructuredToolResult`. It would mean changing both the UI and holmes code.

It would not be enough to update the UI code. This is an issue of giving Holmes access to the unique tool call key.

aantn: I'm in favor. Could we give every tool result a random_key and inject it into what the LLM sees? We're going to be adding a lot of new embeds on the frontend soon, and I want to make sure we can do so without updating Holmes for each one.

holmes/core/tools_utils/tool_context_window_limiter.py
aantn: @nherment don't we handle truncation below on line 23? So we count tokens here before truncation happens, which impacts what the LLM sees. Woudn't this token count be wrong in case of truncation?

nherment: Good catch, I set it to None after the limiter has kicked in. I don't think the UI is using that number.

aantn: @nherment wont that throw off the token counts? Why not included the size of the truncation message which is what the llm actually sees?

PR #987, opened 2025-09-23
ROB-2136 link tool limit to global truncation limit

nherment:


holmes/core/tools_utils/tool_context_window_limiter.py
```diff
@@ -5,19 +5,24 @@
...
         and TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT <= 100
     ):
-        message = tool_call_result.as_tool_call_message()
+        return int(context_window_size * TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT // 100)
+    else:
+        return context_window_size
+
 
+def prevent_overly_big_tool_response(tool_call_result: ToolCallResult, llm: LLM):
+    if tool_call_result.result.status == StructuredToolResultStatus.SUCCESS:
```
aantn: Do we really want to limit this to `StructuredToolResultStatus.SUCCESS`? We should protect context window regardless of error/failure, no?

holmes/core/tools_utils/tool_executor.py
```diff
@@ -9,14 +9,17 @@
...
     ToolsetStatusEnum,
+    ToolInvokeContext,
 )
+from holmes.core.llm import LLM
 from holmes.core.tools_utils.toolset_utils import filter_out_default_logging_toolset
 
 
 class ToolExecutor:
-    def __init__(self, toolsets: List[Toolset]):
+    def __init__(self, toolsets: List[Toolset], llm: Optional[LLM] = None):
```
aantn: Any specific reason to make it optional not mandatory?

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -74,6 +78,57 @@ def logger_name(self) -> str:
         return ""
 
 
+# Approximate number of character per LLM token.
+# There is less chances the auto-truncation will remove too many characters if this number is optimistic (i.e. lower than actual) rather than pessimistic
+APPROXIMATE_CHAR_TO_TOKEN_RATIO = 4
```
aantn: Can you place constants at top of file?

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -74,6 +78,57 @@ def logger_name(self) -> str:
...
+
+TRUNCATION_PROMPT_PREFIX = "[... PREVIOUS LOGS ABOVE THIS LINE HAVE BEEN TRUNCATED]\n"
```
aantn: I assume in the LLM's training data it's mostly seen logs where N lines are fetched, but very few examples where an individual log line is cut-off at the start. Whereas we do truncate log lines halfway through the line. Consider either updating the truncation prompt to be more clear about this so the LLM understands, or counting tokens on a per-line basis and only including complete lines in output.

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -74,6 +78,57 @@ def logger_name(self) -> str:
...
+            MIN_NUMBER_OF_CHARACTERS_TO_TRUNCATE, number_of_characters_to_truncate
+        )
+
+        if len(text) <= number_of_characters_to_truncate:
+            logging.warning(
+                f"The calculated token count for logs is {token_count} (max allowed tokens={token_limit}) but the logs are only {len(text)} characters which is below the intended truncation of {number_of_characters_to_truncate} characters. Logs will no longer be truncated"
+            )
+            return
+        else:
+            text = TRUNCATION_PROMPT_PREFIX + text[number_of_characters_to_truncate:]
```
aantn: This looks like a bug. You're appending TRUNCATION_PROMPT_PREFIX each iteration of the loop.

holmes/core/tools_utils/tool_context_window_limiter.py
nherment: I think success only is fine. We haven't had errors that take a lot of tokens and I don't think we should prevent errors from reaching the LLM. There should be other mechanisms in place to trim the input tokens so that errors/no data tool call results still reach the LLM.

holmes/core/tools_utils/tool_executor.py
nherment: It can be removed. It also seems the invoke function of ToolExecutor is not used anywhere. Added a TODO to remove it.

holmes/plugins/toolsets/logging_utils/logging_api.py
nherment: The logic now tries to cut at the next line feed

holmes/plugins/toolsets/logging_utils/logging_api.py
nherment: Comment added to clarify the logic

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -74,6 +87,64 @@ def logger_name(self) -> str:
...
 
 
+def truncate_logs(
+    logging_structured_tool_result: StructuredToolResult, llm: LLM, token_limit: int
+):
+    token_count = count_tool_response_tokens(
+        llm=llm, structured_tool_result=logging_structured_tool_result
+    )
+    text = None
+    logging.info(f"token_count={token_count}, token_limit={token_limit}")
```
aantn: Please change to debug level to avoid spamming CLI output

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -74,6 +87,64 @@ def logger_name(self) -> str:
...
+            return
+        ratio = token_count / token_limit
+        character_count = len(text)
+        number_of_characters_to_truncate = character_count - ceil(
+            character_count / ratio
+        )
+        number_of_characters_to_truncate = max(
+            MIN_NUMBER_OF_CHARACTERS_TO_TRUNCATE, number_of_characters_to_truncate
+        )
+        logging.info(
```
aantn: Please change to debug level to avoid spamming CLI output

holmes/plugins/toolsets/logging_utils/logging_api.py
aantn: And can we please replace with a single INFO or WARNING level log that is shown ONLY when truncation happens, and includes more information in a user-friendly format. e.g.

Warning: Tool XYZ returned too many logs. Truncating from X tokens to Y tokens.

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -22,6 +26,15 @@
 
 POD_LOGGING_TOOL_NAME = "fetch_pod_logs"
 
+# Approximate number of character per LLM token.
+# There is less chances the auto-truncation will remove too many characters if this number is optimistic (i.e. lower than actual) rather than pessimistic
+APPROXIMATE_CHAR_TO_TOKEN_RATIO = 4
```
aantn: This is never used.

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -91,8 +92,8 @@ class PrometheusConfig(BaseModel):
     rules_cache_duration_seconds: Optional[int] = 1800  # 30 minutes
     additional_labels: Optional[Dict[str, str]] = None
     prometheus_ssl_enabled: bool = True
-    query_response_size_limit: Optional[int] = (
-        DEFAULT_QUERY_RESPONSE_SIZE_LIMIT  # Limit the max number of characters in a query result to proactively prevent token limit issues (roughly 5-6k tokens)
+    query_response_size_limit_pct: Optional[int] = (
```
aantn, line 95: Please set a default value equivalent of 2% (I believe this is equivalent to the previously used threshold.)

aantn: (I think we're currently using a default threshold of 15% when this is not set and we fallback to the global per-tool-call limit, right?)

nherment: correct

PR #989, opened 2025-09-24
add initial feedback command

mainred:
When feedback callback is not None, we'll show the /feedback command though completer and /help command.
The feedback includes both the user feedback and metadata. metadata contains a shortcut of llm and list of ask/response until now, and more metadata is going to be added.

```
User: /feedback
Was this response useful to you? 👍(y)/👎(n): y
Do you want to provide any additional comments for feedback? (press Enter to skip): not bad
✓ Feedback recorded (rating=👍, "not bad")
```

### DIFF
--- a/holmes/core/feedback.py
+++ b/holmes/core/feedback.py
@@ -1,0 +1,176 @@
+from typing import Callable, Optional
+from .llm import LLM
+
+
+class FeedbackLLM:
+    """Class to represent a LLM in the feedback."""
+
+    def __init__(self, model: str, max_context_size: int):
+        self.model = model
+        self.max_context_size = max_context_size
+
+    def update_from_llm(self, llm: LLM):
+        self.model = llm.model
+        self.max_context_size = llm.get_context_window_size()
+
+    def to_dict(self) -> dict:
+        """Convert to dictionary representation."""
+        return {
+            'model': self.model,
+            'max_context_size': self.max_context_size
+        }
+
+
+# TODO: extend the FeedbackLLMResponse to include each tool call results details used for evaluate the overall response.
+# Currenlty tool call details in plan:
+# - toolcall parameter and success/failure, toolcall truncation size
+# - Holmes plan (todo list)
+# - Holmes intermediate output
+class FeedbackLLMResponse:
+    """Class to represent a LLM response in the feedback"""
+
+    def __init__(self, user_ask: str, response: str):
+        self.user_ask = user_ask
+        self.response = response
+
+    def to_dict(self) -> dict:
+        """Convert to dictionary representation."""
+        return {
+            'user_ask': self.user_ask,
+            'response': self.response
+        }
+
+
+class FeedbackMetadata:
+    """Class to store feedback metadata."""
+
+    def __init__(self, llm_responses: Optional[list[FeedbackLLMResponse]] = None):
+        # In iteration mode, there can be multiple ask and response pairs.
+        self.llm_responses = llm_responses if llm_responses is not None else []
+        self.llm = FeedbackLLM("", 0)
+
+    def add_llm_response(self, user_ask: str, response: str) -> None:
+        """Add a LLM response to the metadata."""
+        llm_response = FeedbackLLMResponse(user_ask, response)
+        self.llm_responses.append(llm_response)
+
+    def update_llm(self, llm: LLM) -> None:
+        """Update the LLM information in the metadata."""
+        self.llm.update_from_llm(llm)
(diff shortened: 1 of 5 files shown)


holmes/core/feedback.py
```diff
@@ -0,0 +1,175 @@
+from typing import Callable, Optional
```
mainred, line 2: Instead of extending the holmes/interactive.py which is already pretty heavy, I used a feedback.py to simply the test.
We can consider moving more functionalities out of interactive.py

holmes/interactive.py
```diff
@@ -62,19 +64,25 @@ class SlashCommands(Enum):
...
     SHOW = ("/show", "Show specific tool output in scrollable view")
+    FEEDBACK = ("/feedback", "Provide feedback on the agent's response")
 
     def __init__(self, command, description):
         self.command = command
         self.description = description
 
 
-SLASH_COMMANDS_REFERENCE = {cmd.command: cmd.description for cmd in SlashCommands}
-ALL_SLASH_COMMANDS = [cmd.command for cmd in SlashCommands]
```
mainred, line 72: The two constants can be scoped in SlashCommandCompleter, and their work cross.
We can only use SlashCommandCompleter as the only source of truth of what slashcommands are supported.

holmes/interactive.py
```diff
@@ -62,19 +64,25 @@ class SlashCommands(Enum):
...
-
-
 class SlashCommandCompleter(Completer):
-    def __init__(self):
-        self.commands = SLASH_COMMANDS_REFERENCE
+    def __init__(self, unsupported_commands: Optional[List[str]] = None):
```
mainred, line 75: unsupported_commands is not a common requirement, and we can remove it once we build feedback callback in holmesgpt

holmes/core/feedback.py
```diff
@@ -0,0 +1,185 @@
...
+        self.max_context_size = llm.get_context_window_size()
+
+    def to_dict(self) -> dict:
+        """Convert to dictionary representation."""
+        return self.__dict__
+
+
+# TODO: extend the FeedbackLLMResponse to include each tool call results details used for evaluate the overall response.
+# Currenlty tool call details in plan:
+# - toolcall parameter and success/failure, toolcall truncation size
```
mainred, line 34: For these tool details, I am considering making the tool call related info more extensible to not only be used in feedback not also in toolset metrics

holmes/main.py
```diff
@@ -1,4 +1,33 @@
 # ruff: noqa: E402
+from holmes.utils.file_utils import write_json_file
```
moshemorad: Please move back down, this might break the certification initialization.

holmes/interactive.py
```diff
@@ -1002,13 +1080,20 @@ def get_bottom_toolbar():
                     console.print(
                         f"[bold {HELP_COLOR}]Available commands:[/bold {HELP_COLOR}]"
                     )
-                    for cmd, description in SLASH_COMMANDS_REFERENCE.items():
+                    for cmd, description in slash_completer.commands.items():
+                        # Only show feedback command if callback is available
+                        if (
```
moshemorad, line 1085: aren't we already remove the feedback command if callback isn't provided?

holmes/interactive.py
```diff
@@ -1052,6 +1137,14 @@ def get_bottom_toolbar():
...
                         continue  # User chose not to share or no output, continue to next input
                     user_input = shared_input
+                elif (
+                    command == SlashCommands.FEEDBACK.command
+                    and feedback_callback is not None
+                ):
+                    user_feedback = handle_feedback_command(style, console)
+                    feedback.user_feedback = user_feedback
+                    feedback_callback(feedback)
+                    return
```
moshemorad, line 1147: We want to finish the interactive mode if user provided feedback?

holmes/interactive.py
mainred: This is by design. Microsoft will introduce the callback before Robusta team, so before  we implement the callback function in holmegpt, we hide the /feedback command from holmesgpt user to unblock Microsoft users from using the /feedback command.  As synced with  @aantn

holmes/interactive.py
mainred: We can discuss. Usually, the session ends after the feedback.  I don't have a strong opinion here.

moshemorad: We can start that way, just wanted to make sure it is by design.

moshemorad (approved):

PR #990, opened 2025-09-25
Fix MCP toolset prerequisites check

feiskyer:
There are high chances of empty MCP tools when using remote MCP server.

This PR ensures MCP servers reload their tools even when previously failed by including MCP toolsets in prerequisites check regardless of status.


mainred (commented):
Currently refresh toolsets happens when the first the user runs holmesgpt or specify refresh-toolsets, and mcp server is reachable, its status will be cached.
I think your case is mcp server is not reachable the first time we run hollmesgpt?

feiskyer:
per my local test, the mcp endpoint is always reachable, but holmes failed to fetch its tool list. That's why I have this PR.
And holmes should retry here as well when last refresh has failed.

holmes/core/toolset_manager.py
```diff
@@ -275,7 +275,11 @@ def load_toolset_with_status(
                 toolset.path = cached_status.get("path", None)
             # check prerequisites for only enabled toolset when the toolset is loaded from cache. When the toolset is
             # not loaded from cache, the prerequisites are checked in the refresh_toolset_status method.
-            if toolset.enabled and toolset.status == ToolsetStatusEnum.ENABLED:
+            if toolset.enabled and (
+                toolset.status == ToolsetStatusEnum.ENABLED
+                or toolset.type == ToolsetType.MCP
```
mainred, line 280: We cache mcp server status as well, if refreshing the toolset the mcp server is enabled and work, it's enabled status will be cached.
But as an exception, mcp server is different from the other toolsets which once enabled, we want it's prerequisites are check and then tools are listed all the time.

mainred (approved):

holmes/core/toolset_manager.py
mainred: cc @moshemorad

mainred (approved):

PR #994, opened 2025-09-25
mute unncessary info log for console

mainred:
Make config.dal to optional since dal is passing down to tool_executor and to the tool. This protect the case when we don't use Dal is CLI.

### DIFF
--- a/holmes/config.py
+++ b/holmes/config.py
@@ -9,11 +9,8 @@ import sentry_sdk
 import yaml  # type: ignore
 from pydantic import BaseModel, ConfigDict, FilePath, PrivateAttr, SecretStr
 
-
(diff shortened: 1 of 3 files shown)


holmes/config.py
```diff
@@ -117,10 +114,10 @@ def toolset_manager(self) -> ToolsetManager:
         return self._toolset_manager
 
     @property
-    def dal(self) -> SupabaseDal:
+    def dal(self) -> Optional[SupabaseDal]:
         if not self._dal:
             self._dal = SupabaseDal(self.cluster_name)  # type: ignore
-        return self._dal
+        return self._dal if self._dal else None
```
moshemorad: Not sure i understand in which case it will be none. can you please explain?

PR #1006, opened 2025-09-28
ROB-2197 Allow passing custom sys prompt param in Holmes chat endpoint

mershal:


holmes/core/conversations.py
```diff
@@ -262,7 +262,7 @@ def build_issue_chat_messages(
 
 
 def add_or_update_system_prompt(
-    conversation_history: List[Dict[str, str]], ai: ToolCallingLLM, config: Config
+    conversation_history: List[Dict[str, str]], ai: ToolCallingLLM, config: Config, custom_system_prompt: Optional[str] = None
```
aantn: Possible better to call it `extra_system_prompt` or `additional_system_prompt`, to make it clear you're not overwriting the whole system prompt.

holmes/core/models.py
```diff
@@ -230,6 +230,7 @@ class WorkloadHealthRequest(BaseModel):
 
 class ChatRequest(ChatRequestBaseModel):
     ask: str
+    custom_system_prompt: Optional[str] = None
```
aantn: Consider moving into ChatRequestBaseModel

server.py
```diff
@@ -332,6 +332,7 @@ def chat(chat_request: ChatRequest):
             ai=ai,
             config=config,
             global_instructions=global_instructions,
+            custom_system_prompt=chat_request.custom_system_prompt
```
aantn: Consider also adding to additional functions:
* `investigate_issues`
* `stream_investigate_issues`
* `issue_conversation`

That way it will be available for RCA as well + follow-up questions on RCA

aantn (dismissed):

Avi-Robusta (approved):

PR #1009, opened 2025-09-29
Fix list of latest benchmarks

aantn:
By using an mkdocs plugin to generate dynamically instead of relying on hardcoded markdown links. This also requires changing navigation across the docs due to the new plugin


aantn:
Closing, replaced by https://github.com/robusta-dev/holmesgpt/pull/1017 which includes all changes here.

PR #1014, opened 2025-09-29
ROB-2125 conversation history compaction

nherment:


holmes/plugins/toolsets/investigator/core_investigation.py
```diff
@@ -15,9 +18,44 @@
...
+                id=todo_item.get("id", str(uuid4())),
+                content=todo_item.get("content", ""),
+                status=TaskStatus(todo_item.get("status", "pending")),
+            )
+            tasks.append(task)
+
+    return tasks
+
+
+def get_tasks_from_potential_openai_tool_call(
```
moshemorad: Not in use? am i missing something?

holmes/core/llm.py
```diff
@@ -296,11 +305,14 @@ def count_tokens(
...
             elif role == "tool":
                 tools_tokens += token_count
+            elif role == "assistant":
+                assistant_tokens += token_count
             else:
                 # although this should not be needed,
                 # it is defensive code so that all tokens are accounted for
                 # and can potentially make debugging easier
                 other_tokens += token_count
+                print(json.dumps(message))
```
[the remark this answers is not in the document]
moshemorad: Please remove.

holmes/core/llm.py
```diff
@@ -41,6 +41,13 @@
 
 OVERRIDE_MAX_OUTPUT_TOKEN = environ_get_safe_int("OVERRIDE_MAX_OUTPUT_TOKEN")
 OVERRIDE_MAX_CONTENT_SIZE = environ_get_safe_int("OVERRIDE_MAX_CONTENT_SIZE")
+
+
+def get_context_window_compaction_threshold_pct() -> int:
+    """Get the compaction threshold percentage at runtime to support test overrides."""
+    return environ_get_safe_int("CONTEXT_WINDOW_COMPACTION_THRESHOLD_PCT", default="80")
```
moshemorad: why 80% are we sure that it isn't too early? 

also any good reason it is function that parse it every call and not like the above vars?

holmes/core/tools_utils/tool_context_window_limiter.py
```diff
@@ -28,37 +34,46 @@ def get_max_token_count_for_single_tool(llm: LLM) -> int:
...
+        and 0 < TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT
+        and TOOL_MAX_ALLOCATED_CONTEXT_WINDOW_PCT <= 100
+    ):
+        message = tool_call_result.as_tool_call_message()
 
-    if messages_token > max_tokens_allowed:
-        relative_pct = ((messages_token - max_tokens_allowed) / messages_token) * 100
+        tokens = llm.count_tokens(messages=[message])
+        context_window_size = llm.get_context_window_size()
+        max_tokens_allowed: int = int(
```
moshemorad: WDYT about moving it to llm as property and reuse everywhere?

holmes/core/truncation/compaction.py
```diff
@@ -0,0 +1,67 @@
+import logging
+from typing import Optional
+from holmes.common.env_vars import (
+    ENABLE_CONVERSATION_HISTORY_COMPACTION,
+)
+from holmes.core.llm import LLM
+from holmes.plugins.prompts import load_and_render_prompt
+from litellm.types.utils import ModelResponse
+
+TOKENS_RESERVED_FOR_SUMMARIZATION_PROMPT = 500
```
moshemorad: Unused?

holmes/core/tool_calling_llm.py
```diff
@@ -420,40 +421,63 @@ def call(  # type: ignore
...
             maximum_output_token = self.llm.get_maximum_output_token()
-            perf_timing.measure("count tokens")
+            # Try conversation history compaction first if approaching context window limit
+            if (initial_tokens.total_tokens + maximum_output_token) > (
+                max_context_size * get_context_window_compaction_threshold_pct() / 100
+            ):
+                compacted_messages = compact_conversation_history(
+                    original_conversation_history=messages, llm=self.llm
+                )
+                compacted_tokens = self.llm.count_tokens(
```
moshemorad: If the compaction env var is set to false this flow still will be run right? feels very unintoative. Can we please refactor it and make sure that it is only run if it is enabled?

holmes/core/tool_calling_llm.py
```diff
@@ -905,40 +925,70 @@ def call_stream(
...
 
             tools = None if i == max_steps else tools
             tool_choice = "auto" if tools else None
 
-            tokens = self.llm.count_tokens(messages=messages, tools=tools)  # type: ignore
+            initial_tokens = self.llm.count_tokens(messages=messages, tools=tools)  # type: ignore
             max_context_size = self.llm.get_context_window_size()
             maximum_output_token = self.llm.get_maximum_output_token()
-            perf_timing.measure("count tokens")
+            if (initial_tokens.total_tokens + maximum_output_token) > (
```
moshemorad: Can we please extract and reuse it in both of the `call` functions?

holmes/core/tool_calling_llm.py
```diff
@@ -905,40 +925,70 @@ def call_stream(
...
+                compacted_messages = compact_conversation_history(
+                    original_conversation_history=messages, llm=self.llm
+                )
+                compacted_tokens = self.llm.count_tokens(
+                    compacted_messages, tools=tools
+                )
+                compacted_total_tokens = compacted_tokens.total_tokens
+
+                if compacted_total_tokens < initial_tokens.total_tokens:
+                    messages = compacted_messages
```
moshemorad: Just to make sure im getting the hand of it. If it happens in the middle of the loop the llm will not get the previous messages and more from now on but only the compaction and new messages right?

moshemorad: Follow up question, i ran it and saw that in the UI when having follow up question, when compaction accourd we send the history message only from the compaction on, is that right?

holmes/core/tool_calling_llm.py
```diff
@@ -905,40 +925,70 @@ def call_stream(
...
+                    compaction_message = f"The conversation history has been compacted from {initial_tokens.total_tokens} to {compacted_total_tokens} tokens"
+                    logging.info(compaction_message)
+                    yield StreamMessage(
```
moshemorad: What the UI is doing with those events? are they familiar with them? we need to open sub tasks for follow ups?

holmes/core/tool_calling_llm.py
```diff
@@ -905,40 +925,70 @@ def call_stream(
...
+                        data={
+                            "content": compaction_message,
+                            "messages": compacted_messages,
+                            "metadata": {
+                                "initial_tokens": initial_tokens.total_tokens,
+                                "compacted_tokens": compacted_total_tokens,
+                            },
+                        },
+                    )
+                    yield StreamMessage(
```
moshemorad: why we need to have this ai_message? is it containing relevant data?

holmes/core/tool_calling_llm.py
moshemorad: Another question, since we are resetting the messages to the compaction message this means that all the tool calls are gone. What if the LLM would like to run a tool call which already ran? it won't let him right? Is that intended?

tests/llm/fixtures/test_ask_holmes/161_conversation_compaction/test_case.yaml
```diff
@@ -0,0 +1,25 @@
+before_test: |
```
moshemorad, line 1: I think we need to might consider add here couple of more evals.
1. New eval category for compaction testing and make sure the summarization is good enough.
2. Eval to check compaction in middle of the iteration loop with more complex scenario
3. Is this eval is good enough? it doesn't contain any tool calls and feel like it is not close enough to real life scenarios. Can we think about couple more scenarios that we would like to test here?

holmes/core/llm.py
nherment: Happy to change that number, it is somewhat arbitrary.
When we get close to the context window limit (esp. if it's big) the LLM's efficacy will be reduced. Compacting early could help the LLM in maintaining focus on the last tool calls or user's questions.

holmes/core/tool_calling_llm.py
nherment: > If it happens in the middle of the loop the llm will not get the previous messages and more from now on but only the compaction and new messages right?

Yes

> when compaction accourd we send the history message only from the compaction on, is that right?

Yes. Compaction is lossy. It consists in rewriting the conversation and then this compacted version becomes the new conversation history going forward. The original conversation is lost.

> What if the LLM would like to run a tool call which already ran? it won't let him right? Is that intended?

Compaction removes the historical tool calls. Therefore the LLM will be allowed to repeat previous tool calls. Thinking about it, it's true that in theory the LLM will then be able to repeat tool calls again.

holmes/core/tool_calling_llm.py
nherment: Right now the UI will not react to it but Natan was thinking of implementing support for this.

holmes/core/tool_calling_llm.py
nherment: This is implemented because Natan requested that users be notified that the conversation history is compacted.

tests/llm/fixtures/test_ask_holmes/161_conversation_compaction/test_case.yaml
nherment: Added some evals

holmes/core/llm.py
moshemorad: Thinking here loud, 20% out of 200K is 40K which is a lot. 
If the remining tool calls and message were able to fit inside it, do you think that the compacation cause a better result or might effect the actual result quaility?
My gut feelling is that original data before compacation is better but honestly i don't know that.
If we don't have a clear understanding maybe it worth to get it back to 100% (around 95%) and open a ticket to research it. WDYT?

holmes/core/tool_calling_llm.py
```diff
@@ -421,40 +312,68 @@ def call(  # type: ignore
...
             tool_choice = "auto" if tools else None
 
-            tokens = self.llm.count_tokens(messages=messages, tools=tools)
+            initial_tokens = self.llm.count_tokens(messages=messages, tools=tools)
+
             max_context_size = self.llm.get_context_window_size()
             maximum_output_token = self.llm.get_maximum_output_token()
-            perf_timing.measure("count tokens")
+            # Try conversation history compaction first if approaching context window limit
+            if ENABLE_CONVERSATION_HISTORY_COMPACTION and (
```
moshemorad: Shouldn't this also use the limiter logic?

holmes/core/tool_calling_llm.py
```diff
@@ -916,42 +817,29 @@ def call_stream(
...
-                    messages, max_context_size, maximum_output_token
-                )
-                metadata["truncations"] = [
-                    t.model_dump() for t in truncated_res.truncations
-                ]
-                messages = truncated_res.truncated_messages
-                perf_timing.measure("truncate_messages_to_fit_context")
-            else:
-                metadata["truncations"] = []
+            # TODO RESTORE CUTOUT CONTEXT WINDOW SHORTNEING
```
moshemorad: Can you please explain what this todo mean?

holmes/core/tool_calling_llm.py
moshemorad: I might be missing something, but when the compcation happen in the middle of the loop i don't think we are clearing the old tool calls from `tool_calls` var. Am i missing it?

holmes/core/tool_calling_llm.py
nherment: removed

holmes/core/llm.py
nherment: Moved to 95%

PR #1018, opened 2025-09-30
add eval for incorrect tool call requiring approval in CLI

nherment:


aantn:
@nherment if I understand correctly, we were unable to reproduce the issue reported with Opus where Holmes uses AWK to answer this question, correct? If so, we can close the PR. The eval doesn't help if it doesn't reproduce the problem.

PR #1022, opened 2025-10-01
ROB-2227 avoid env exception when using bedrock with model list

RoiGlinik:
When using Bedrock model without AWS ENV Vars check_llm function raises an exception


Sheeproid (approved):

holmes/core/llm.py
```diff
@@ -144,6 +148,10 @@ def check_llm(
             os.environ.get("AWS_PROFILE") or os.environ.get("AWS_BEARER_TOKEN_BEDROCK")
         ):
             model_requirements = {"keys_in_environment": True, "missing_keys": []}
+        elif provider == "bedrock" and (
+            args.get("aws_access_key_id") and args.get("aws_secret_access_key")
+        ):
+            return  # break fast.
```
mainred: Do we need to combine the bedrock provide check?

mainred (commented):
Thanks for the change.
W

Sheeproid (approved):

PR #1028, opened 2025-10-02
ROB-2233: model attributes returned by the new relay models api

nherment:
:warning:  Breaking change for Robusta SaaS

This change depends on a new relay API created here: https://github.com/robusta-dev/relay/pull/291

### DIFF
--- a/tests/config_class/test_config_api_base_version.py
+++ b/tests/config_class/test_config_api_base_version.py
@@ -1,5 +1,6 @@
 from unittest.mock import MagicMock, patch
 
+from pydantic import SecretStr
 import yaml
 
 from holmes.config import Config
@@ -61,7 +62,7 @@ def test_config_get_llm_with_azure_openai_parameters():
     """Test that Config._get_llm passes api_base and api_version to DefaultLLM."""
     config = Config(
         model="azure/gpt-4o",
-        api_key="test-key",
+        api_key=SecretStr("test-key"),
         api_base="https://test.api.base",
         api_version="2023-12-01",
     )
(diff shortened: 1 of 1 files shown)


holmes/clients/robusta_client.py
```diff
@@ -14,13 +14,14 @@ class HolmesInfo(BaseModel):
...
-    models_args: Dict[str, Any] = Field(
-        default_factory=dict, alias="models_holmes_args"
-    )
-    default_model: Optional[str] = None
+    holmes_args: dict[str, Any]
+    model: str
+    is_default: bool
+
+class RobustaModelsResponse(BaseModel):
+    models: Dict[str, RobustaModel]
```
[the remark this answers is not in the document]
nherment: outdated

holmes/core/llm.py
```diff
@@ -109,6 +116,13 @@ def check_llm(
         api_version: Optional[str],
         args: Optional[dict] = None,
     ):
+        if getattr(
+            self, "is_robusta_model", False
+        ):  # using getattr because some tests call check_llm() without properly instantiating the class.
```
moshemorad: Can we fix the tests instead of handle it in the actual code?

holmes/core/llm.py
```diff
@@ -455,16 +486,19 @@ def configure_robusta_ai_model(self) -> None:
...
-                logging.info(f"Loading Robusta AI model: {model}")
-                args = robusta_models.models_args.get(model)
-                self._llms[model] = self._create_robusta_model_entry(model, args)
-
-            if robusta_models.default_model:
-                logging.info(
-                    f"Setting default Robusta AI model to: {robusta_models.default_model}"
+            default_model = None
+            for model_name in robusta_models.models:
+                model_data = robusta_models.models[model_name]
```
moshemorad: Instead of fetching it by key you could iterate over it with items and get both the key and the value in the for loop.

moshemorad (dismissed):

holmes/core/llm.py
nherment: Will do/

holmes/core/llm.py
nherment: ok

moshemorad (approved):

PR #1036, opened 2025-10-07
fix: allow the user to config api key in config file

mainred:
We introduced LLM Model Registry in https://github.com/robusta-dev/holmesgpt/pull/954 to better manage the LLM model, and removed api_key from Config class, which leads to user configured api_key not working anymore.
This change brings the api_key back to config file, and introduces ModelEntry class to clarify the format of the LLM parameters.


holmes/core/llm.py
```diff
@@ -36,6 +37,29 @@
 OVERRIDE_MAX_CONTENT_SIZE = environ_get_safe_int("OVERRIDE_MAX_CONTENT_SIZE")
 ROBUSTA_AI_MODEL_NAME = "Robusta"
 
+# ModelEntry represents a single LLM model configuration
+
+
+class ModelEntry(BaseModel):
+    model: str
+    # TODO: the name field seems to be redundant, can we remove it?
+    name: Optional[str] = None
```
mainred, line 46: I did not find the name filed is used any place except in L493? and though it's passed to DefaultLLM, DefaultLLM does not actually use this parameters.  Can we remove it? @moshemorad 

https://github.com/robusta-dev/holmesgpt/blob/7a4fe4c4c4872ccf4ef64c85ca3c444c4fbe0fb8/holmes/config.py#L467-L469

moshemorad: I think that for now it better to keep it. We are refactoring some of the init steps and we might find it useful.
Thanks.

PR #1048, opened 2025-10-09
Show braintrust traces on exceptions

aantn:
When testing deepseek with SambaNova, I ran into an issue where SambaNova was returning errors. I wanted to troubleshoot it with `--trace=braintrust`, but couldn't do so because traces are not shown on exceptions from the LLM. This PR fixes that.


holmes/interactive.py
mainred, line 1261: how about printing trace_ur in the finally logic and removing L1256?

aantn: Good idea, fixed.

aantn:
@mainred do you mean re-reviewing please?

PR #1049, opened 2025-10-09
ROB-2206 history changes improvements

RoiGlinik:
add support for workload and namespace filtering.
limit number of rows returned to a maximum of 200.
return only metadata info on change function and hint llm to use fetch_finding to get detailed change info.


holmes/core/supabase_dal.py
```diff
@@ -237,70 +237,58 @@ def get_resource_recommendation(
...
-            changes_response = (
+            query = (
                 self.client.table(ISSUES_TABLE)
-                .select("id", "subject_name", "subject_namespace", "description")
+                .select(
+                    "id",
+                    "subject_name",
+                    "subject_namespace",
+                    "description",
+                    "creation_date",
```
moshemorad: Do you think that we might want to add here more field that might improve the usage?
Like:
1. priority
2. instead of creation_date what about start_date and end_date
3. title
4. finding_type
5. subject_type

WDYT? @arikalon1 any thoughts?

holmes/core/supabase_dal.py
```diff
@@ -237,70 +237,58 @@ def get_resource_recommendation(
...
 
-    def get_configuration_changes(
-        self, start_datetime: str, end_datetime: str
+    def get_configuration_changes_metadata(
+        self,
+        start_datetime: str,
+        end_datetime: str,
+        limit: int = 100,
+        workload: Optional[str] = None,
+        ns: Optional[str] = None,
```
moshemorad, line 246: what about adding filter for cluster name? holmes is per cluster and it might be beneficial as well. WDYT?

RoiGlinik: The cluster name filtering is done by the supabase class so I close this.

holmes/core/supabase_dal.py
RoiGlinik: added title and subject type and timestamps

PR #1054, opened 2025-10-13
ROB-1933: make sure approved or denied tool call triggers a SSE event

nherment:
Ensures SSE messages are sent to the UI for tool results when tools are approved or denied. Otherwise the UI does not get updated.

### DIFF
--- a/holmes/core/tool_calling_llm.py
+++ b/holmes/core/tool_calling_llm.py
@@ -907,8 +907,7 @@ class ToolCallingLLM:
         if msgs and tool_decisions:
             logging.info(f"Processing {len(tool_decisions)} tool decisions")
             msgs, events = self.process_tool_decisions(msgs, tool_decisions)
-            for event in events:
-                yield event
+            yield from events
 
         messages: list[dict] = []
         if system_prompt:


holmes/core/tool_calling_llm.py
```diff
@@ -680,11 +681,13 @@ def _get_tool_call_result(
                 f"Failed to parse arguments for tool: {tool_name}. args: {tool_arguments}"
             )
 
-        tool_response = prevent_overly_repeated_tool_call(
-            tool_name=tool_name,
-            tool_params=tool_params,
-            tool_calls=previous_tool_calls,
-        )
+        tool_response = None
+        if not user_approved:
```
moshemorad, line 685: why we changed it? if the user approved it but it was already ran once we still would like to run it?

holmes/core/tool_calling_llm.py
```diff
@@ -893,11 +896,20 @@ def call_stream(
...
         """
         This function DOES NOT call llm.completion(stream=true).
         This function streams holmes one iteration at a time instead of waiting for all iterations to complete.
         """
+
+        # Process tool decisions if provided
+        if msgs and tool_decisions:
+            logging.info(f"Processing {len(tool_decisions)} tool decisions")
+            msgs, events = self.process_tool_decisions(msgs, tool_decisions)
+            for event in events:
```
moshemorad: In that case instead of for loop just to yield the item for the iterable you can use `yield from events`

holmes/core/tool_calling_llm.py
nherment: I changed it because I noticed it. My thinking is that if a user actively approves a command then it should run, regardless of any restriction, including repeated commands.

moshemorad: Not sure i fully agreed with that. maybe worth discuss it with product and make sure?

PR #1059, opened 2025-10-15
ROB-2156 limit loki logs

RoiGlinik:
Prevent option to use large limits on loki logs tool calls


RoiGlinik:
new logs truncate logic makes this probably unnecessary

PR #1061, opened 2025-10-19
[ROB-1734] Robusta runbook support + better runbook usage

Avi-Robusta:
This is an example of what is added to the beginning of the user prompt for the runbooks if there are global instructions, runbooks, or issue/workload instructions (the sections are created dynamically in the prompt in order based on what the user has configured)

```
# Runbook Selection
You (HolmesGPT) have access to runbooks with step-by-step troubleshooting instructions. If one of the following runbooks relates to the user's issue, you MUST fetch it with the fetch_runbook tool.
You (HolmesGPT) must follow runbook sources in this priority order:
1) Runbook Catalog (priority #1)
2) Global Instructions (priority #2)
## Runbook Catalog (priority #1)
   Here are MD runbooks:
   * networking/dns_troubleshooting_instructions.md | description: Runbook to investigate DNS resolution issue in Kubernetes clusters
   * upgrade/upgrade_troubleshooting_instructions.md | description: Runbook to troubleshoot upgrade issues in Azure Kubernetes Service clusters
   Here are Robusta runbooks:
   * id='8fe8e24d-6b53-47a6-92a5-b389938fa823' | title='Kafka Topic and Application Mapping' | symptom='Kafka lag issues, consumer falling behind'
   * id='40296b5c-2cb5-41df-b1f5-93441f894c44' | title='Pod Crashlooping Debugging Runbook' | symptom='Pod continuously crashing and restarting'
   * id='b2ffd311-f339-46a2-b379-1e53eb0bc1ed' | title='Signup Service Debugging Runbook' | symptom='Signup service failing, users can't register'
## Global Instructions (priority #2)
   * Always check service dependencies first
   * Verify environment variables are set correctly
   * Check logs for error patterns before investigating further
   * Contact the interlock team on all issues that could be related to STRIPE or the Stripe API
   * On all other issues contact the capricorn team
   


If a runbook might match the user's issue, you MUST:
1. Fetch the runbook with the `fetch_runbook` tool.
2. Decide based on the runbook's contents if it is relevant or not.
3. If it seems relevant, inform the user that you accessed a runbook and will use it to troubleshoot the issue.
4. To the maximum extent possible, follow the runbook instructions step-by-step.
5. Provide a detailed report of the steps you performed, including any findings or errors encountered.
6. If a runbook step requires tools or integrations you don't have access to, tell the user that you cannot perform that step due to missing tools.
```

This is an example of how it looks fetching a a robusta runbook
```
  id: b2ffd311-f339-46a2-b379-1e53eb0bc1ed
  symptom: Signup service failing, users can't register
  title: Signup Service Debugging Runbook
  instruction: |-
    # Signup Service Debugging Runbook

    ## Overview
    When the signup service is experiencing issues, the problem is often related to the payments service dependency. This runbook provides systematic debugging steps.

    ## Step 1: Check Signup Service Status
    ```bash
    kubectl get pods -l app=signup-service
    kubectl describe pod <signup-pod-name>
    kubectl logs <signup-pod-name>
    ```

    Look for:
    - Pod status and restart count
    - Error messages in logs
  ...
```


holmes/plugins/prompts/investigation_procedure.jinja2
```diff
@@ -6,6 +6,35 @@ CRITICAL: For multi-step questions, you MUST start by calling the TodoWrite tool
...
+# MANDATORY Fetching runbooks:
+Before starting any investigation, ALWAYS fetch all relevant runbooks using the `fetch_runbook` tool. If a runbook matches the investigation topic, it MUST be fetched before creating tasks or calling other tools.
+
+# CRITICAL RUNBOOK COMPLIANCE:
+- After fetching ANY runbook, you MUST read the "instruction" field IMMEDIATELY
+- If the instruction contains specific actions, you MUST execute them BEFORE proceeding
+- DO NOT proceed with investigation if runbook says to stop
+- Runbook instructions take ABSOLUTE PRIORITY over all other investigation steps
+
+# MANDATORY RUNBOOK EXECUTION PROCESS:
```
moshemorad: Looks like it is duplicating mostly the above section. are we sure that we need both of them?

holmes/plugins/runbooks/__init__.py
```diff
@@ -4,18 +4,56 @@
...
+
+    @staticmethod
+    def _repr_str(dumper, s: str):
+        s = s.replace("\\n", "\n")
+        return dumper.represent_scalar(
+            "tag:yaml.org,2002:str", s, style="|" if "\n" in s else None
+        )
+
+    # register representer (PyYAML API)
+    _LiteralDumper.add_representer(str, _repr_str)  # type: ignore
```
moshemorad, line 51: Can you explain why you need everything that going on here with yaml?

holmes/plugins/runbooks/__init__.py
```diff
@@ -66,17 +104,46 @@ class RunbookCatalogEntry(BaseModel):
...
+
+    def list_available_runbooks(self) -> list[str]:
+        return [entry.to_list_string() for entry in self.catalog]
+
+    def split_by_type(
+        self,
+    ) -> Tuple[List[RunbookCatalogEntry], List[RobustaRunbookInstruction]]:
+        md: List[RunbookCatalogEntry] = []
+        robusta: List[RobustaRunbookInstruction] = []  #
+        for e in self.catalog:
```
moshemorad: Please give better name to vars.

holmes/plugins/runbooks/__init__.py
```diff
@@ -85,14 +152,26 @@ def load_runbook_catalog() -> Optional[RunbookCatalog]:
     try:
         with open(catalogPath) as file:
             catalog_dict = json.load(file)
-            return RunbookCatalog(**catalog_dict)
+            catalog = RunbookCatalog(**catalog_dict)
     except json.JSONDecodeError as e:
         logging.error(f"Error decoding JSON from {catalogPath}: {e}")
+        return None
```
moshemorad: Why we return on error? we wouldn't like to load runbook from dal on error?

holmes/plugins/runbooks/__init__.py
```diff
@@ -85,14 +152,26 @@ def load_runbook_catalog() -> Optional[RunbookCatalog]:
...
     except Exception as e:
         logging.error(
             f"Unexpected error while loading runbook catalog from {catalogPath}: {e}"
         )
-    return None
+        return None
+
+    # Append additional runbooks from SupabaseDal if provided
+    if dal and dal.enabled:
```
moshemorad: Any reason to check if it is enabled here if we anyway check in the get runbook function?

holmes/core/supabase_dal.py
```diff
@@ -403,6 +404,69 @@ def get_issue_data(self, issue_id: Optional[str]) -> Optional[Dict]:
...
+            )
+            if not res.data:
+                return None
+
+            instructions = []
+            for row in res.data:
+                id = row.get("runbook_id")
+                symptom = row.get("symptoms")
+                title = row.get("subject_name")
+                if not id or not symptom:
```
moshemorad: in which cases id could be none or empty?

holmes/core/supabase_dal.py
```diff
@@ -403,6 +404,69 @@ def get_issue_data(self, issue_id: Optional[str]) -> Optional[Dict]:
...
+                    continue
```
moshemorad, line 436: we don't have validation in the UI for symproms? maybe worth to add log here in case it happens?

holmes/core/supabase_dal.py
```diff
@@ -403,6 +404,69 @@ def get_issue_data(self, issue_id: Optional[str]) -> Optional[Dict]:
...
+        row = res.data[0]
+        id = row.get("runbook_id")
+        symptom = row.get("symptoms")
+        title = row.get("subject_name")
+        raw_instruction = row.get("runbook").get("instructions")
+        try:
+            if isinstance(raw_instruction, list) and len(raw_instruction) == 1:
+                instruction = raw_instruction[0]
+            else:
+                instruction = json.loads(raw_instruction)[0]
```
moshemorad: Can you explain the use case here?
Honeslty im not sure i understand what is going on here. if it is list and has more then one item we will try to load it as string? it won't cause a failiure?

holmes/core/supabase_dal.py
```diff
@@ -403,6 +404,69 @@ def get_issue_data(self, issue_id: Optional[str]) -> Optional[Dict]:
...
+        except Exception:
+            instruction = str(raw_instruction)
```
moshemorad, line 481: Again can you explain what we just stringfy and return it that way? I feel like im missing something.

holmes/plugins/toolsets/runbook/runbook_fetcher.py
```diff
@@ -21,42 +22,52 @@
...
             description="Get runbook content by runbook link. Use this to get troubleshooting steps for incidents",
             parameters={
-                "link": ToolParameter(
-                    description=f"The link to the runbook (non-empty string required). Must be one of: {runbook_list}",
+                "runbook_id": ToolParameter(
+                    description=f"The runbook_id: either a UUID or a .md filename. Must be one of: {runbook_list}",
+                    type="string",
+                    required=True,
+                ),
+                "type": ToolParameter(
```
moshemorad: Is there a good reason that we need to move this type to the llm to provide other than how we are loading the runbooks? if so i think we just need to align all runbooks to work the same and it should be transparent to the llm.

When loading runbooks from url we can assign them some id and just do it in the code instead no?

holmes/plugins/toolsets/runbook/runbook_fetcher.py
```diff
@@ -21,42 +22,52 @@
 from holmes.plugins.toolsets.utils import toolset_name_for_one_liner
 
 
-# TODO(mainred): currently we support fetch runbooks hosted internally, in the future we may want to support fetching
-# runbooks from external sources as well.
+class RunbookType(str, Enum):
```
moshemorad: Looks like eval 42 is failed, can you double check it isn't due to the changed in this pr? since the eval scenario is about eval

holmes/core/supabase_dal.py
Avi-Robusta: left over from prev version

holmes/core/supabase_dal.py
Avi-Robusta: this is a json the response for the "runbook" column, no db guarenteed format of the instructions object
```
{
  "documents": [],
  "instructions": [
    "check pod logs"
  ]
}
```

so this enforces that whatever output for instructions if its a list or object or string is passed to holmes.
I think older versions of it were strings.
It means holmes wont break if anyone changes the instructions + documents in the table

holmes/plugins/runbooks/__init__.py
Avi-Robusta: our DB runbooks are multiline which messes with the output format of jsons and yaml 
`_LiteralDumper.add_representer` is used to output multi line strings correctly like: 
```
id: 2e936edf-b5ca-4335-ad50-262384efa207
symptom: test symptom
title: test title
instruction: |-
  test
  test2
  test3
  test4
```

instead of 
```
id: 2e936edf-b5ca-4335-ad50-262384efa207
symptom: test symptom
title: test title
instruction: test
test2
test3
test4
```
which is not a valid yaml and if the runbook is written poorly it has even confused holmes

i guess i need to comment it better to explain why its critical

holmes/core/supabase_dal.py
```diff
@@ -403,6 +404,69 @@ def get_issue_data(self, issue_id: Optional[str]) -> Optional[Dict]:
...
+        if not res.data or len(res.data) != 1:
+            return None
+
+        row = res.data[0]
+        id = row.get("runbook_id")
+        symptom = row.get("symptoms")
+        title = row.get("subject_name")
+        raw_instruction = row.get("runbook").get("instructions")
+        if isinstance(raw_instruction, list) and len(raw_instruction) >= 0:
+            instruction = raw_instruction[0]
```
moshemorad, line 469: Honestly i just don't get what is going on here...
Why if it is list we are only taking the first element?

Let's take it f2f

moshemorad (changes requested):

PR #1065, opened 2025-10-21
ROB-2290 fetch resource issues tool

RoiGlinik:


holmes/plugins/toolsets/robusta/robusta.py
```diff
@@ -285,6 +289,26 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            ),
+            add_cluster_filter=True,
+        )
+
+    def _fetch_change_history(self, params: Dict) -> Optional[List[Dict]]:  # type: ignore
+        return super()._fetch_change_history(params, finding_type=FindingType.ISSUE)
+
+    def get_parameterized_one_liner(self, params: Dict) -> str:
+        return f"Robusta: fetch resource issues metadata {params}"
+
```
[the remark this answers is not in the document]
RoiGlinik: you are wrong

holmes/plugins/toolsets/robusta/robusta.py
```diff
@@ -285,6 +289,26 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            name="fetch_resource_issues_metadata",
+            description=(
+                "Fetch issues and alert metadata in a given time range. "
+                "Must be filtered on a given namespace and specific kubernetes resource such as pod, deployment, job, etc."
+                "Use fetch_finding_by_id to get further information on a specific issue or alert."
+            ),
+            add_cluster_filter=True,
+        )
+
+    def _fetch_change_history(self, params: Dict) -> Optional[List[Dict]]:  # type: ignore
```
moshemorad: It is very weird that the name of this method is fetch_change_history. Please rename it.

holmes/core/supabase_dal.py
```diff
@@ -53,6 +54,12 @@
 ENRICHMENT_BLACKLIST_SET = set(ENRICHMENT_BLACKLIST)
 
 
+class FindingType(str, Enum):
+    ALL = "all"
```
moshemorad: In which cases we are using all? im trying to understand the use case for it and not fully sure i understand it.

holmes/plugins/toolsets/robusta/robusta.py
```diff
@@ -168,7 +168,7 @@ def __init__(
                 required=True,
             ),
             END_TIME: ToolParameter(
-                description="The starting time boundary for the search period. String in RFC3339 format.",
+                description="The ending time boundary for the search period. String in RFC3339 format.",
```
moshemorad, line 171: Nice catch :)

holmes/core/supabase_dal.py
```diff
@@ -53,6 +54,12 @@
 ENRICHMENT_BLACKLIST_SET = set(ENRICHMENT_BLACKLIST)
 
 
+class FindingType(str, Enum):
```
moshemorad, line 57: Looks like datadog evals are failling can you please also take a look on that?

holmes/core/supabase_dal.py
RoiGlinik: Yea its kind of a for the future so ill delete it for now

holmes/core/supabase_dal.py
RoiGlinik: That test case uses 128+ tools some not from the toolset list might be some other bug. butt removed some tools so it passes.

moshemorad (approved):

PR #1085, opened 2025-10-28
Fix helm customServiceAccountName

### LINKED ISSUE #1083: helm: customServiceAccountName does not change the name of created ServiceAccount
### What happened?

When setting `customServiceAccountName` in values.yaml, the Deployment correctly references the custom service account name, but the ServiceAccount resource created by the chart still uses the hardcoded default name `{{ .Release.Name }}-holmes-service-account`. This causes pods to fail to start 

I would like to contribute a PR to fix this issue if that's helpful.

Chart Version: 0.16.0

When `customServiceAccountName` is set, the created ServiceAccount should use that custom name instead of the default `{{ .Release.Name }}-holmes-service-account`. Both the ServiceAccount creation and Deployment reference should use the same name.

Set `customServiceAccountName: "holmes-gpt"` in values.yaml
run helm template

_No response_

KJone1:
Fixes #1083

Helper Logic:
  ```yaml
 if customServiceAccountName is set → use it (regardless of createServiceAccount)
 else if createServiceAccount is true → use {{ .Release.Name }}-holmes-service-account
 else → use "default"
```

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,178 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: helm/holmes/templates/_helpers.tpl [hunk 1]
```diff
@@ -0,0 +1,14 @@
+{{/*
+Return the service account name to use
+*/}}
+{{- define "holmes.serviceAccountName" -}}
```
moshemorad, line 4: Not sure that this is the right logic. 

I think there are couple of use cases.
A. createServiceAccount = true & customServiceAccountName = true -> we would like to return customServiceAccountName and make sure that we also set it in `holmesgpt-service-account.yaml` (i think we missed it there)

B. createServiceAccount = false & customServiceAccountName = true -> we still want to use the customServiceAccountName in all places (If you already have service account that you would like to use it.)

C. createServiceAccount = false & customServiceAccountName = false -> default.


Is that making sense?

KJone1: legit, updated the flow @moshemorad

moshemorad (approved):

moshemorad:
Looks good!
@KJone1 thanks for your contribution

PR #1089, opened 2025-10-29
[ROB-2429 No ssl verify option for litellm args

Avi-Robusta:
https://docs.litellm.ai/docs/providers/openai#set-ssl_verifyfalse


aantn:
@Avi-Robusta please make sure to document.

aantn:
Meaning, we need to explain this in public facing holmes docs.

Avi-Robusta:
@arikalon1 said not relevant since there are other ways to set ssl cert so it should always be verifyable

PR #1102, opened 2025-11-05
[ROB-2201] Mcp streamable-http support

Avi-Robusta:
example config:
```
  kubernetes_streamable:
    description: "kubernetes mcp (SSE mode)."
    config:
      mode: streamable-http
      url: "http://localhost:5555/mcp"
      headers: ...
    llm_instructions: "used to get kubernetes information"
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,11 +25,11 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 # Needed for kubectl
 ENV VERIFY_CHECKSUM=true \
     VERIFY_SIGNATURES=true
-RUN curl -fsSL https://pkgs.k8s.io/core:/stable:/v1.32/deb/Release.key -o Release.key
+RUN curl -fsSL https://pkgs.k8s.io/core:/stable:/v1.34/deb/Release.key -o Release.key
 
 # Set the architecture-specific kube lineage URLs
-ARG KUBE_LINEAGE_ARM_URL=https://github.com/robusta-dev/kube-lineage/releases/download/v2.2.3/kube-lineage-macos-latest-v2.2.3
-ARG KUBE_LINEAGE_AMD_URL=https://github.com/robusta-dev/kube-lineage/releases/download/v2.2.3/kube-lineage-ubuntu-latest-v2.2.3
+ARG KUBE_LINEAGE_ARM_URL=https://github.com/robusta-dev/kube-lineage/releases/download/v2.2.4/kube-lineage-macos-latest-v2.2.4
+ARG KUBE_LINEAGE_AMD_URL=https://github.com/robusta-dev/kube-lineage/releases/download/v2.2.4/kube-lineage-ubuntu-latest-v2.2.4
 # Define a build argument to identify the platform
 ARG TARGETPLATFORM
 # Conditional download based on the platform
@@ -44,10 +44,8 @@ RUN chmod 777 kube-lineage
 RUN ./kube-lineage --version
 
 # Set the architecture-specific argocd URLs
-# Freezing to argocd 2.13.5 as it has fixes CVE-2025-21613 and CVE-2025-21614.
-# The argocd release 2.14.2 (latest as 2025-02-19) unfortunately has these CVEs.
-ARG ARGOCD_ARM_URL=https://github.com/argoproj/argo-cd/releases/download/v2.13.5/argocd-linux-arm64
-ARG ARGOCD_AMD_URL=https://github.com/argoproj/argo-cd/releases/download/v2.13.5/argocd-linux-amd64
+ARG ARGOCD_ARM_URL=https://github.com/argoproj/argo-cd/releases/download/v3.2.0/argocd-linux-arm64
+ARG ARGOCD_AMD_URL=https://github.com/argoproj/argo-cd/releases/download/v3.2.0/argocd-linux-amd64
 # Conditional download based on the platform
 RUN if [ "$TARGETPLATFORM" = "linux/arm64" ]; then \
     curl -L -o argocd $ARGOCD_ARM_URL; \
@@ -103,7 +101,7 @@ RUN apt-get update \
 # Set up kubectl
 COPY --from=builder /app/Release.key Release.key
 RUN cat Release.key |  gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg \
-    && echo 'deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v1.32/deb/ /' | tee /etc/apt/sources.list.d/kubernetes.list \
+    && echo 'deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v1.34/deb/ /' | tee /etc/apt/sources.list.d/kubernetes.list \
     && apt-get update
 RUN apt-get install -y kubectl
 
--- a/helm/holmes/values.yaml
+++ b/helm/holmes/values.yaml
@@ -81,7 +81,7 @@ mcpAddons:
       name: "aws-api-mcp-sa"
       annotations: {}  # Add EKS IRSA annotations here, e.g., eks.amazonaws.com/role-arn
 
-    image: "aws-api-mcp-server:1.0.0"
+    image: "aws-api-mcp-server:1.0.1"
     registry: "us-central1-docker.pkg.dev/genuine-flight-317411/devel"
 
     resources:
--- a/holmes/plugins/toolsets/git.py
+++ b/holmes/plugins/toolsets/git.py
@@ -24,10 +24,12 @@ class GitHubConfig(BaseModel):
     git_repo: str
     git_credentials: str
     git_branch: str = "main"
+    git_url: str = "https://api.github.com"
 
 
 class GitToolset(Toolset):
     git_repo: Optional[str] = None
+    git_url: Optional[str] = None
     git_credentials: Optional[str] = None
     git_branch: Optional[str] = None
     _created_branches: set[str] = set()  # Track branches created by the tool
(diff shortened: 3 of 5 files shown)


tests/test_mcp_toolset.py
```diff
@@ -99,3 +105,257 @@ def test_mcpserver_no_headers():
...
+        name="list_authorizations",
+        inputSchema={"type": "object", "properties": {}, "required": []},
+        description="List all available authorization IDs",
+    )
+
+    mcp_tool = RemoteMCPTool.create(
+        "http://localhost:1234/mcp/messages", tool, mode=MCPMode.STREAMABLE_HTTP
+    )
+
+    mock_read_stream = AsyncMock()
```
moshemorad: Please don't mock the actual seesions and client.
Please use https://lundberg.github.io/respx/

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -11,19 +11,48 @@
...
+
+class MCPMode(str, Enum):
+    SSE = "sse"
+    STREAMABLE_HTTP = "streamable-http"
+
+
+def get_mode_from_dict(data: dict[str, Any] | None) -> MCPMode:
+    if data is None:
+        return MCPMode.SSE
+    mode_value = data.get("mode") or (data.get("config") or {}).get("mode") or "sse"
```
moshemorad: Why we are trying to get it from config?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -11,19 +11,48 @@
...
+    try:
+        return MCPMode(mode_value)
+    except ValueError:
+        logging.error(f'Mode "{mode_value}" is not supported, falling back to SSE')
+        return MCPMode.SSE
```
moshemorad: So if the user has a typo when setting mode to streamable-http we will just change it to sse for the user?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -91,30 +144,45 @@ class RemoteMCPToolset(Toolset):
...
         return self.config and self.config.get("headers")
 
-    @field_validator("url", mode="before")
-    def append_sse_if_missing(cls, v):
-        if isinstance(v, str) and not v.rstrip("/").endswith("/sse"):
-            v = v.rstrip("/") + "/sse"
-        return v
+    @model_validator(mode="before")
+    @classmethod
+    def append_suffix_if_missing(cls, data: Any) -> Any:
```
moshemorad: I think that since this validator is running in before mode just creates huge headache
Can we consider to change his mode to after and that way you will have everything already converted? instead of fetching mode from dict etc..

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -91,30 +144,45 @@ class RemoteMCPToolset(Toolset):
...
+                and mode == MCPMode.SSE
+                and not url.rstrip("/").endswith("/sse")
+            ):
+                data["url"] = url.rstrip("/") + "/sse"
+        return data
 
     # used as a CallablePrerequisite, config added for that case.
     def init_server_tools(self, config: dict[str, Any]) -> Tuple[bool, str]:
         try:
+            self.mode = get_mode_from_config(config)
```
moshemorad: Why are we doing this? isn't the toolset already has the correct mode?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -11,19 +11,48 @@
...
+    mode_value = config.get("mode") or "sse"
+    try:
+        return MCPMode(mode_value)
+    except ValueError:
+        logging.error(f'Mode "{mode_value}" is not supported, falling back to SSE')
+        return MCPMode.SSE
 
 
 class RemoteMCPTool(Tool):
     url: str
```
moshemorad: I think it is a good chance to refacotr this.
No reason that the tool wil duplicate every arg that the toolset has.

Can we just init the tool with the parent toolset? that way we will not need to add every no arg to here + the remote mcp toolset parent can expose function that creates the client session for the tool to use.

WDYT?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -123,14 +191,23 @@ def init_server_tools(self, config: dict[str, Any]) -> Tuple[bool, str]:
...
-        ):
-            async with ClientSession(read_stream, write_stream) as session:
-                _ = await session.initialize()
-                return await session.list_tools()
+        if self.mode == MCPMode.SSE:
+            async with sse_client(str(self.url), headers=self.get_headers()) as (
+                read_stream,
+                write_stream,
+            ):
+                return await self._list_tools_with_session(read_stream, write_stream)
```
moshemorad: I think that we need to change and refactor the order here. 

Since list toolset need the client session and not the opposite way around it would be better to change the call order. That way we will have funciton that creates the client session and all relevant function can reuse it.

docs/data-sources/remote-mcp-servers.md
```diff
@@ -3,35 +3,103 @@
...
+
+The transport mode is specified in the `config` section of your MCP server configuration:
+
+```yaml
+mcp_servers:
+  my_server:
+    description: "My MCP server"
+    url: "http://example.com:8000/mcp/messages"  # Path depends on your server (could be /mcp, /mcp/messages, etc.)
+    config:
+      mode: streamable-http  # Explicitly set the mode
```
moshemorad, line 35: Please move it out of the config.

docs/data-sources/remote-mcp-servers.md
```diff
@@ -3,35 +3,103 @@
...
+
+!!! note "URL Path Variability"
+    Different MCP servers may use different endpoint paths. Common examples include:
+    - `/mcp/messages` - Used by some servers
+    - `/mcp` - Used by other servers (e.g., Azure Kubernetes MCP)
+    - Custom paths as defined by your server
+
+    The streamable-http client automatically handles the protocol regardless of the path. Consult your MCP server's documentation to determine the correct endpoint URL.
+
+If no mode is specified, the system defaults to `sse` for backward compatibility. However, **you should explicitly set `mode: streamable-http`** for new servers.
```
moshemorad: Add depraction note and remind again that this going to change in the near future.

docs/data-sources/remote-mcp-servers.md
```diff
@@ -3,35 +3,103 @@
...
+        key: "{{ env.my_mcp_server_key }}"  # You can use holmes environment variables as headers for the MCP server requests.
+    llm_instructions: "This is a cluster-local MCP server that provides internal cluster data and operations. Use it for accessing cluster-specific information, internal services, or custom tooling deployed within the Kubernetes environment."
 ```
 
 ## Example: Working with Stdio MCP servers
 
-MCP currently supports three transport mechanisms: stdio, Server-Sent Events (SSE), and Streamable HTTP.
-At this time, HolmesGPT is compatible only with MCP servers that use SSE.
+MCP currently supports three transport mechanisms: stdio, Server-Sent Events (SSE - deprecated), and Streamable HTTP.
+HolmesGPT supports both **streamable-http** (recommended) and **SSE** (deprecated) transport modes.
```
moshemorad, line 100: Maybe its worth to mention that soon it will be supported?

tests/test_mcp_toolset.py
```diff
@@ -99,3 +105,257 @@ def test_mcpserver_no_headers():
     )
 
     assert mcp_toolset1.get_headers() is None
+
+
+def test_streamable_http_list_authorizations():
```
moshemorad: A. Let start and reorginize our tests better. Seperate the sse tests from http test.
B. Change tests name to describe what they are checking and not what they are doing. e.g. `test_streamable_http_list_authorizations` -> describe that the test call list_authrizations but not what it checks. consider use given__when__then pattern.

moshemorad (changes requested):

docs/data-sources/remote-mcp-servers.md
Avi-Robusta: not worth it, ill change it in the docs in a day when its added - its not worth the risk of forgetting to remove it in a day or so

holmes/plugins/toolsets/mcp/toolset_mcp.py
Avi-Robusta: refactored, needed for cleaner error handling and clearer to user on misconfiguration

tests/test_mcp_toolset.py
Avi-Robusta: It wont work for mcp protocl
respx intercepts HTTP correctly, but MCP's internal stream architecture requires network-like timing that instant mocked responses don't provide, causing deadlocks in the zero-buffer stream synchronization
I tried it and it deadlocks all the time

docs/data-sources/remote-mcp-servers.md
Avi-Robusta: What is the benifit of moving it out of the config? 
Here I can handle it cleaner and show a more clear errors when its incorrect.
```
Loaded models: ['...']                                                            
Mode "streamable-https" is not supported                                                                                                                                                                                                           
❌ Toolset kubernetes_streamable: Invalid mode streamable-https in kubernetes_streamable defined, only 'sse' or 'streamable-http' supported                                                                                                        
Using 2 datasources (toolsets). To refresh: use flag `--refresh-toolsets`
```
to do that will just add more code for cosmetic benefit from what i can tell

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -84,53 +118,94 @@ def parse_input_schema(
...
 
     def model_post_init(self, __context: Any) -> None:
-        self.prerequisites = [CallablePrerequisite(callable=self.init_server_tools)]
-
-    def get_headers(self) -> Optional[Dict[str, str]]:
-        return self.config and self.config.get("headers")
+        self.prerequisites = [
+            CallablePrerequisite(callable=self.prerequisites_callable)
+        ]
+
```
moshemorad, line 136: why we need mode & url in the toolset level if we are having them in the config?

PR #1104, opened 2025-11-05
[ROB-2428] service account docs

Avi-Robusta:


docs/reference/kubernetes-permissions.md
```diff
@@ -0,0 +1,201 @@
+# Kubernetes Required Permissions
+
+This document details the Kubernetes RBAC permissions required by HolmesGPT when deployed via Helm chart in a Kubernetes cluster.
```
moshemorad: @Avi-Robusta I think maybe we need to change the emphasis here. 

e.g. instead of running inside/cli worth to change it to permissions that need by holmes to work and give the best results.

@aantn WDYT?

aantn: Agreed. I don't think we need to document the exact permissions in this doc, as we can just refer users to the yaml in the github, which will always be up to date.

The points to get across here IMO are:
1. HolmesGPT inherits permissions for accessing Kubernetes from it's environment. When running locally, that is your current kubectl context and when running in-cluster that is the ServiceAccount defined in the helm chart (link it).
2. You can modify these permissions and HolmesGPT will automatically adjust itself. If it tries to run kubectl commands that doesn't have permissions for, HolmesGPT will discover the lack of permissions and adjust its behaviour.
3. For most users, we recommend giving read-access to all non-sensitive resources in the cluster.
4. If you want to adjust this, you can do so by ....

Avi-Robusta: @moshemorad @aantn how about the latest version

aantn: Looks better. Only comment is on this part:

> Edit the Helm chart's ServiceAccount template or modify the ClusterRole after deployment:

You show an example of how to edit the ClusterRole after deployment, but that's extremely inconvenient. Most users will want to edit it via the Helm chart instead. I would show that, not show a kubectl edit command.

PR #1110, opened 2025-11-09
ROB-2346  resync robusta models if robusta model is not found

RoiGlinik:
This case probably means robusta models is out of sync


holmes/core/llm.py
```diff
@@ -633,9 +633,17 @@ def get_model_params(self, model_key: Optional[str] = None) -> ModelEntry:
 
         if model_key:
             model_params = self._llms.get(model_key)
-            if model_params is not None:
+            if model_params:
                 logging.info(f"Using selected model: {model_key}")
-                return model_params.copy()
+                return model_params.model_copy()
+
+            if model_key.startswith("Robusta/"):
```
moshemorad: @RoiGlinik can you please explain what is the use case here?

RoiGlinik:
@coderabbitai review

PR #1114, opened 2025-11-10
chore: allow cli to pick model from model list

mainred:
When the CLI user define model list, and specify the model flag, holmesgpt will be overwrite the existing model with empty ones if the other llm configurations are not present in the config.
We allow the user to specify the model only when there's no api_key defined the configuration file. 

## Test

The model_list.yaml contains
```yaml
azure/gpt-4.1:
  api_key: "***"
  model: azure/gpt-4.1
  api_base: https://**.openai.azure.com/
  api_version: 2025-01-01-preview
azure/gpt-5:
  api_key: "****"
  model: azure/gpt-5
  api_base: https://**.openai.azure.com/
  api_version: 2025-01-01-preview
```

config.yaml contains
```
model: "azure/gpt-4.1"
```

```
 poetry run holmes ask "what pods are running?"
Loaded models: ['azure/gpt-4.1', 'azure/gpt-5']
✅ Toolset docker/core
✅ Toolset kubernetes/core
✅ Toolset aks/core
Using 28 datasources (toolsets). To refresh: use flag `--refresh-toolsets`
NO ENABLED LOGGING TOOLSET
Using selected model: azure/gpt-4.1
Using model: azure/gpt-4.1 (1,047,576 total tokens, 32,768 output tokens)
Welcome to HolmesGPT: Type '/exit' to exit, '/help' for commands.
User: what pods are running?

Thinking...

The AI requested 1 tool call(s).
Running tool #1 kubectl_get_by_kind_in_cluster: kubectl get -A --show-labels -o wide pod
  Finished #1 in 0.30s, output length: 59,202 characters (163 lines) - /show 1 to view contents
... ...

```

same case when I use  `poetry run holmes ask "what pods are running?" --model=azure/gpt-4.1`

### DIFF
--- a/server.py
+++ b/server.py
@@ -82,6 +82,11 @@ dal = config.dal
 
 
 def sync_before_server_start():
+    if not dal.enabled:
+        logging.info(
+            "Skipping holmes status and toolsets synchronization - not connected to Robusta platform"
+        )
+        return
     try:
         update_holmes_status_in_db(dal, config)
     except Exception:


holmes/core/llm.py
```diff
@@ -538,14 +537,25 @@ def _init_models(self):
...
-                is_robusta_model=False,
-                api_key=self.config.api_key,
-                api_version=self.config.api_version,
-            )
+            # Chances are that for the cli, the user configures model_list and specify which model to use in the flag,
+            # so we don't override existing model entry from the model list file.
+            existing_model = self._llms.get(self.config.model)
+            # When self.config.api_key and self.config.model are both set, we assume the user wants to override the
+            # model entry from the model list file.
+            if not existing_model or self.config.api_key:
```
moshemorad, line 545: Thinking here loudly, I think that this is very confusing, in which cases the user wants to override the api key of the models. 

Also, if he would like to override it he can't just set the value in the model list? we already have couple of places (toolset for example) that we are allowing those kind od override and it is complexing the code very much.

mainred: I just want to make it safe when both model and api_key are specified in the configuration file(~/.homes/config), we pick the configuration file instead of from the model list.
I agree too much check makes the code complicated. I think we can allow the user either to specify the model config in either model list or configuration file, and in cli when model list is defined, he can specify which model to use through model flag.

mainred: This merge the cli and server behavior when model list is defined.

mainred:
@moshemorad May I get a second review on this one?

holmes/core/llm.py
moshemorad: The thing that confuse me in that case is what happen if you set all the params like base_url in the model list and not in the config. It will override it? 

I think it is very cumbersome. you either set all in the model list or set all in the config. In that case if the user set only the api key we will also overrider other params imlicitly and im not sure that this is the right behivour. 

Are you sure that this is what we want to do?

mainred: If the user set all in the model list, how will he select which model to use through CLI?

moshemorad: Looks im missing something. In the suggested implementation we are overriding the model with the api_key, how this is connected to selecting the right model though cli?

@Avi-Robusta has an open pr that still need tests here: https://github.com/HolmesGPT/holmesgpt/pull/1096/files#diff-008177b02b41ee19dca44738e38891469986b5f51dfd58bc1b2ae22231471679R550-R556 

In that pr you can select models from the model list and use them.
We still need to add tests + docs for this flow. 

Can you help me understand what im missing in your suggested pr?

mainred: I think https://github.com/HolmesGPT/holmesgpt/pull/1096 addresses the same issue as I want to in this PR. Since the PR is opened for review, let's focus https://github.com/HolmesGPT/holmesgpt/pull/1096 instead.

mainred: The user scenario is the user specify the model list and select the model from CLL.

mainred:
Duplicate of https://github.com/HolmesGPT/holmesgpt/pull/1096

holmes/core/llm.py
moshemorad: Gotcha, thanks @mainred !

PR #1115, opened 2025-11-12
ROB-2566 grafana healthcheck improvement

RoiGlinik:
2. use the grafana healthcheck without proxy if data soruce id provided.
3. support both cloud and local loki cases

retry them in that order.

### DIFF
--- a/holmes/plugins/toolsets/grafana/grafana_api.py
+++ b/holmes/plugins/toolsets/grafana/grafana_api.py
@@ -22,6 +22,14 @@ def _try_health_url(url: str, headers: dict) -> None:
 
 
 def grafana_health_check(config: GrafanaConfig) -> Tuple[bool, str]:
+    """
+    Tests a healthcheck url for grafna loki.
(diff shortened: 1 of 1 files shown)


holmes/plugins/toolsets/grafana/grafana_api.py
```diff
@@ -6,37 +6,51 @@
...
-        error_msg = f"Failed to fetch grafana health status at {url}. {str(e)}"
-
-        # Add helpful hint if this looks like a common misconfiguration
-        if config.grafana_datasource_uid and ":3100" in config.url:
-            error_msg += (
-                "\n\nPossible configuration issue: grafana_datasource_uid is set but URL contains port 3100 "
-                "(typically used for direct Loki connections). Please verify:\n"
-                "- If connecting directly to Loki: remove grafana_datasource_uid from config\n"
-                "- If connecting via Grafana proxy: ensure URL points to Grafana (usually port 3000)"
+    health_urls = []
```
moshemorad, line 33: Please add here doc string that elaborate about which system works in which way for knowledge sharing.

PR #1120, opened 2025-11-16
Doing concurrent calls to the AWS MCP caused errors 

arikalon1:
changed to sequential calls per MCP

Add 120 seconds MCP timeout
Fix aws mcp tool description to contain the cli command


holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -20,8 +21,21 @@
...
 import logging
+import threading
 from enum import Enum
 
+# Lock per MCP server URL to serialize calls to the same server
+_server_locks: Dict[str, threading.Lock] = {}
+_locks_lock = threading.Lock()
+
+
+def get_server_lock(url: str) -> threading.Lock:
```
moshemorad, line 34: Thinking here loud, any reason that it is global and not part of the RemoteMCPToolset?

PR #1121, opened 2025-11-17
mcp exception fix

Avi-Robusta:
circular import when dumping to db fix

# When Circular Reference Issues Occur
## The circular reference happens when:
1) A Tool subclass defines toolset: SomeToolset as a class-level type annotation (making it a Pydantic model field)
2) AND does not override __init__() to pass toolset as an extra kwarg
3) In this case, Pydantic treats toolset as a model field that gets serialized
4) When model_dump() is called on the toolset, it tries to serialize:
Toolset → Tools → Toolset → Tools → ... (circular reference)
## When It Does NOT Happen
- A Tool subclass overrides __init__() and passes toolset via super().__init__(toolset=toolset)
- In this case, toolset is stored as a regular Python attribute (not a model field) and is excluded from serialization
- Examples: GitReadFileWithLineNumbers, BaseRabbitMQTool, BaseKafkaTool, etc.

Added additional tests to validate holmes_sync_toolsets_status dumps the models to the DB without error

### DIFF
--- a/tests/test_holmes_sync_toolsets.py
+++ b/tests/test_holmes_sync_toolsets.py
@@ -1,4 +1,3 @@
-import logging
 import os
 import subprocess
 from typing import Any, Dict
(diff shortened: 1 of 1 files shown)


holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -58,7 +58,7 @@ async def get_initialized_mcp_session(
 
 
 class RemoteMCPTool(Tool):
-    toolset: "RemoteMCPToolset"
+    toolset: "RemoteMCPToolset" = Field(exclude=True)
```
moshemorad, line 61: Can you elaborate in which cases it happens?

Also please add small test so it won't break again.

Thanks.

moshemorad (changes requested):

tests/test_holmes_sync_toolsets.py
```diff
@@ -388,3 +394,80 @@ def test_sync_toolsets_with_toolset_having_failing_callable_prerequisite(
...
+def test_toolsets_dumpable_with_bad_tool_test(mock_dal, mock_config):
+    """Test RemoteMCPToolset serialization - fails without Field(exclude=True), passes with it."""
+    original_toolsets = list(mock_config.create_tool_executor.return_value.toolsets)
+
+    bad_toolset = _create_bad_toolset()
+    all_toolsets_bad = list(original_toolsets)
+    all_toolsets_bad.append(bad_toolset)
+    mock_config.create_tool_executor.return_value.toolsets = all_toolsets_bad
+
+    try:
```
moshemorad: Please use pytest raises and split the test to  2 usecases https://docs.pytest.org/en/stable/reference/reference.html#pytest-raises

PR #1124, opened 2025-11-18
ROB-2584: improved krr tool

nherment:


holmes/core/supabase_dal.py
```diff
@@ -253,42 +254,92 @@ def sign_in(self) -> str:
...
+        scans_results_response = query.execute()
+
+        if not len(scans_results_response.data):
             return None
 
+        results = scans_results_response.data
+
+        # If sorting by priority, we already ordered in the query
+        if sort_by == "priority":
+            return results[:limit]
```
moshemorad: If it is already by priority we can do the limit on the db as well no?

holmes/core/supabase_dal.py
```diff
@@ -253,42 +254,92 @@ def sign_in(self) -> str:
...
+        # For priority sorting, we can use the database's order
+        if sort_by == "priority":
+            query = query.order("priority", desc=True)
+
+        scans_results_response = query.execute()
+
+        if not len(scans_results_response.data):
             return None
 
+        results = scans_results_response.data
```
moshemorad, line 329: In some cases like when matching name it could return only 1 result. I think it might be a good idea just return in that case. WDYT?

holmes/core/supabase_dal.py
```diff
@@ -253,42 +254,92 @@ def sign_in(self) -> str:
...
+
+        # If sorting by priority, we already ordered in the query
+        if sort_by == "priority":
+            return results[:limit]
+
+        # Sort by calculated savings (descending)
+        results_with_savings = [
+            (result, calculate_krr_savings(result, sort_by)) for result in results
+        ]
+        results_with_savings.sort(key=lambda x: x[1], reverse=True)
```
moshemorad, line 342: In the ticke there is this use case `show me top KRR recommendations for CPU for non-GPU workloads` how the non-GPU workload requirement is met in that case? we expect to return the results and that the llm will filter it? or there is posibility that after we will filter there will be some workload gpu and hence we won't get the requested reqcomendation? Or am i missing something?

nherment: As far as I could see, KRR does not return GPU info. Therefore it'll be up to the LLM to find out what kubernetes resources use or do not use GPU workloads.

PR #1136, opened 2025-11-25
Improvements to test framework and bugfix in bash toolset return code handling

Sheeproid:
To learn more look at individual commits


tests/llm/utils/env_vars.py
moshemorad, line 1: Please rename the file, it isn't file that added tests but a util file. No need for the test_ prefix in the file name.

moshemorad (approved):

tests/llm/utils/env_vars.py
Sheeproid: There are multiple files named 'test' in this folder, I think that why it is also named test_. anyway, renamed it to env_var.

PR #1137, opened 2025-11-25
[ROB-2574] also [ROB-2548] Evals fixes and benchmark streamlining

Avi-Robusta:
additionally bug Fixes in evals 
Eval bugs that were fixed
1) on evals with toolsets.yaml defined, additional tools were loaded
2) on evals where no toolsets.yaml was defined 156 tools could be loaded (depending on the computer running it) - majority not even relevant to the eval - ( for example opensearch/aks/eks tools can be automatically loaded during tests even when not enabled)
   - many llms have a hard limit of 128 tools - i.e. gpt-5.1
   - A `default_toolsets.yaml` file was created to set the tools we want for evals when none is specified

Additional features
1) Added model_list.yaml support for evals, you can now run the evals/benchmarks from model list
2) Added some code to support being able to set different reasoning models per model in model_list.yaml
3) Added docs for evaluating a new llm
4) Added a tag one-test for evaluating a singular test that should pass, updated it in docs and notion

### DIFF
--- a/docs/data-sources/.nav.yml
+++ b/docs/data-sources/.nav.yml
@@ -2,5 +2,5 @@ nav:
   - index.md
   - Built-in Toolsets: builtin-toolsets
   - Custom Toolsets: custom-toolsets.md
-  - Remote MCP Servers: remote-mcp-servers.md
+  - MCP Servers: remote-mcp-servers.md
   - Adding Permissions for Additional Resources: permissions.md
--- a/docs/data-sources/index.md
+++ b/docs/data-sources/index.md
@@ -14,7 +14,7 @@ HolmesGPT connects to your monitoring and observability tools to provide compreh
 
     Create your own integrations for proprietary or specialized tools.
 
--   **[Remote MCP Servers](remote-mcp-servers.md)**
+-   **[MCP Servers](remote-mcp-servers.md)**
 
     Connect to Model Context Protocol servers for extended capabilities.
 
--- a/docs/data-sources/remote-mcp-servers.md
+++ b/docs/data-sources/remote-mcp-servers.md
@@ -1,28 +1,36 @@
-# Remote MCP Servers
+# MCP Servers
 
-!!! warning
-    Remote MCP servers are in **Tech Preview** stage.
-
-!!! note "Configuration Change"
-    **The MCP server configuration format has been updated.** The `url` field must now be specified inside the `config` section instead of at the top level. The old format (with `url` at the top level) is still supported for backward compatibility but will log a migration warning. Please update your configurations to use the new format.
-
-!!! warning "SSE Mode Deprecation"
-    The SSE (Server-Sent Events) transport mode for MCP is being deprecated across the MCP ecosystem.
-    **We strongly recommend using `streamable-http` mode** for new MCP server integrations.
-    SSE mode support is maintained for backward compatibility but may be removed in future versions.
-
-HolmesGPT can integrate with remote MCP servers using either **streamable-http** (recommended) or **SSE** (deprecated) transport modes.
-This capability enables HolmesGPT to access external data sources and tools in real time.
-This guide provides step-by-step instructions for configuring HolmesGPT to connect with remote MCP servers.
+HolmesGPT can integrate with MCP (Model Context Protocol) servers to access external data sources and tools in real time. This guide provides step-by-step instructions for configuring HolmesGPT to connect with MCP servers.
 
 ## Transport Modes
 
-HolmesGPT supports two MCP transport modes:
+HolmesGPT supports three MCP transport modes:
 
 1. **`streamable-http`** (Recommended): Modern transport mode that uses HTTP POST requests with JSON responses. This is the preferred mode for new integrations.
-2. **`sse`** (Deprecated): Legacy transport mode using Server-Sent Events. Maintained for backward compatibility only.
+2. **`stdio`**: Direct process communication using standard input/output. Recommended for **Holmes CLI** usage only. For in-cluster deployments, see the [workaround using Supergateway](#working-with-stdio-mcp-servers-via-supergateway).
+3. **`sse`** (Deprecated): Legacy transport mode using Server-Sent Events. Maintained for backward compatibility only. We strongly recommend using `streamable-http` mode for new integrations.
 
-### Configuring Transport Mode
+## Configuration Structure
+
+**`mcp_servers` is a separate top-level key** in the configuration file, alongside `toolsets`. Both can coexist in the same config file:
+
+```yaml
+toolsets:
+  my_custom_toolset:
+    # ... toolset configuration
+
+mcp_servers:
+  my_mcp_server:
+    # ... MCP server configuration
+```
+
+Internally, MCP servers are treated as toolsets with `type: MCP` and are merged with other toolsets. This means MCP servers appear alongside regular toolsets in HolmesGPT's toolset list and can be enabled/disabled like any other toolset.
+
+## Streamable-HTTP (Recommended)
+
+Streamable-HTTP is the recommended transport mode for all new MCP server integrations. It uses HTTP POST requests with JSON responses and provides better compatibility and future-proofing.
+
+### Configuration
 
 The transport mode and URL are specified in the `config` section of your MCP server configuration:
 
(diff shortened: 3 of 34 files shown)


docs/development/evaluations/benchmarking-new-models.md
```diff
@@ -0,0 +1,112 @@
+# Benchmarking New Models
+
+This guide walks you through the process of benchmarking a new LLM model in HolmesGPT's evaluation framework.
+
+## Prerequisites
+
+- At least 4 nodes
+- Prometheus installed in cluster
```
Sheeproid: @Avi-Robusta can you explain that? why do we need this? (both to me personally and for the read)

docs/development/evaluations/benchmarking-new-models.md
```diff
@@ -0,0 +1,112 @@
...
+- Prometheus port-forwarded to 9090
+
+## Step 1: Create Model List File
+
+Create a YAML file listing all models you want to benchmark. For provider-specific configuration options, see the [AI Providers documentation](../../ai-providers/index.md).
+
+```yaml
+# Example: model_list_eval.yaml
+gpt-5.1:
+  api_key: "API_KEY_HERE"  # Use environment variable, not hardcoded secrets
```
Sheeproid: @Avi-Robusta can you show an example of using an env var? I'm not sure what is the syntax

docs/development/evaluations/benchmarking-new-models.md
```diff
@@ -0,0 +1,112 @@
...
+export MODEL_LIST_FILE_LOCATION=/path/to/your/model_list_eval.yaml
+```
+
+## Step 2: Run Initial Test
+
+Set required environment variables:
+
+```bash
+export MODEL="your-model-name"  # From your model list
+export RUN_LIVE=true
```
Sheeproid: I changed it to be default true now in master, so I don't think it needs setting

docs/development/evaluations/benchmarking-new-models.md
```diff
@@ -0,0 +1,112 @@
...
+```bash
+export MODEL_LIST_FILE_LOCATION=/path/to/your/model_list_eval.yaml
+```
+
+## Step 2: Run Initial Test
+
+Set required environment variables:
+
+```bash
+export MODEL="your-model-name"  # From your model list
```
Sheeproid, line 45: why do we need to set this in addition to the model list? I"m confused

docs/development/evaluations/benchmarking-new-models.md
```diff
@@ -0,0 +1,112 @@
...
+- Verify your Kubernetes cluster is accessible (if testing Kubernetes-related evals)
+- Check that all required toolsets are properly configured
+
+## Step 4: Run Benchmarks
+
+Run the benchmark script with your new model (along with other models you have configured in your model list):
+
+```bash
+unset MODEL # to be safe
+export RUN_LIVE=true
```
Sheeproid: again, no need to set RUN_LIVE as true anymore

docs/development/evaluations/benchmarking-new-models.md
```diff
@@ -0,0 +1,112 @@
...
+export CLASSIFIER_MODEL=gpt-4.1  # Use gpt-4.1 for consistent evaluation
+./run_benchmarks_local.sh 'your-new-model,gpt-4.1,gpt-4o,...'
```
Sheeproid, line 80: what is the base list that we need to test in Robusta? where is it defined? Notion?

tests/llm/fixtures/test_ask_holmes/09_crashpod/test_case.yaml
```diff
@@ -51,3 +51,4 @@ after_test: |
 tags:
   - easy
   - kubernetes
+  - one-test
```
Sheeproid, line 54: what is one-test tag?

holmes/core/llm.py
```diff
@@ -397,6 +397,12 @@ def completion(
                 "reasoning_effort"
             ]  # can be removed after next litelm version
 
+        existing_allowed = self.args.pop("allowed_openai_params", None)
```
Sheeproid, line 400: is this a must for gpt-5.1 or can we have a non code alternative?

tests/llm/utils/classifiers.py
```diff
@@ -19,43 +20,62 @@
 
 def create_llm_client():
     """Create OpenAI/Azure client with same logic used by tests"""
```
Sheeproid: there is a lot of if/else that does the same thing, can we consildate, for example let create_llm_client get parameters?

tests/llm/utils/default_toolsets.yaml
```diff
@@ -0,0 +1,24 @@
...
+    enabled: true
+  kubernetes/logs:
+    enabled: true
+  
+  # Helm toolset
+  helm/core:
+    enabled: true
+  
+  # Docker toolset
+  docker/core:
```
Sheeproid: @moshemorad can you weigh-in on the default list of tools here? I don't believe that 'docker' is a basic toolset that everyone should have. Same for helm.

The robusta toolset is based on issues on the platform and is useless for the testing, it keeps making the ai to do calls to get configuration changes and more details that are always missing, making the braintrust log clusttered with redundant tool calls and slowing down everything.

tests/llm/utils/mock_toolset.py
```diff
@@ -657,6 +657,13 @@ def _initialize_toolsets(self):
         config_path = os.path.join(self.test_case_folder, "toolsets.yaml")
         custom_definitions = self._load_custom_toolsets(config_path)
 
+        # If no test-specific toolsets.yaml exists, use default toolsets.yaml
+        if not custom_definitions:
```
Sheeproid: Let's say I have the runbook tool enabled, and the toolset.yaml does not mention it.
Should the runbook tool be enabled in that case?
What happens here? and what do you think the behavior should be?

tests/llm/conftest.py
```diff
@@ -391,12 +404,15 @@ def check_llm_api_with_test_call():
...
-            print(resp)
+            if llm:
+                llm.completion(messages=[{"role": "user", "content": "test"}])
+            else:
+                resp = litellm.completion(
+                    model=model_name,
+                    messages=[{"role": "user", "content": "test"}],
+                    max_tokens=1000,
+                )
+                print(resp)
```
Sheeproid: we can remove that print here, right?

tests/llm/conftest.py
```diff
@@ -451,6 +476,12 @@ def check_llm_api_with_test_call():
         error_msg += f"  - ANTHROPIC_API_KEY: {'set' if os.environ.get('ANTHROPIC_API_KEY') else 'not set'}\n"
         error_msg += f"  - AZURE_API_KEY: {'set' if os.environ.get('AZURE_API_KEY') else 'not set'}\n"
         error_msg += f"  - AZURE_API_BASE: {azure_base or 'not set'}\n"
+        # Show classifier model - if CLASSIFIER_MODEL env var is unset, show the actual value being used
+        classifier_env = os.environ.get("CLASSIFIER_MODEL")
```
Sheeproid: If we reuse the env vars in multiple places, I suggest moving them to a new file I created nambed env_var.py in holmes llm testing utils (there is one more files named that way)

docs/development/evaluations/benchmarking-new-models.md
Sheeproid: I saw it is defined in the script. in this case, we should indicate here that there is a default list so the user won't think this is a mandatory parameter

holmes/core/llm.py
Avi-Robusta: This is a must for azure if we want to test different models with reasoning against eachother. i.e. gpt-5 minimal against gpt-5.1 high
any non code alternative would be global across all models which wont work for our use case.
we could always make allowed_openai_params an env var but than we will have to fix it the next time an issue comes along or different models have different lists so its better to solve it completly now

tests/llm/fixtures/test_ask_holmes/09_crashpod/test_case.yaml
Avi-Robusta: to run a benchmark on tag one-test that is reliable fast single test that passes to verify the benchmarks/models is set up correctly.
documented in notion for our benchmarks
see more details here
https://github.com/HolmesGPT/holmesgpt/pull/1137/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

tests/llm/utils/default_toolsets.yaml
Avi-Robusta: I agree with docker.
helm is enabled in cluster, if its noisy and not so useful than we should remove the toolset and not just from the tests.
Any toolset that we enable by default in cluster should be enabled in tests IMO or else the eval isnt with realistic toolsets.

The robusta toolset is should be with a mocked supabase with fake files even when running live and many evals expect it is enabled for changes and runbooks issue etc so its intended to be used in tests

Are you sure its useless and throwing braintrust issues? Maybe we can make a ticket to fix the mock supabase so it throws less errors if you have examples?

tests/llm/utils/mock_toolset.py
Avi-Robusta: Not sure I understand your premis with 'i have runbook enabled but its not enabled in the toolset.yaml', how would it be enabled if its not enabled in toolset.yaml that was explicitly written in the test? Isnt writing the toolset.yaml to explicitly decide which tools to explicitly enable in our tests?

Avi-Robusta: I guess we could re-enable the default toolsets if theyre not explicitly disabled?

docs/development/evaluations/benchmarking-new-models.md
Avi-Robusta: I think you misunderstand what the model list and model param.

`Model list `is all the available models configured i.e. you have 20 models
when you run a robusta investigation in the UI you don't run it on all those models, you choose which `MODEL` to run on.

In our tests `MODEL` can be multiple models but we dont expect every one to be used. i.e. we have a gpt-4.1 in our model list but we dont want to run evals on it

tests/llm/utils/default_toolsets.yaml
Avi-Robusta: open ticket to disable to reduce noise

PR #1143, opened 2025-11-27
feat: add support for user runbooks catalog

michael-todorovic:
Hello,
This PR adds support for user runbooks catalog, without requiring a custom holmes build. This uses the same mechanisms than `holmes/plugins/runbooks`


moshemorad:
Hi @michael-todorovic can you check why the precommit is failed? Looks like there is some issue with types.

holmes/config.py
```diff
@@ -84,6 +84,7 @@ class Config(RobustaBaseConfig):
     opsgenie_query: Optional[str] = None
 
     custom_runbooks: List[FilePath] = []
+    custom_runbook_catalogs: List[Union[str, FilePath]] = []
```
moshemorad, line 87: What is the difference between custom_runbooks and custom_rubook_catalogs? can't we join those 2 vars and use them seemlessly?

holmes/config.py
```diff
@@ -114,6 +115,7 @@ def toolset_manager(self) -> ToolsetManager:
                 custom_toolsets=self.custom_toolsets,
                 custom_toolsets_from_cli=self.custom_toolsets_from_cli,
                 global_fast_model=self.fast_model,
+                custom_runbook_catalogs=self.custom_runbook_catalogs,
```
moshemorad, line 118: In the past we used to pass a lot of args from the config to the loading toolsets. We want to change it since it cause a lot of issues and a lot of places to change while adding new args. I think it would be much easier to support it as config for the toolset like we are doing with other toolsets. WDYT?

moshemorad (changes requested):

holmes/config.py
moshemorad: NVM i see now that this is different feature, can ignore.

holmes/config.py
moshemorad: If you think it will cause huge refactoring let me know and we might proceed as is and improve it later on.

PR #1150, opened 2025-12-01
Minor - cleanup of org name references

pavangudiwada:
Skipped these folders and files because changing them doesn't make sense
image

Updated https://deepwiki.com/HolmesGPT/holmesgpt. Repo indexing is in progress at the time of creation of this PR


aantn:
Thank you!

PR #1151, opened 2025-12-01
Updated text to clarify LLM connection issues

pavangudiwada:
Purely UX change
1. Instead of **logging "No llm models were loaded"**(missed in text) **-> loading Tools -> Raising exception**(image below), **raise an exception immediately**. 
image
2. Updated text to provide context on what could be the issue and how to fix it. 

## Result
image

### DIFF
--- a/holmes/config.py
+++ b/holmes/config.py
@@ -131,16 +131,6 @@ class Config(RobustaBaseConfig):
             self._llm_model_registry = LLMModelRegistry(self, dal=self.dal)
         return self._llm_model_registry
 
-    def log_useful_info(self):
-        if self.llm_model_registry.models:
-            logging.info(
-                f"Loaded models: {list(self.llm_model_registry.models.keys())}"
-            )
-        else:
-            raise Exception(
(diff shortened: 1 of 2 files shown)


moshemorad:
@pavangudiwada can you check tests please?

holmes/config.py
```diff
@@ -137,7 +137,9 @@ def log_useful_info(self):
                 f"Loaded models: {list(self.llm_model_registry.models.keys())}"
             )
         else:
-            logging.warning("No llm models were loaded")
+            raise Exception(
```
moshemorad: This called in both flows after init config is done. Can we consdier move this check there? Is there a reason to keep it in the log_useful_info and not move it to the llm_registry init (in the property?)

Also there are couple of tests failed since until now loading the config were finished succesfully even if no model were loaded. We need to handle those test flows.

moshemorad (changes requested):

holmes/config.py
pavangudiwada: Hi, don't think it's needed in two places.

PR #1154, opened 2025-12-01
Disable precommit checks

moshemorad:


mainred:
@moshemorad  why we disable the precommit check? I don't think the developers always run `make check` locally before committing a change.

mainred:
Per Moshe, "It was disabled since we saw it created friction for the new contributors.
Natan and I talked about it and decide to disable it for now until we will improve the devexp and the tutorial for new contributors."

PR #1156, opened 2025-12-02
[ROB-2556] added sentry message on failure

Avi-Robusta:
1) so we know how often these errors come up
2) We also are enabling structured output on sonnet 4 to also prevent this issue from occuring

### DIFF
--- a/holmes/utils/sentry_helper.py
+++ b/holmes/utils/sentry_helper.py
@@ -54,11 +54,10 @@ def capture_sections_none(content: Optional[str]):
         else:
             display_content = content
 
-    message = (
-        f"following message could not be broken into sections\n\n'{display_content}'"
-    )
-
-    sentry_sdk.capture_message(
-        message,
-        level="warning",
-    )
+    with sentry_sdk.push_scope() as scope:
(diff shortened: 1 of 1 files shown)


holmes/utils/sentry_helper.py
```diff
@@ -39,3 +41,24 @@ def capture_structured_output_incorrect_tool_call():
...
+    display_content = ""
+    if content:
+        if len(content) > _MAX_DISPLAY_LENGTH * 2:
+            # Show first and last portions of content
+            display_content = f"{content[:_MAX_DISPLAY_LENGTH]}...\n\n...{content[-_MAX_DISPLAY_LENGTH:]}"
+        else:
+            display_content = content
+
+    message = (
+        f"following message could not be broken into sections\n\n'{display_content}'"
```
moshemorad: @Avi-Robusta this will create different issue for each of the rows since the message is different. Let's have message to be something that is aligned across al of the messages and let's see if we can add the content as data to the event. 

e.g. something like "Holmes answer didn't contain sections" or anything that you think is better. 

WDYT?

PR #1170, opened 2025-12-06
Mcp docs

arikalon1:


docs/data-sources/builtin-toolsets/aws-mcp.md
```diff
@@ -0,0 +1,369 @@
...
+    Create a file named `aws-mcp-deployment.yaml`:
+
+    ```yaml
+    apiVersion: v1
+    kind: Namespace
+    metadata:
+      name: holmes-mcp
+    ---
+    apiVersion: v1
+    kind: ServiceAccount
```
moshemorad, line 26: Thinking here loudly, do we actually think that the best way to deploy mcps locally is k8s? why not just docker? or run it as process?

Im think here loud we can just open a ticket for future improvements.

moshemorad (approved):

PR #1175, opened 2025-12-08
fixes env vars to make eval work with openai models

zyxue:
it looks at its current status, eval only works with azure models.


moshemorad:
Hi @zyxue can you please sign you commit? we have dco enabled and it check it out.

For more instruction please click on the failing check.

zyxue:
@moshemorad , ah, signed

PR #1180, opened 2025-12-10
[ROB-2722] - Coralogix improvments

Avi-Robusta:


holmes/plugins/toolsets/coralogix/toolset_coralogix.py
```diff
@@ -0,0 +1,172 @@
...
+        if error:
+            return StructuredToolResult(
+                status=StructuredToolResultStatus.ERROR,
+                error=error,
+                params=params,
+            )
+
+        result_dict = {
+            "tool_name": self.name,
+            "query": params["query"],
```
RoiGlinik: This is to improve query handling?

the params Inside the structure output format does not give similar results?
Also you could use NO_DATA status for the no result case for the tool

holmes/plugins/toolsets/coralogix/toolset_coralogix.py
```diff
@@ -0,0 +1,172 @@
...
+            try:
+                tier = CoralogixTier[tier_str]
+            except KeyError:
+                return StructuredToolResult(
+                    status=StructuredToolResultStatus.ERROR,
+                    error=f"Invalid tier '{tier_str}'. Must be 'FREQUENT_SEARCH' or 'ARCHIVE'",
+                    params=params,
+                )
+
+        start_time = normalize_datetime(params.get("start_date"))
```
RoiGlinik, line 85: Does  process_timestamps_to_rfc3339  can work here as well?  we use this function on some tools (prometheus tool as well) and it handles some edge cases. Can be updated in PR after this one, but its worth a check.

RoiGlinik (approved):

holmes/plugins/toolsets/coralogix/toolset_coralogix.py
Avi-Robusta: Ill handle it in a different PR

PR #1187, opened 2025-12-12
feat: add healthz and readyz endpoint for pod probe

mainred:
## Tests

- filter out healthz and readyz probe access log
```
kubectl logs -n holmesgpt holmesgpt-holmes-5b8878b7dc-mngrl  -f
setting up colored logging
2025-12-12 06:24:30.066 INFO     logger initialized using INFO log level
2025-12-12 06:24:30.066 INFO     No robusta config in /etc/robusta/config/active_playbooks.yaml
2025-12-12 06:24:30.067 INFO     Loaded models: ['azure/gpt-4.1']
2025-12-12 06:24:30.068 INFO     Skipping sentry initialization - not an official release and DEVELOPMENT_MODE not enabled
2025-12-12 06:24:30.073 INFO     Skipping holmes status and toolsets synchronization - not connected to Robusta platform
2025-12-12 06:24:30,114 INFO     Started server process [1]
2025-12-12 06:24:30,115 INFO     Waiting for application startup.
2025-12-12 06:24:30,115 INFO     Application startup complete.
2025-12-12 06:24:30,116 INFO     Uvicorn running on http://0.0.0.0:5050 (Press CTRL+C to quit)
```

- api chat call
```
 curl -X POST http://localhost:8080/api/chat \
  -H "Content-Type: application/json" \
  -d '{"ask": "list pods in namespace default?"}'
{"analysis":"Pods in namespace default:\n- alertmanage...
```


server.py
```diff
@@ -415,6 +415,36 @@ def get_model():
...
+def readiness_check():
+    try:
+        models_list = config.get_models_list()
+        return {"status": "ready", "models": models_list}
+    except Exception as e:
+        logging.error(f"Readiness check failed: {e}", exc_info=True)
+        raise HTTPException(status_code=503, detail="Service not ready")
+
+
+class EndpointFilter(logging.Filter):
```
moshemorad: Please move this class to different file something like logging_utils.

server.py
```diff
@@ -415,6 +415,36 @@ def get_model():
...
+    def __init__(self, path: str, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self._path = path
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        return record.getMessage().find(self._path) == -1
+
+
+# Filter out periodical healniss and readiness probe.
+uvicorn_logger = logging.getLogger("uvicorn.access")
```
moshemorad: Move to init_logging function

helm/holmes/templates/holmes.yaml
```diff
@@ -43,6 +43,25 @@ spec:
         image: "{{ .Values.registry }}/{{ .Values.image }}"
         imagePullPolicy: Always
         command: ["python3", "-u", "server.py"]
+        ports:
+        - containerPort: 5050
```
moshemorad, line 47: Holmes port could be changed using `HOLMES_PORT` env var. In that case it will break. Please fix.

mainred: What's the point of using HOLMES_PORT to customize the port?
In the case of helm charts, we deploy holmes as a non-hostnetwork pod, so the ip address and then ip:port are unique across the cluster, which says the port is local the holmes deployment.

mainred: After searching HOLMES_PORT in the repo, I found HOLMES_PORT is used in several places related ag-ui. Is ag-ui deployed as a pod? If so, we can hardcode the port of ag-ui as a deployment?

moshemorad: ag-ui is something experimental that added by community user. I don't think it is matter in that case.

I do worried that for some reason it would be change and broke it, if you think it is far fetched we can keep it as it is for now.

mainred: I think it's enough to use a local hard-coded port for helm charts.
If the user wants to run the server locally, they can customize the port to avoid port in use issue.
Or if we want to bring flexibility to the port, we can expose the port through the helm values. But it brings almost no value.

mainred:
@moshemorad I think all comments so far have been addressed.

moshemorad (approved):

PR #1189, opened 2025-12-14
ROB-2713 improve healthcheck for tempo and  loki

RoiGlinik:


holmes/plugins/toolsets/grafana/loki/toolset_grafana_loki.py
```diff
@@ -29,6 +29,26 @@
...
+        _ = execute_loki_query(
+            base_url=get_base_url(c),
+            api_key=c.api_key,
+            headers=c.headers,
+            query='{job="test_endpoint"}',
+            start=start,
+            end=end,
+            limit=1,
+        )
+        return True, ""
```
moshemorad, line 91: the execute loki query will result in exception if failed (due to the raise_on_error) we just want it to bubble up to the try catch in the BaseGrafanaToolset pre requisite check and have that genric message?  

IMO it would be better to catch it here and handle it properly with more relevant message. WDYT?

holmes/plugins/toolsets/grafana/base_grafana_toolset.py
```diff
@@ -39,12 +39,19 @@ def prerequisites_callable(self, config: dict[str, Any]) -> Tuple[bool, str]:
...
         except Exception as e:
             logging.exception(f"Failed to set up grafana toolset {self.name}")
             return False, str(e)
 
+    def health_check(self) -> Tuple[bool, str]:
+        """
+        Base function to check if toolset is healthy.
+        :return: Raise for errors or return False, "error message" for success True, ""
+        """
+        return grafana_health_check(self._grafana_config)
```
moshemorad: Are we sure we want to keep it as default/fallback? I think we already saw it isn't working as expected and when we keep it we also need to keep the health check in the config which just confuse users. WDYT? can we remove this fallback and remove the field from the config? how it will effect user with it already configured?

tests/plugins/toolsets/grafana/test_grafana_loki.py
```diff
@@ -1,66 +1,33 @@
...
+    }
+    toolset.check_prerequisites()
 
-def test_grafana_loki_health_check(loki_config):
-    success, error_message = grafana_health_check(loki_config)
-
-    assert not error_message
-    assert success
+    assert toolset.error is None
+    assert toolset.status == ToolsetStatusEnum.ENABLED
```
moshemorad: Please add test also for failed health check.

tests/plugins/toolsets/grafana/test_grafana_tempo.py
```diff
@@ -53,48 +53,45 @@ def test_process_trace_json():
...
-        url=GRAFANA_URL,
-        grafana_datasource_uid=GRAFANA_TEMPO_DATASOURCE_UID,
-    )
+    def mock_get_raises(*args, **kwargs):
+        nonlocal call_count
+        call_count += 1
+        raise requests.exceptions.ConnectionError("Connection refused")
 
-    success, error_message = grafana_health_check(config)
+    with patch("requests.get", side_effect=mock_get_raises):
```
moshemorad: Can we use responses instead? https://github.com/getsentry/responses?tab=readme-ov-file#exception-as-response-body

IMO the way we currently keep patching functions cause a lot of harm since we are patching the functions we implemented and hence we aren't checking their logic (it doesn't the case with the current test but i think we should start and work more with responses instead of mock functions)

moshemorad (changes requested):

moshemorad (approved):

PR #1190, opened 2025-12-15
[ROB-2769] Links toolsets

Avi-Robusta:
Added links and tests for tempo, loki, grafana dashboards

### DIFF
--- a/holmes/plugins/toolsets/grafana/loki/toolset_grafana_loki.py
+++ b/holmes/plugins/toolsets/grafana/loki/toolset_grafana_loki.py
@@ -31,6 +31,8 @@ from holmes.core.tools import StructuredToolResult, StructuredToolResultStatus
 def _build_grafana_loki_explore_url(
     config: GrafanaConfig, query: str, start: str, end: str, limit: int = 100
 ) -> Optional[str]:
+    if not config.grafana_datasource_uid:
+        return None
     try:
         base_url = config.external_url or config.url
         datasource_uid = config.grafana_datasource_uid or "loki"
--- a/holmes/plugins/toolsets/grafana/toolset_grafana.py
+++ b/holmes/plugins/toolsets/grafana/toolset_grafana.py
@@ -1,6 +1,6 @@
 import os
-from typing import ClassVar, Dict, List, Optional, Type, cast
-from urllib.parse import urljoin, quote
+from typing import Any, ClassVar, Dict, Optional, Type, cast
+from urllib.parse import urlencode, urljoin
 from abc import ABC
 from holmes.core.tools import (
     StructuredToolResult,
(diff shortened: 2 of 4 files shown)


holmes/plugins/toolsets/grafana/loki/toolset_grafana_loki.py
```diff
@@ -85,28 +121,38 @@ def _invoke(self, params: dict, context: ToolInvokeContext) -> StructuredToolRes
...
-                query=params.get(
-                    "query", '{query="no_query_fallback"}'
-                ),  # make sure a string returns. fall back to query that will return nothing.
+                query=query_str,
                 start=start,
                 end=end,
                 limit=params.get("limit") or DEFAULT_LOG_LIMIT,
             )
+
+            explore_url = _build_grafana_loki_explore_url(
```
RoiGlinik, line 138: Does it work for both cases when you use Loki directly or when using with grafana proxy ( the datasource_uid case)? 
There is some hidden logic in base_url=get_base_url(config) ( which is bad) but handles url for both these cases?

Just wondering

holmes/plugins/toolsets/grafana/toolset_grafana.py
```diff
@@ -26,6 +26,26 @@ class GrafanaDashboardConfig(GrafanaConfig):
...
+    uid: Optional[str] = None,
+    query_params: Optional[List[str]] = None,
+) -> Optional[str]:
+    try:
+        base_url = config.external_url or config.url
+        if uid:
+            return f"{base_url.rstrip('/')}/d/{uid}"
+        else:
+            # Build search page URL
+            query_string = "&".join(query_params) if query_params else ""
```
RoiGlinik: In general I think using  using  urllib.parse import urlencode would be a bit better then doing this manually and quoting the fields manually on line 176.

holmes/plugins/toolsets/grafana/toolset_grafana.py
```diff
@@ -209,7 +250,18 @@ def __init__(self, toolset: GrafanaToolset):
 
     def _invoke(self, params: dict, context: ToolInvokeContext) -> StructuredToolResult:
         uid = params["uid"]
-        return self._make_grafana_request(f"/api/dashboards/uid/{uid}", params)
+        result = self._make_grafana_request(f"/api/dashboards/uid/{uid}", params)
```
RoiGlinik, line 248: Whats the difference between the 2 urls in that case. It used to return a URL before .

It also uses the get_base_url in make_grafana_request just want to make sure it does not effect the new dashboard_url case

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
```diff
@@ -32,6 +36,103 @@
...
+    start: Optional[int] = None,
+    end: Optional[int] = None,
+    limit: int = 20,
+    trace_id: Optional[str] = None,
+    filters: Optional[List[Dict[str, Any]]] = None,
+    tags: Optional[str] = None,
+) -> Optional[str]:
+    try:
+        base_url = config.external_url or config.url
+        datasource_uid = config.grafana_datasource_uid
```
RoiGlinik, line 53: Also here not sure about the case when using tempo directly and not grafana proxy with datasource_uid.

holmes/plugins/toolsets/grafana/toolset_grafana.py
Avi-Robusta: No `api` in the url passed in the url param since its for users
```
grafana_base/api/dashboards/uid...
grafana_base/dashboards/uid...
```

holmes/plugins/toolsets/grafana/loki/toolset_grafana_loki.py
Avi-Robusta: not making url now if tempo/loki directly

holmes/plugins/toolsets/grafana/toolset_grafana_tempo.py
Avi-Robusta: run all tools for tempo until they all succeed with a result

PR #1193, opened 2025-12-15
ROB-2758 healthcheck prom mimir

RoiGlinik:


holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -54,7 +54,7 @@
 class PrometheusConfig(BaseModel):
     # URL is optional because it can be set with an env var
     prometheus_url: Optional[str]
-    healthcheck: str = "-/healthy"
+    healthcheck: str = "api/v1/query?query=up"
```
moshemorad: Do we really still need to keep this config?

tests/plugins/toolsets/grafana/conftest.py
```diff
@@ -10,7 +10,7 @@ def is_service_running(host: str, port: int, timeout: float = 1.0) -> bool:
         return False
 
 
-def check_grafana_running():
-    if is_service_running("localhost", 3000):
+def check_grafana_running(port: int = 3000):
```
moshemorad: It isn't checking only grafana now right? let's change the name sincew it is very confusing.

tests/plugins/toolsets/prometheus/test_prometheus.py
```diff
@@ -0,0 +1,32 @@
...
+    toolset.config = {
+        "prometheus_url": "http://localhost:9000/api/datasources/proxy/uid/PAE45454D0EDB9216",
+    }
+    toolset.check_prerequisites()
+
+    assert toolset.error is None
+    assert toolset.status == ToolsetStatusEnum.ENABLED
+
+
+def test_mimir_direct_toolset_health_check():
```
moshemorad, line 37: Maybe add test also for failure?

PR #1206, opened 2025-12-18
ROB-2062 disable sentry dsn for self host environments 

RoiGlinik:
## Summary by CodeRabbit

* **Chores**
  * Self-hosted deployments will now automatically clear the Sentry DSN when the deployment appears self-hosted and a default DSN value is present, effectively disabling Sentry error tracking in those setups.
  * SENTRY_DSN resolution has been centralized to ensure consistent behavior across Helm renders.


helm/holmes/templates/_helpers.tpl
```diff
@@ -10,3 +10,22 @@ Return the service account name to use
...
+{{- $isSelfHosted := false -}}
+{{- range .Values.additionalEnvVars -}}
+{{- if eq .name "ROBUSTA_API_ENDPOINT" -}}
+{{- if ne .value "https://stg.api.robusta.dev" -}}{{- $isSelfHosted = true -}}{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- $isSelfHosted -}}
+{{- end -}}
+
+{{- define "holmes.isDefaultSentryDSN" -}}
```
moshemorad, line 25: Can you explain what is the use case with this default sentry dsn?

helm/holmes/templates/_helpers.tpl
```diff
@@ -10,3 +10,22 @@ Return the service account name to use
 default
 {{- end -}}
 {{- end -}}
+
+
+{{- define "holmes.isSelfHosted" -}}
+{{- $isSelfHosted := false -}}
+{{- range .Values.additionalEnvVars -}}
+{{- if eq .name "ROBUSTA_API_ENDPOINT" -}}
+{{- if ne .value "https://stg.api.robusta.dev" -}}{{- $isSelfHosted = true -}}{{- end -}}
```
moshemorad, line 19: Shouldn't this be api.robusta.dev? 

Also i think that we are doing it very specific. 

maybe we can change the terminology to make it more general.

e.g. isConnectedToRobustaSaas for example. The leap step that we are taking from not connected to robusta saas to assume it self host is ok but not so clear.

Also in that case it will stop enable sentry also for stg (after the fix) right? 

It will also break the new resion saas or not?

helm/holmes/templates/_helpers.tpl
```diff
@@ -10,3 +10,22 @@ Return the service account name to use
...
+{{- end -}}
+{{- end -}}
+{{- $isSelfHosted -}}
+{{- end -}}
+
+{{- define "holmes.isDefaultSentryDSN" -}}
+{{- eq .Values.sentryDSN "https://51f9cd9bd2fdee16144db08fc423cd3b@o1120648.ingest.us.sentry.io/4508799804702720" -}}
+{{- end -}}
+
+{{- define "holmes.sentryDSN" -}}
```
moshemorad, line 29: Shouldn't we also disable it in runner + relay? Other places we are missing?

moshemorad (changes requested):

PR #1208, opened 2025-12-19
feat: Add flux to built-in toolsets

shlok007:
## Summary by CodeRabbit

* **New Features**
  * Added a Flux CD toolset exposing read-only Flux CLIs for health, status, sources, kustomizations, helmreleases, diffs, exports, tracing, and debugging aids.

* **Documentation**
  * Added comprehensive Flux CD docs (prereqs, config snippets, examples, capabilities table, troubleshooting) and integrated into site navigation and toolset index.

* **Tests**
  * Added unit tests and fixtures covering Flux command parsing, safety rules, edge cases, and representative Flux outputs.

### DIFF
--- a/docs/data-sources/builtin-toolsets/.nav.yml
+++ b/docs/data-sources/builtin-toolsets/.nav.yml
@@ -12,6 +12,7 @@ nav:
   - DataDog: datadog.md
   - Datetime: datetime.md
   - Docker: docker.md
+  - Flux CD: flux.md
   - GitHub: github.md
   - Grafana Dashboards: grafanadashboards.md
   - Loki: grafanaloki.md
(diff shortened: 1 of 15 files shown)


aantn:
Hi @shlok007, thank you for the PR.
You've added the flux toolset in two separate ways in this PR - once as a YAML toolset and once via the bash toolset as a whitelisted command. You don't need both of these! I would recommend:

1. Doing this only as a YAML toolset
2. Reducing the number of tools to a much smaller number of more powerful tools. e.g. flux_get_sources_all, flux_get_sources_git, flux_get_sources_helm, etc could all be expressed as a single flux_get command with a kind parameter

Finally, eval 176 won't actually pass - you don't actually setup the test scenario in the cluster. There are also a bunch of unrelated files. 

I strongly suggest reading a few evals first to understand the format, then writing an eval of your own, then trying to make sure it passes reliably.

PR #1214, opened 2025-12-22
Pass tool context to MCP server tools via _tool_context argument field

gossion:
##  Summary

  This PR implements passing ToolInvokeContext to MCP servers using a reserved _tool_context field in tool arguments, enabling MCP servers to access execution context information such as tool number, user approval status, and token limits, etc.
  - Structure: {"_tool_context": {"user_approved": "..."}}

##  Motivation

  MCP servers need access to the tool invocation context to make informed decisions. For example:
  - Adjusting behavior based on user approval status
  - Respecting token limits when generating responses
  - Tracking tool execution sequences via tool numbers
  - Accessing tool call metadata for logging/debugging

##  Configuration for what to pass

  Users can customize which fields to pass via config:

  ```yaml
  mcp_servers:
    my_mcp_server:
      config:
        url: http://example.com:8000/mcp
        mode: streamable-http
        context_fields: ["tool_name", "tool_number", "user_approved"]
   ```

  Design decisions:
  - When context_fields is not specified: no field is passed
  - When context_fields is specified: passes only those fields (full user control)
  - Allows arbitrary field names for future extensibility without code changes

## Summary by CodeRabbit

* **New Features**
  * Per-invocation context injection for MCP requests: selectable context fields are serialized and sent with requests; invalid or non-serializable fields produce warnings.
  * Example configuration now exposes configurable context fields.

* **Tests**
  * Added tests for context serialization, selective field inclusion, injection behavior, and config loading.


holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -129,6 +141,34 @@ def _invoke(self, params: dict, context: ToolInvokeContext) -> StructuredToolRes
...
+        """Inject _tool_context into params for universal transport support (stdio/SSE/http)."""
+        context_fields = self.toolset.get_context_fields()
+        context_dict = self._serialize_context(context, context_fields)
+
+        if not context_dict:
+            return params
+
+        # Create a copy of params and inject _tool_context
+        params_with_context = params.copy()
+        params_with_context["_tool_context"] = context_dict
```
mainred, line 155: _tool_context parameter is not predefined mcp tool call convention, right? It's specific to the mcp server

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -119,8 +127,12 @@ def _invoke(self, params: dict, context: ToolInvokeContext) -> StructuredToolRes
                 raise ValueError("MCP config not initialized")
 
             lock = get_server_lock(str(self.toolset._mcp_config.get_lock_string()))
+
+            # Inject context into params instead of headers for universal transport support
+            params_with_context = self._inject_tool_context(params, context)
+
             with lock:
-                return asyncio.run(self._invoke_async(params))
+                return asyncio.run(self._invoke_async(params_with_context, additional_headers=None))
```
mainred: in L114, we have another tool invoke context, should we consider merging the context? I think we don't have to, since the two contexts are different. The ToolInvokeContext in for internal use and may not be useful in the mcp call, while the context your introduced is defined by the user, and will function in the mcp call

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -290,5 +339,6 @@ def get_example_config(self) -> Dict[str, Any]:
             url=AnyUrl("http://example.com:8000/mcp/messages"),
             mode=MCPMode.STREAMABLE_HTTP,
             headers={"Authorization": "Bearer YOUR_TOKEN"},
+            context_fields=["tool_name", "tool_number", "request_context"],
```
mainred: In holmesgpt, we can leverage BaseModel to init a struct, so that we don't need _serialize_context.
I think we can remove the duplicated field that are defined already in ToolInvokeContext, basically tool_name and tool_number, which are for internal use for CLI experience.

holmes/plugins/toolsets/mcp/toolset_mcp.py
gossion: right, it is outside of mcp protocol. mcp doesn't define the request context, but the context is very important for many cases.

holmes/plugins/toolsets/mcp/toolset_mcp.py
gossion: the context in params_with_context is generated from ToolInvokeContext . currently ToolInvokeContext  doesn't have user context yet, I will have another PR to introduce user input context into ToolInvokeContext .

holmes/plugins/toolsets/mcp/toolset_mcp.py
gossion: removed the fields.

mainred:
Thanks for explanation. It looks good, let's wait for the Robusta team to take a look.

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -119,8 +127,12 @@ def _invoke(self, params: dict, context: ToolInvokeContext) -> StructuredToolRes
                 raise ValueError("MCP config not initialized")
 
             lock = get_server_lock(str(self.toolset._mcp_config.get_lock_string()))
+
+            # Inject context into params instead of headers for universal transport support
+            params_with_context = self._inject_tool_context(params, context)
```
moshemorad, line 133: Not sure i fully understand what is the use case. Can you please elaborate how it is going to work?

The mcp isn't familiar with the tool_context param we are passing so not sure what should happen there. 

What will happen if we pass this param to unsupported mcp server tool call?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -129,6 +141,34 @@ def _invoke(self, params: dict, context: ToolInvokeContext) -> StructuredToolRes
...
+
+        if not context_dict:
+            return params
+
+        # Create a copy of params and inject _tool_context
+        params_with_context = params.copy()
+        params_with_context["_tool_context"] = context_dict
+        return params_with_context
+
+    def _serialize_context(
```
moshemorad, line 158: Please extract it to helper function.

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -66,7 +64,9 @@ def get_lock_string(self) -> str:
 
 
 @asynccontextmanager
-async def get_initialized_mcp_session(toolset: "RemoteMCPToolset"):
+async def get_initialized_mcp_session(
+    toolset: "RemoteMCPToolset", additional_headers: Optional[Dict[str, str]] = None
```
moshemorad: Can you please explain the usage of additional_headers? I don't sure i understand how it is being used, all places either not pass it or put there None. Am i missing something?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -203,6 +247,11 @@ class RemoteMCPToolset(Toolset):
     icon_url: str = "https://registry.npmmirror.com/@lobehub/icons-static-png/1.46.0/files/light/mcp.png"
     _mcp_config: Optional[Union[MCPConfig, StdioMCPConfig]] = None
 
+    def get_context_fields(self) -> Optional[List[str]]:
+        if isinstance(self._mcp_config, MCPConfig):
```
moshemorad, line 264: We intentionally didn't support stdio mcp?

moshemorad (changes requested):
Thanks for your contribution.

Left couple of comments.

Im not sure i fully understand the use case and how this should work. 
If you can provide couple of examples it could be very helpful.

Thanks

holmes/plugins/toolsets/mcp/toolset_mcp.py
gossion: additional_headers are not used anymore, I removed them. The story is: originally I tried to pass the ctx into the request header, but realized that there is no header available for stdio, and then I change to use the params_with_context. by using params_with_context, we can support ctx for stdio/sse/http, no additional_header is needed.

holmes/plugins/toolsets/mcp/toolset_mcp.py
gossion: thanks for reviewing the PR. Here is the use case and why:
1. mcp provides the tool calling schema, but doesn't have a protocol to define a field for the context of the tools which doesn't be a part of input by llm. for example, to debug problem of cluster_a using 'call_kubectl' tool:   llm has ability to execute 'call_kubectl', but infomation/metadata/auth of cluster_a is handled by context of toolcalling directly. llm doesn't need to know information/metadata/auth for several reasons:  security, multi-tenant, etc. mcp tool combines the llm input 'call_kubectl' + cluster_a info/meta/auth to do kubectl command for *cluster_a* (flow: get kubeconfig of cluster_a -> call_kubectl with kubeconfig -> return result). With the context, we can expand the capability of mcp tools easily.
2. you are right, 'tool_context ' is not part of mcp protocol, but it is very useful, like what you have done for other built-in tools in holmesgpt.  mcp only cares the 'name' and 'arguments', and ‘_tool_context’ is ignored normally. 
3. In this PR, _tool_context is not injected by default, so the origin behavior is not changed: if context_fields is not defined in mcp_config, no context will be passed to mcp.
4. if we pass this param to unsupported mcp server tool call unexpectedly, the ‘_tool_context’  would be skipped normally.

holmes/plugins/toolsets/mcp/toolset_mcp.py
gossion: Thanks for pointing this out, I didn't realize that stdio has a different config class. Fixed it.

holmes/plugins/toolsets/mcp/toolset_mcp.py
gossion: sorry, I don't get it. where the code should go?

gossion:
@moshemorad @arikalon1 can you take another look?

gossion:
ping @moshemorad @arikalon1

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -28,6 +28,16 @@
 _server_locks: Dict[str, threading.Lock] = {}
 _locks_lock = threading.Lock()
 
+# Whitelist of JSON-serializable context fields from ToolInvokeContext
+# Excludes 'llm' (non-serializable) and other complex objects
+SERIALIZABLE_CONTEXT_FIELDS = {
```
gossion, line 33: ‘user_request' needs to be added.

PR #1316, opened 2026-01-04
ROB-2572 new relic multiple accounts

RoiGlinik:
## Summary by CodeRabbit

* **New Features**
  * Multi-account New Relic support: enable multi-account mode, list organization accounts, and require/handle account_id for NRQL queries; explicit base URL handling.

* **Tests**
  * Added extensive fixtures and test cases covering multi-account flows, account discovery, and varied NRQL query scenarios.

* **Chores**
  * Updated tooling defaults (pre-commit Python version) and minor test import/type-annotation tweaks.

### DIFF
--- a/docs/data-sources/builtin-toolsets/aws.md
+++ b/docs/data-sources/builtin-toolsets/aws.md
@@ -312,6 +312,217 @@ aws iam attach-role-policy \
   --policy-arn arn:aws:iam::ACCOUNT_ID:policy/HolmesMCPReadOnly
 ```
 
+## Multi-Account Setup
+
+For scenarios where you need to access multiple AWS accounts from your EKS clusters, the AWS MCP server supports multi-account access using cross-account IAM roles and EKS token projection.
+
+### When to Use Multi-Account Setup
+
+- You have multiple AWS accounts (dev, staging, prod, etc.)
+- You want pods in any cluster to access resources in target accounts
+- You need centralized IAM role management across accounts
+- You're using AWS Organizations or multi-account architectures
+
+### How It Works
+
+When multi-account mode is enabled, the MCP server:
+
+1. Uses **EKS token projection** instead of IRSA (IAM Roles for Service Accounts)
+2. Mounts an `accounts.yaml` configuration file that defines target accounts and their IAM roles
+3. Uses `assume_role_with_web_identity` to assume roles in target accounts
+4. Allows the LLM to specify which account to use via the `--profile` flag
+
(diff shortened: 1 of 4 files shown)


holmes/plugins/toolsets/newrelic/newrelic.py
```diff
@@ -170,16 +187,70 @@ def get_parameterized_one_liner(self, params) -> str:
...
+        accounts = api.get_organization_accounts()
+
+        result_with_key = {
+            "accounts": accounts,
+            "total_count": len(accounts),
+            "is_eu": self._toolset.is_eu_datacenter,
+        }
+
+        # Build New Relic accounts URL
+        base_url = (
```
moshemorad: Please extract and reuse

holmes/plugins/toolsets/newrelic/newrelic.py
```diff
@@ -170,16 +187,70 @@ def get_parameterized_one_liner(self, params) -> str:
...
+            ),
+            parameters={},
+        )
+        self._toolset = toolset
+
+    def _invoke(self, params: dict, context: ToolInvokeContext) -> StructuredToolResult:
+        if not self._toolset.nr_api_key:
+            raise ValueError("NewRelic API key is not configured")
+
+        api = NewRelicAPI(
```
moshemorad: think about extract and reuse, we have the same code in the nrql query tool as well.

holmes/plugins/toolsets/newrelic/newrelic.py
```diff
@@ -170,16 +187,70 @@ def get_parameterized_one_liner(self, params) -> str:
         return f"{toolset_name_for_one_liner(self._toolset.name)}: Execute NRQL ({description})"
 
 
+class ListOrganizationAccounts(Tool):
```
moshemorad, line 181: Thinking here loudly. Do we really need it as tool call?
Why aren't we doing it as part of the callback and adding it to the prompt? In which case we think that it is better in tool than in prompt? 

I just don't think that this tool call should result in different data for each call and tring to think what would be better. 

Any way it is just a thought. If we think it might take more time to look into it than let's merge it as it is and get back to it later on.

RoiGlinik: The current implementation just fits our design. Its a interesting idea but I prefer to get back to it later.

Avi-Robusta (dismissed):

Avi-Robusta (approved):
code rabbit has a comment

PR #1324, opened 2026-01-05
Add support for eval additional system prompt URL

mershal:
## Summary by CodeRabbit

* **New Features**
  * Added a frontend chart-embed test case to validate namespace comparison embeds.

* **Tests**
  * Added a pytest marker to target frontend tests.
  * Introduced a test option to supply an external URL for an additional system prompt.
  * Test infrastructure now conditionally fetches and applies the external prompt only when frontend tests are detected and threads it through the frontend test flow.


tests/llm/conftest.py
```diff
@@ -49,6 +51,50 @@
...
+        data = response.json()
+        if isinstance(data, dict) and "additional_system_prompt" in data:
+            prompt = data["additional_system_prompt"]
+            if not isinstance(prompt, str):
+                raise ValueError("'additional_system_prompt' must be a string")
+            return prompt
+        if isinstance(data, str):
+            return data
+    except ValueError:
+        # Not JSON or malformed; fall back to raw text
```
Sheeproid: @mershal can we have a warning here at least?

tests/llm/conftest.py
```diff
@@ -94,6 +140,29 @@ def mock_generation_config(request):
...
+    if not should_fetch:
+        return None
+
+    with force_pytest_output(request):
+        url_type = "custom" if custom_url else "default"
+        print(f"\n📥 Fetching additional system prompt from {url_type} URL: {url}")
+
+    try:
+        return _fetch_additional_system_prompt(url)
+    except Exception as e:  # pragma: no cover - defensive error propagation
```
Sheeproid, line 161: Can you catch the relevant exceptions only?

tests/llm/conftest.py
```diff
@@ -49,6 +51,50 @@
...
+        if isinstance(data, str):
+            return data
+    except ValueError:
+        # Not JSON or malformed; fall back to raw text
+        pass
+
+    return response.text
+
+
+def _has_frontend_tests(session: pytest.Session) -> bool:
```
Sheeproid, line 88: I think relying on tags is confusing. The tags has another functions, which is to choose the tests to run. fetching the additional prompt this is an implicit behavior that was hard for me to spot at first glance.

I suggest implementing this as an additional property of the test_case.yaml

tests/llm/conftest.py
```diff
@@ -94,6 +140,29 @@ def mock_generation_config(request):
     return MockGenerationConfig(generate_mocks, regenerate_all_mocks, mode)
 
 
+@pytest.fixture(scope="session")
+def additional_system_prompt(request) -> Optional[str]:
```
Sheeproid, line 145: can you rename this to get_additional_system_prompt or something similar? it will make it easier to distinguish the function for the string

tests/llm/fixtures/test_ask_holmes/194_frontend_chart_embed/test_case.yaml
```diff
@@ -0,0 +1,20 @@
+user_prompt:
+  - Show me a chart embed comparing namespaces that start with app-194, such as app-194-kube and app-194-sphere
+
+expected_output:
+  - 'Includes a block embed such as <<{ "type": "chart", "data": JSON, "title": "namespace app-194 substring comparison" }>>'
```
Sheeproid, line 7: I think it is better to be more strict on the format. we had troubles before with the LLM not doing <<{ and }>> so we should specify here what the exact format is and what are the string values so the judge can make an informed descision

tests/llm/fixtures/test_ask_holmes/194_frontend_chart_embed/test_case.yaml
```diff
@@ -0,0 +1,20 @@
+user_prompt:
+  - Show me a chart embed comparing namespaces that start with app-194, such as app-194-kube and app-194-sphere
+
```
Sheeproid, line 5: there is a "description:" field that you can add that explains the goal of the test to us (not for the llm). can you add a short sentence please?

tests/llm/fixtures/test_ask_holmes/194_frontend_chart_embed/test_case.yaml
```diff
@@ -0,0 +1,20 @@
+user_prompt:
+  - Show me a chart embed comparing namespaces that start with app-194, such as app-194-kube and app-194-sphere
```
Sheeproid: what do you mean by comparing namespaces? it is a bit vague for me - is it resources? cpu consumption? . It is vague on purpose?

tests/llm/fixtures/test_ask_holmes/194_frontend_chart_embed/test_case.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - The embed type is exactly equal to "chart"
+  - The embed includes the count or percentage for each group
+  - The embed doesn't contain any line-break characters
+  - The `data` JSON doesn't contain fields that their values are js scriptable. It should only contain fixed values!
+
+tags:
+  - frontend
+
+before_test: |
+  kubectl create namespace app-194-kube || true
```
Sheeproid, line 18: related to https://github.com/HolmesGPT/holmesgpt/pull/1324/changes#r2661419308

Are those just empty namespaces? what we expect Holmes to do with them?

tests/llm/conftest.py
mershal: @Sheeproid I get what you're saying, but it actually better to keep it, as this is a fixture and being used as-is. If we rename to `get_additional...` it will look weird here:

Image

tests/llm/fixtures/test_ask_holmes/194_frontend_chart_embed/test_case.yaml
mershal: Yes, they're empty. I just want to have some dummy data that Holmes can generate graphs on.

In this case, compare number of namespaces starting with X with the count of the namespaces that don't.

It's a very silly comparison, but it's just a sanity test that makes sure the bare minimum of FE system prompt & charting works.

tests/llm/fixtures/test_ask_holmes/194_frontend_chart_embed/test_case.yaml
mershal: Absolutely

tests/llm/fixtures/test_ask_holmes/194_frontend_chart_embed/test_case.yaml
mershal: I'll explain it better

tests/llm/conftest.py
mershal: Actually, on a second thought, I preferred to be strict and fail the tests completely if we don't get the expected format.

tests/llm/conftest.py
mershal: Why? I still want to catch everything, say that there was a problem with additional system prompt, and raise

PR #1326, opened 2026-01-05
Minor toolset improvements

aantn:
* Update docs and example configs for grafana toolset to match implementation
* Add verify_ssl option to grafana
* Rename rabbitmq ssl-related option to be consistent (preserving backwards compat)


aantn:
/eval
filter: grafana

PR #1355, opened 2026-01-10
Claude/tool approval all tools l mw xe

arikalon1:
## Summary by CodeRabbit

* **New Features**
  * Approval workflow for operations that require explicit user authorization.
  * Tools can be marked restricted and enabled via runbook/authorization; tool lists refresh when runbook state changes.
  * YAML/config support for restriction and approval rules.
  * Bash commands perform safety validation and honor user approval before execution.

* **Other**
  * Interaction state is reset on empty user input to ensure fresh request state.


arikalon1:
/eval
filter: restricted_tool
iterations: 1

pyproject.toml
```diff
@@ -140,7 +140,8 @@ markers = [
     # - Tests core HolmesGPT functionality that should never regress
     # - Fast execution to avoid slowing CI/CD pipelines
     "regression: Critical regression tests that must always pass",
-    "benchmark: Challenging tests for evaluating and benchmarking model capabilities"
+    "benchmark: Challenging tests for evaluating and benchmarking model capabilities",
+    "tool-restriction: Tests for tool approval and restriction mechanisms"
```
Sheeproid: this 'tool-restriction' tag is a great idea but is not used

holmes/plugins/prompts/_general_instructions.jinja2
```diff
@@ -70,3 +70,7 @@
 You are able to make tool calls / function calls. Recognise when a tool has already been called and reuse its result.
 If a tool call returns nothing, modify the parameters as required instead of repeating the tool call.
 When searching for resources in specific namespaces, test a cluster level tool to find the resource(s) and identify what namespace they are part of.
+
+# Tool Approval
+
+Some tools may require user approval before execution. If a tool requires approval, go ahead and call it - you will be prompted for confirmation. This is normal and expected for potentially destructive operations.
```
Sheeproid: Is this general instruction needed?

holmes/core/tool_calling_llm.py
```diff
@@ -175,6 +175,19 @@ def __init__(
             Callable[[StructuredToolResult], tuple[bool, Optional[str]]]
         ] = None
 
+        # Restriction tracking - these are set per-conversation
+        self._runbook_in_use: bool = False
+        self._restricted_tools_enabled: bool = False
+
+    def reset_conversation_state(self) -> None:
```
Sheeproid: this function is unused

holmes/core/tool_calling_llm.py
```diff
@@ -175,6 +175,19 @@ def __init__(
...
+        """Reset conversation-level state for a new conversation."""
+        self._runbook_in_use = False
+        # Note: _restricted_tools_enabled is not reset here as it's set per-request
+
+    def set_restricted_tools_enabled(self, enabled: bool) -> None:
```
Sheeproid: this is not used?

holmes/core/tool_calling_llm.py
```diff
@@ -175,6 +175,19 @@ def __init__(
             Callable[[StructuredToolResult], tuple[bool, Optional[str]]]
         ] = None
 
+        # Restriction tracking - these are set per-conversation
+        self._runbook_in_use: bool = False
+        self._restricted_tools_enabled: bool = False
```
Sheeproid: _restricted_tools_enabled is always false?

holmes/core/tool_calling_llm.py
```diff
@@ -506,8 +537,20 @@ def _directly_invoke_tool_call(
...
             )
             tool_response = tool.invoke(tool_params, context=invoke_context)
+
+            # Track runbook usage - if fetch_runbook is called successfully,
+            # restricted tools become available for the rest of the conversation
+            if (
+                tool_name == "fetch_runbook"
+                and tool_response.status == StructuredToolResultStatus.SUCCESS
+            ):
+                self._runbook_in_use = True
```
Sheeproid, line 544: I think that if a user asks a followup question then this will we False. is that intentional? it was hard for me to understand the intent by reading the code.

holmes/core/tool_calling_llm.py
```diff
@@ -175,6 +175,19 @@ def __init__(
             Callable[[StructuredToolResult], tuple[bool, Optional[str]]]
         ] = None
 
+        # Restriction tracking - these are set per-conversation
+        self._runbook_in_use: bool = False
```
Sheeproid, line 178: Because this was hard for me to follow, I though that there is a bug because once it is set is can NEVER be set back to False.

I asked Claude to check and there is indeed a bug but just in the interactive CLI

  Server context (not affected):
  - A new ToolCallingLLM instance is created for each request (ai = config.create_toolcalling_llm(...) inside each endpoint)
  - Each conversation gets a fresh instance, so _runbook_in_use starts as False

  Interactive CLI context (affected):
  - A single ToolCallingLLM instance is created once and reused for the entire session
  - When user runs /clear (line 1188-1199 in interactive.py):
  messages = None  # Reset messages
  last_response = None
  all_tool_calls_history.clear()
  - But ai.reset_conversation_state() is never called
  - So _runbook_in_use stays True from the previous conversation

  The fix: Add ai.reset_conversation_state() to the /clear command handler, something like:

holmes/core/tool_calling_llm.py
```diff
@@ -506,8 +537,20 @@ def _directly_invoke_tool_call(
                 max_token_count=self.llm.get_max_token_count_for_single_tool(),
                 tool_name=tool_name,
                 tool_call_id=tool_call_id,
+                restricted_tools_enabled=self._restricted_tools_enabled,
+                runbook_in_use=self._runbook_in_use,
             )
             tool_response = tool.invoke(tool_params, context=invoke_context)
+
+            # Track runbook usage - if fetch_runbook is called successfully,
+            # restricted tools become available for the rest of the conversation
```
Sheeproid: I think 'restricted tools become available for the rest of the conversation' it is not true for follow-up chats? see other comment about it

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -192,6 +192,11 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
             if isinstance(args, list):
                 return f"gcloud {' '.join(str(arg) for arg in args)}"
 
+        if self.name and params and "args" in params:
```
Sheeproid, line 195: just to confirm - this is not related to the change of tool approvals, right?

holmes/core/tool_calling_llm.py
```diff
@@ -175,6 +175,19 @@ def __init__(
             Callable[[StructuredToolResult], tuple[bool, Optional[str]]]
         ] = None
 
+        # Restriction tracking - these are set per-conversation
+        self._runbook_in_use: bool = False
+        self._restricted_tools_enabled: bool = False
+
+    def reset_conversation_state(self) -> None:
+        """Reset conversation-level state for a new conversation."""
+        self._runbook_in_use = False
```
Sheeproid, line 184: perhaps "_runbook_in_use" should just be a local variable in the call and call_stream function

holmes/core/tools.py
```diff
@@ -157,6 +163,9 @@ class ToolInvokeContext(BaseModel):
     tool_call_id: str
     tool_name: str
 
+    restricted_tools_enabled: bool = False
```
Sheeproid: I think both of these variables are not used at all.

holmes/core/tools.py
```diff
@@ -214,9 +227,14 @@ def model_post_init(self, __context) -> None:
             self._transformer_instances = None
 
     def get_openai_format(self, target_model: str):
+        # Add [RESTRICTED] prefix if tool is restricted
+        description = self.description
+        if self._is_restricted():
+            description = f"[RESTRICTED] {description}"
```
Sheeproid: why do we add [RESTRICTED] to the tools format for the open-ai tool list?

holmes/core/tools.py
```diff
@@ -230,11 +248,24 @@ def invoke(
         logger.info(
             f"Running tool {tool_number_str}[bold]{self.name}[/bold]: {self.get_parameterized_one_liner(params)}"
         )
+
+        if not context.user_approved:
```
Sheeproid, line 245: please check the bash toolset with local holmes CLI if you haven't done so.
The way approval works for CLI is different from how it is implemented for the server(UI) flow and I want to make sure this does not break it.

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -192,6 +192,11 @@ def get_parameterized_one_liner(self, params: Dict) -> str:
...
+            args = params.get("args", [])
+            if isinstance(args, list):
+                return f"{self.name} {' '.join(str(arg) for arg in args)}"
```
Sheeproid, line 198: I only saw that the bash toolset implements 'requires_approval' but I did not see any tool having the 'approval_required_tools' defined for it. is it defined in some tool?

holmes/core/tools.py
```diff
@@ -249,6 +280,45 @@ def invoke(
         )
         return transformed_result
 
+    def _is_restricted(self) -> bool:
+        if self.restricted:
+            return True
+
+        toolset = getattr(self, "toolset", None)
```
Sheeproid, line 280: using getattr is so werid and unclear. I understand that 'Tool' does not have a 'toolset' class variable by default but this is super weird.

Sheeproid: also the fact that we need to go from the 'Tool' back to the 'Toolset' to understand that this tool requires approval or is restricted it kind of weird, I'd expect it to be defined in the tool itself

Sheeproid (dismissed):

holmes/plugins/toolsets/mcp/toolset_mcp.py
arikalon1: Semi related.
The tool description is what shown on the approval.
I had to improve the mcp tool description, so the approval will be displayed properly

holmes/plugins/toolsets/mcp/toolset_mcp.py
arikalon1: This is for mcp/yaml toolsets

pyproject.toml
arikalon1: yeah, I had tests but removed it. Fixing

holmes/core/tools.py
arikalon1: you're right, it's not required. Fixing

holmes/plugins/prompts/_general_instructions.jinja2
arikalon1: No, it's not. Removing it

holmes/core/tool_calling_llm.py
arikalon1: you're right. Removing this

holmes/core/tools.py
arikalon1: right again, removing both

holmes/core/tool_calling_llm.py
arikalon1: removing

holmes/core/tool_calling_llm.py
arikalon1: nice catch. Fixed by resetting runbook_in_use, every user interaction in the cli interactive loop

holmes/core/tools.py
arikalon1: From reading the code I think it works properly (the tool implementation is called)
I opened a task to validate it

PR #1360, opened 2026-01-13
Redesigned bash toolset that have a prefix based approvals

Sheeproid:
- Allows prefixes to be approved
- Comes with a built-in list of prefixes (for server mode)
- Upgraded CLI approval experience

## Release Notes

* **New Features**
  * Bash tool now enabled by default with simplified prefix-based command validation
  * Interactive approval flow for non-approved commands with option to save prefixes for future sessions
  * Session prefix memory persists approved command prefixes across conversations
  * Added `--bash-always-deny` and `--bash-always-allow` CLI flags for controlling bash command approval behavior
  * Configurable allow/deny lists for bash commands via toolset configuration

* **Improvements**
  * Enhanced command validation clarity with better error messaging for blocked operations
  * Reduced approval friction through prefix persistence

### DIFF
--- a/tests/llm/fixtures/test_ask_holmes/111_tool_hallucination/toolsets.yaml
+++ b/tests/llm/fixtures/test_ask_holmes/111_tool_hallucination/toolsets.yaml
@@ -7,8 +7,6 @@ toolsets:
     enabled: False
   kubernetes/kube-lineage-extras:
     enabled: False
-  aws/rds:
-    enabled: False
   bash:
     enabled: False
   argocd/core:
@@ -18,8 +16,6 @@ toolsets:
   confluence:
     enabled: False
   docker/core:
-    enabled: False
-  aws/security:
     enabled: False
   aks/core:
     enabled: False
--- a/tests/llm/fixtures/test_ask_holmes/158_slack_chat_correct_date/toolsets.yaml
+++ b/tests/llm/fixtures/test_ask_holmes/158_slack_chat_correct_date/toolsets.yaml
@@ -17,10 +17,6 @@ toolsets:
     enabled: false
   kafka/admin:
     enabled: false
-  aws/security:
-    enabled: false
-  aws/rds:
-    enabled: false
   aks/core:
     enabled: false
   kubernetes/live-metrics:
--- a/tests/llm/fixtures/test_ask_holmes/22_high_latency_dbi_down/test_case.yaml
+++ b/tests/llm/fixtures/test_ask_holmes/22_high_latency_dbi_down/test_case.yaml
@@ -1,4 +1,6 @@
 # for instructions, see "Medium Credentials for 22_high_latency_dbi_down" in 1Password
+skip: true
+skip_reason: "Test requires aws/rds toolset which no longer exists"
 user_prompt: "Why is there high latency with the customer-orders-for-medium deployment?"
 tags:
   - chain-of-causation
--- a/tests/llm/fixtures/test_ask_holmes/22_high_latency_dbi_down/toolsets.yaml
+++ b/tests/llm/fixtures/test_ask_holmes/22_high_latency_dbi_down/toolsets.yaml
@@ -7,5 +7,3 @@ toolsets:
     enabled: true
   kubernetes/logs:
     enabled: true
-  aws/rds:
-    enabled: true
--- a/tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/toolsets.yaml
+++ b/tests/llm/fixtures/test_ask_holmes/55_kafka_runbook/toolsets.yaml
@@ -14,10 +14,6 @@ toolsets:
         - name: "kafka"
           kafka_broker: "kafka:9092" # local DNS likely need to resolve kafka to localhost
 
-  aws/security:
-    enabled: false
-  aws/rds:
-    enabled: false
   aks/core:
     enabled: false
   kubernetes/live-metrics:
--- a/tests/llm/fixtures/test_ask_holmes/57_cluster_name_confusion/toolsets.yaml
+++ b/tests/llm/fixtures/test_ask_holmes/57_cluster_name_confusion/toolsets.yaml
@@ -9,10 +9,6 @@ toolsets:
     enabled: true
   kafka/admin:
     enabled: false
-  aws/security:
-    enabled: false
-  aws/rds:
-    enabled: false
   aks/core:
     enabled: false
   kubernetes/live-metrics:
--- a/tests/llm/fixtures/test_ask_holmes/89_runbook_missing_cloudwatch/toolsets.yaml
+++ b/tests/llm/fixtures/test_ask_holmes/89_runbook_missing_cloudwatch/toolsets.yaml
@@ -5,10 +5,3 @@ toolsets:
 
   kubernetes/logs:
     enabled: true
-
-  # Explicitly disable all AWS toolsets
-  aws/security:
-    enabled: false
-
-  aws/rds:
-    enabled: false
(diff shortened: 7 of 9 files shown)


Sheeproid:
/eval
markers: regression or benchmark
iterations: 5

Sheeproid:
/eval
markers: regression or benchmark
iterations: 5

docs/data-sources/builtin-toolsets/bash.md
```diff
@@ -1,91 +1,119 @@
...
-- Repository management
-- Release inspection
+    See [default_lists.py](https://github.com/HolmesGPT/holmesgpt/blob/master/holmes/plugins/toolsets/bash/common/default_lists.py) for the complete list.
 
-**ArgoCD**
+## Command Approval
 
-- Application management
-- Deployment status checking
+When Holmes tries to run a command not in your allow list, you'll see a prompt:
```
aantn: Prefixes?

holmes/core/tool_calling_llm.py
```diff
@@ -724,6 +820,33 @@ def _handle_tool_call_approval(
             tool_call_result.result.status = StructuredToolResultStatus.ERROR
             return tool_call_result
 
+        # Re-check if approval is still needed (prefix may have been approved by another tool call)
+        tool = self.tool_executor.get_tool_by_name(tool_call_result.tool_name)
+        if tool:
```
Avi-Robusta: can all the added code below be refactored to another function for readability, maybe check_if_tool_approved

holmes/interactive.py
```diff
@@ -618,19 +623,101 @@ def prompt_for_llm_sharing(
...
+    Run an inline menu with arrow key navigation.
+
+    Args:
+        options: List of option strings to display
+        console: Rich console for output
+
+    Returns:
+        Index of selected option (0-based), or None if cancelled
+    """
+    from prompt_toolkit.application import Application
```
Avi-Robusta: why import here and not at top of file?

Avi-Robusta: honestly this might be best in its own file

holmes/interactive.py
```diff
@@ -618,19 +623,101 @@ def prompt_for_llm_sharing(
...
+                lines.append(("bold", f"> {i + 1}. {option}\n"))
+            else:
+                lines.append(("", f"  {i + 1}. {option}\n"))
+        lines.append(("class:hint", "\nEsc to cancel"))
+        return lines
+
+    bindings = KeyBindings()
+
+    @bindings.add("up")
+    @bindings.add("k")
```
Avi-Robusta, line 653: why k and j for up and down?
ws maybe,  kj seems random, am i missing something?

holmes/interactive.py
```diff
@@ -618,19 +623,101 @@ def prompt_for_llm_sharing(
...
+    def _down(event):
+        selected[0] = (selected[0] + 1) % len(options)
+
+    @bindings.add("enter")
+    def _enter(event):
+        result[0] = selected[0]
+        event.app.exit()
+
+    @bindings.add("escape")
+    @bindings.add("c-c")
```
Avi-Robusta, line 668: what is c-c

holmes/interactive.py
```diff
@@ -639,28 +726,44 @@ def handle_tool_approval(
...
+        return True, None
+    else:  # No (option 3) or Cancelled (Esc) - prompt for optional feedback
+        temp_session = PromptSession(history=InMemoryHistory())  # type: ignore
         feedback_prompt = temp_session.prompt(
             [("class:prompt", "Optional feedback for the AI (press Enter to skip): ")],
             style=style,
         )
-
         feedback = feedback_prompt.strip() if feedback_prompt.strip() else None
         return False, feedback
```
Avi-Robusta, line 762: i know you didnt write this, but when is feedback used?

holmes/plugins/toolsets/bash/common/cli_prefixes.py
```diff
@@ -0,0 +1,89 @@
...
+    global _cli_mode_enabled
+    _cli_mode_enabled = True
+
+
+def is_cli_mode() -> bool:
+    """Check if CLI mode is enabled."""
+    return _cli_mode_enabled
+
+
+def load_cli_approved_prefixes() -> List[str]:
```
Avi-Robusta: would specifiy bash so people know context
```suggestion
def load_cli_bash_tools_approved_prefixes() -> List[str]:
```

holmes/plugins/toolsets/bash/common/cli_prefixes.py
```diff
@@ -0,0 +1,89 @@
...
+            with open(prefixes_file, "r") as f:
+                data = yaml.safe_load(f)
+                if isinstance(data, dict) and "approved_prefixes" in data:
+                    return data["approved_prefixes"]
+        except Exception as e:
+            logging.warning(f"Failed to load approved prefixes: {e}")
+    return []
+
+
+def save_cli_approved_prefixes(prefixes: List[str]) -> None:
```
Avi-Robusta:
```suggestion
def save_cli_bash_tools_approved_prefixes(prefixes: List[str]) -> None:
```

holmes/plugins/toolsets/bash/common/cli_prefixes.py
Avi-Robusta: we have a config file, model list file, cached tools file, and now a bash prefixes file
keeping track of these additional files we create is getting a bit out of hand 
can this be added/removed from the config file or toolset config file somehow? this seems like it can be part of the bash config instead of a whole new file 
maybe ask ai to see how it can be added via the config/cache  and see if its easy to do

holmes/plugins/toolsets/kubectl_run/validation.py
```diff
@@ -1,33 +1,35 @@
...
 def validate_image_and_commands(
-    image: str, container_command: str, config: Optional[BashExecutorConfig]
+    image: str, container_command: str, config: Optional[KubectlRunConfig]
 ) -> None:
     """
     Validate that the image is in the whitelist and commands are allowed.
     Raises ValueError if validation fails.
     """
-    if not config or not config.kubectl or not config.kubectl.allowed_images:
+    if not config or not config.allowed_images:
```
Avi-Robusta, line 19: did you verify this isnt breaking retroactivly?

tests/llm/utils/default_toolsets.yaml
```diff
@@ -20,3 +20,10 @@ toolsets:
     enabled: true
   connectivity_check:
     enabled: true
+  bash:
+    enabled: true
+    config:
+      include_default_allow_deny_list: true # this is off by default for CLI but mimics the default in the UI
```
Avi-Robusta, line 26: how does this work in evals? will it stop them half way when they have a bash command or always approve?

Sheeproid: In the eval it returns a 'fail' that the command is not approved and the ai needs to recover from it. when running the benchmarks, there was little to no effect as it is usually not using the bash toolset.

Maybe later this will become a problem and I'll need to build a smarter mechanims that tell the llm that the user is not going to approve because it is disabled so it shouldn't even try.

holmes/plugins/toolsets/bash/common/cli_prefixes.py
Sheeproid: I agree this is another file and it is not ideal
re. adding it to the bash config - it cannot work because this is a dynamic list and the user would expect the toolset config won't change

WDYT

holmes/interactive.py
Sheeproid: yeah I think this might be broken. I'm not sure if I broke it or it was broken
If you use the CLI and click 'Deny' the llm might miss this

I'll test it again

holmes/interactive.py
Sheeproid: @Avi-Robusta vim style binding

holmes/interactive.py
Sheeproid: @Avi-Robusta  control c

holmes/interactive.py
Sheeproid: tested it, it works

PR #1397, opened 2026-01-21
feat: support passing token to mcp in header

gossion:
## Summary by CodeRabbit

* **New Features**
  * Per-request context is now propagated across investigations, chats, streaming flows, and tool invocations.
  * Remote toolsets support templated extra-headers with env and request-context substitutions (case-insensitive).

* **Bug Fixes**
  * API endpoints map auth and rate-limit failures to consistent HTTP status codes.
  * Sensitive values in request context are redacted in dumps/logs.

* **Documentation**
  * Added docs for dynamic headers with request-context and examples.

* **Tests**
  * Expanded tests for header rendering, context passthrough, and header extraction.

### DIFF
--- a/tests/test_approval_workflow.py
+++ b/tests/test_approval_workflow.py
@@ -1,5 +1,5 @@
 import json
-from typing import List, Optional
+from typing import Any, Dict, List, Optional
 from unittest.mock import MagicMock, patch
 
 import pytest
@@ -139,6 +139,7 @@ def test_streaming_chat_approval_workflow_requires_approval(
         tool_call_id: str,
         tool_number: Optional[int] = None,
         session_approved_prefixes: Optional[List[str]] = None,
+        request_context: Optional[Dict[str, Any]] = None,
     ) -> StructuredToolResult:
         return StructuredToolResult(
             status=StructuredToolResultStatus.APPROVAL_REQUIRED,
@@ -266,7 +267,7 @@ def test_streaming_chat_approval_workflow_approve_and_execute(
 
     # Mock process_tool_decisions to simulate approval and execution
     ai.process_tool_decisions = MagicMock(
-        side_effect=lambda messages, tool_decisions: (
+        side_effect=lambda messages, tool_decisions, request_context=None: (
             messages
             + [
                 {
@@ -413,7 +414,7 @@ def test_streaming_chat_approval_workflow_reject_command(
 
     # Mock process_tool_decisions to simulate rejection
     ai.process_tool_decisions = MagicMock(
-        side_effect=lambda messages, tool_decisions: (
+        side_effect=lambda messages, tool_decisions, request_context=None: (
             messages
             + [
                 {
--- a/tests/test_mcp_toolset.py
+++ b/tests/test_mcp_toolset.py
@@ -1415,7 +1415,7 @@ class TestRequestContextPassthrough:
 
         captured_headers = None
 
-        def capture_sse_client_call(_url, headers, *, sse_read_timeout):
+        def capture_sse_client_call(_url, headers, *, sse_read_timeout, httpx_client_factory=None):
             nonlocal captured_headers
             captured_headers = headers
             return mock_client_context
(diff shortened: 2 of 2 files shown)


holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -46,7 +48,8 @@ class MCPMode(str, Enum):
 class MCPConfig(BaseModel):
     url: AnyUrl
     mode: MCPMode = MCPMode.SSE
-    headers: Optional[Dict[str, str]] = None
+    extra_headers: Optional[Dict[str, str]] = None
+    context_fields: Optional[List[str]] = None
```
gossion: not need?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -240,6 +251,119 @@ class RemoteMCPToolset(Toolset):
...
+                        final_headers[header_name] = rendered_value
+                except Exception as e:  # noqa: BLE001
+                    logging.warning(
+                        f"MCP toolset '{self.name}': Failed to render header template "
+                        f"'{header_name}': {e}"
+                    )
+
+        return final_headers if final_headers else None
+
+    def _render_template(
```
arikalon1, line 363: Is it possible to simplify the implementation to use jinja2 templates for rendering?
I think it will simplify a lot, and will be consistent with templates rendering on other places in Holmes.
something like:
```
from jinja2 import Template
import os

templates = {
    "Header-Name": "hardcoded value",
    "Header-Name-2": "{{ request_context.headers['foo'] }}",
    "Header-Name-3": "{{ env.CORALOGIX_API_KEY }}",
}

context = {
    "env": os.environ,  # or dict(os.environ)
    "request_context": {
        "headers": {"foo": "bar_value"}
    }
}

resolved = {}
for key, value in templates.items():
    resolved[key] = Template(value).render(context)

```

server.py
```diff
@@ -381,8 +391,28 @@ def already_answered(conversation_history: Optional[List[dict]]) -> bool:
...
+def extract_passthrough_headers(request: Request) -> dict:
+    """
+    Extract pass-through headers from the request, excluding sensitive auth headers.
+    These headers are forwarded to MCP servers for authentication and context.
+
+    Returns:
+        dict: {"headers": {"X-Foo-Bar": "...", "ABC": "...", ...}}
+    """
+    # Blacklist of headers that should NOT be passed through
+    BLOCKED_HEADERS = {"authorization", "cookie", "set-cookie"}
```
arikalon1: Can we make it configurable?
Maybe take it from an env var, with those as defaults.
In case someone do want to pass the authorization

gossion: good suggestion, thank you.

holmes/plugins/toolsets/mcp/toolset_mcp.py
gossion: thanks, this is really a good suggestion.

gossion:
had some conflicts, I fixup all commits and rebased it.

server.py
```diff
@@ -381,8 +391,36 @@ def already_answered(conversation_history: Optional[List[dict]]) -> bool:
     return False
 
 
+def extract_passthrough_headers(request: Request) -> dict:
+    """
+    Extract pass-through headers from the request, excluding sensitive auth headers.
```
mainred, line 396: But we want to pass the sensitive auth header info to mcp tools?

holmes/core/tools.py
```diff
@@ -170,6 +170,22 @@ class ToolInvokeContext(BaseModel):
     session_approved_prefixes: List[
         str
     ] = []  # Bash prefixes approved during this session
+    request_context: Optional[Dict[str, Any]] = None
+
+    def model_dump(self, **kwargs):
```
mainred, line 175: 👍

server.py
gossion: that is just to pass them to `request_context.headeres` which can be seen by holmes mcp toolset. We have `extra_headers` in MCP config to determine which KEY in `request_context.headers` will be passed to the real mcp tool call, by default `extra_headers` is empty, so event the senstive auth header can be seen by holmes mcp toolset, they are not passed to real mcp tool call.

PR #1430, opened 2026-01-27
Support using config file in server mode

gossion:
## Summary by CodeRabbit

* **New Features**
  * Added a standard server entry point and configuration initialization that loads from a default location or environment.
  * Startup now initializes background scheduled prompts processing before serving requests.

* **Improvements**
  * More robust startup sequencing with pre-start synchronization and readiness tasks.
  * Clearer startup and runtime logging for better observability.


server.py
```diff
@@ -95,10 +97,65 @@ def init_logging():
 
 if ENABLE_CONNECTION_KEEPALIVE:
     patch_socket_create_connection()
+# Initialize config and dal at module load time
+# Will be re-initialized in main() if --config is provided
```
mainred: I think we can try to organize the configuration with the config file, and migrating the existing configure from env to the configuration file backend by a configure map.
what do you think? @arikalon1

server.py
```diff
@@ -95,10 +97,65 @@ def init_logging():
...
+        # Check if specified config file exists
+        if not config_file.exists():
+            raise FileNotFoundError(
+                f"Config file not found: {config_file}. "
+                "Please check the path or omit --config to use environment variables."
+            )
+
+        # Load from file with environment variable overrides
+        env_overrides = {}
+        for field_name in [
```
mainred: We don't have to iterate the fields in main.py, but move the list/detail to Config to simply the server code.

server.py
```diff
@@ -544,14 +601,47 @@ def readiness_check():
         raise HTTPException(status_code=503, detail="Service not ready")
 
 
-if __name__ == "__main__":
+def main():
+    """Holmes AI Server entry point"""
```
arikalon1, line 463: Can we avoid sending args to the server?

Maybe we can just use env variables? (it's easy, because it's already supported in the helm chart)

So, if there's a config file under `DEFAULT_CONFIG_LOCATION`, use it.
By default, there's nothing there on the server.

And if the user wants to use it, it will mount a configmap with the config file, to a volume mound on the server.

WDYT?
(is that what you meant @mainred ?)

server.py
arikalon1: Can you please elaborate @mainred , not sure I fully understood your suggestion
Please look at my other comment below as well, about using env var instead of args

server.py
mainred: Second Arik on using a config map since we have config maps including model list and custom toolset config for mcp and toolset

gossion: sure, removed args.

server.py
```diff
@@ -95,8 +96,29 @@ def init_logging():
...
+    Returns:
+        tuple: (config, dal) - The initialized Config object and its DAL instance
+    """
+    default_config_path = Path(DEFAULT_CONFIG_LOCATION)
+    if default_config_path.exists():
+        logging.info(f"Loading config from file: {default_config_path}")
+        config = Config.load_from_file(default_config_path)
+    else:
+        logging.info("No config file found, loading from environment variables")
+        config = Config.load_from_env()
```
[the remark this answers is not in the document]
gossion: it is already with expanduser

arikalon1 (approved):

PR #1457, opened 2026-01-30
Allow running Prometheus evals in CI/CD

aantn:
Sets up KIND cluster to support this


aantn:
/eval
marker: prometheus

PR #1500, opened 2026-02-05
Use enum parameter for runbook_id validation instead of description

aantn:
Uses enum and tool schema to select runbooks - instead of taking a string
DO NOT MERGE AS IS. The enum is set once on holmes startup so when new server-side runbooks are added after startup, they will not be included in the enum. This needs to be addressed before merging.

## Summary by CodeRabbit

* **Refactor**
  * Runbook selection clarified: local runbooks are listed separately from remote entries and initialization prompts updated for per-source guidance.
* **Bug Fixes**
  * Now requires choosing exactly one source (local file or remote UUID); improved validation, clearer error messages, and stricter local path checks.
* **Tests**
  * Cleaned up and updated several legacy test fixtures and test-case files to remove outdated sample data.


aantn:
/eval
marker: runbook

aantn:
/eval
marker: runbooks

PR #1511, opened 2026-02-06
Allow loops, conditionals, and subshells with user approval

aantn:
## Summary
This PR changes the bash toolset's security model from permanently blocking compound statements (loops, conditionals) and subshells to requiring user approval for these constructs. Hardcoded blocks like `sudo` and `su` remain permanently blocked even within approved scripts.

## Key Changes

- **Updated validation logic**: Compound statements and subshells now return `APPROVAL_REQUIRED` status instead of `DENIED`, allowing users to approve them before execution
- **Added `check_hardcoded_blocks_in_raw_command()`**: New validation function that detects hardcoded blocks anywhere in unparsed commands using word boundaries, enabling security checks within scripts
- **Enhanced security for scripts**: Even when compound statements/subshells are approved, hardcoded blocks like `sudo` and `su` are still permanently denied
- **Updated documentation**: 
  - Reframed "NOT SUPPORTED" as "Requires user approval" in instructions
  - Added guidance on providing prefixes for key operations inside scripts
  - Removed "NEVER USE" sections that implied permanent blocking
- **Updated tool description**: Clarified that loops, conditionals, and subshells are supported with approval
- **Updated tests**: Changed test expectations from `DENIED` to `APPROVAL_REQUIRED` for compound statements and subshells; added new tests for hardcoded block detection in raw commands

## Implementation Details

The validation flow now:
1. Detects subshells → checks for hardcoded blocks → returns `DENIED` if found, else `APPROVAL_REQUIRED`
2. Attempts to parse segments → catches `CompoundStatementError` → checks for hardcoded blocks → returns `DENIED` if found, else `APPROVAL_REQUIRED`
3. Continues with normal segment-by-segment validation for simple commands

This maintains security by preventing privilege escalation attempts while enabling legitimate use of control flow structures with explicit user consent.

https://claude.ai/code/session_01MfpKyxJL5v9BYRHDDZFQM7

## Summary by CodeRabbit

* **Documentation**
  * Reworded bash guidance, added script-specific prefix examples, clarified approval prompts for loops, conditionals, subshells, and compound fragments.

* **New Features / Behavior**
  * Support for dynamic pre-approved command entries with a fallback message.
  * Compound scripts, loops, conditionals and subshells now follow an approval workflow instead of blanket blocking.

* **Refactor**
  * Validation now reports compound detection alongside parsed segments and separates raw-block checks from per-segment checks.

* **Tests**
  * Expanded tests for approval flows, raw-command blocking, compound handling, and updated parsing expectations.

### DIFF
--- a/holmes/core/tools.py
+++ b/holmes/core/tools.py
@@ -287,8 +287,8 @@ class Tool(ABC, BaseModel):
                 logger.info(
                     f"  [yellow]Tool '{self.name}' requires approval: {approval_check.reason}[/yellow]"
                 )
-                # Override suggested_prefixes with filtered list (for bash toolset)
-                if approval_check.prefixes_to_save:
+                # Bash toolset: override suggested_prefixes with filtered list
+                if approval_check.prefixes_to_save is not None:
                     params["suggested_prefixes"] = approval_check.prefixes_to_save
                 return StructuredToolResult(
                     status=StructuredToolResultStatus.APPROVAL_REQUIRED,
--- a/tests/test_bash_toolset_validation.py
+++ b/tests/test_bash_toolset_validation.py
@@ -4,8 +4,11 @@ Unit tests for the bash toolset validation module.
 Tests prefix-based command validation, subshell detection, and allow/deny list handling.
 """
 
+from unittest.mock import MagicMock
+
 import pytest
 
+from holmes.plugins.toolsets.bash.bash_toolset import BashExecutorToolset, RunBashCommand
 from holmes.plugins.toolsets.bash.common.config import (
     HARDCODED_BLOCKS,
     BashExecutorConfig,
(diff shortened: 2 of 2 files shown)


holmes/plugins/toolsets/bash/validation.py
```diff
@@ -345,22 +368,37 @@ def validate_command(
                 message=f"Suggested prefix '{prefix}' does not appear in the command.",
             )
 
-    # Check for subshells
+    # Check for subshells - these require user approval (but still block hardcoded patterns)
     if detect_subshells(command):
```
Sheeproid: @aantn I think the whole 'detect_subshells' function can be removed.

holmes/plugins/toolsets/bash/validation.py
```diff
@@ -345,22 +368,37 @@ def validate_command(
...
+            message="Command contains subshell constructs ($(), ``, <(), >()) which require approval.",
+            prefixes_needing_approval=suggested_prefixes,
         )
 
     # Parse command into segments (may raise CompoundStatementError)
+    # Compound statements require user approval (but still block hardcoded patterns)
     try:
         segments = parse_command_segments(command)
     except CompoundStatementError:
+        blocked = check_hardcoded_blocks_in_raw_command(command)
```
Sheeproid: this is NOT good, the 'CompoundStatementError' will stop processing thus 'sgements' will not be checked for all the 'denied' commands.
Instead, we need to just return a bool/flag from the traversal named 'contains_compund_command' and in that case require an approval even if 'any_needs_approval' is False later in the function

kbpranay:
This looks like a great improvement to security controls! Managing the balance between power and safety in AI-driven bash tools is tricky. I especially like the move toward explicit approval for compound statements.

Since you are working on refined control for AI agents, you might find **OpenClaw** (or its managed version ClawOnCloud) interesting. It is a self-hosted personal AI assistant framework that focuses on giving users full agency over their tools and privacy while providing a robust skill system for exactly these kinds of operations. Check it out at https://clawoncloud.com if you ever need a reference implementation for agentic safety and tool execution.

PR #1530, opened 2026-02-08
Add holems operator health checks

moshemorad:
## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Kubernetes operator to manage and run HealthCheck resources with full lifecycle, status, and notifications
  * Operator exposes a public HTTP client to execute checks against the Holmes API

* **Infrastructure**
  * Helm chart support for operator deployment with RBAC, configurable settings, and secure container image

* **Tests**
  * Component tests covering successful, failed, and error health-check executions

### DIFF
--- a/Dockerfile.operator
+++ b/Dockerfile.operator
@@ -2,9 +2,6 @@ FROM python:3.11-slim
 
 WORKDIR /app
 
-RUN apt-get update && apt-get install -y && \
-    rm -rf /var/lib/apt/lists/*
-
 COPY pyproject.toml poetry.lock* ./
 RUN pip install --no-cache-dir poetry && \
     poetry config virtualenvs.create false && \


holmes_operator/client/holmes_api_client.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+    async def execute_check(
+        self,
+        check_name: str,
+        query: str,
+        timeout: int,
+        mode: str,
+        destinations: list,
+        model: Optional[str] = None,
+    ) -> Dict:
```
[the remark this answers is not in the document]
moshemorad: The timeout is for the check and not the request.

holmes/checks/checks_api.py
```diff
@@ -0,0 +1,209 @@
...
+checks_app = FastAPI()
+
+_CONFIG: Config
+
+
+def init_checks_app(main_app: FastAPI, config: Config):
+    global _CONFIG
+    _CONFIG = config
+
+    main_app.mount("/api/checks", checks_app)
```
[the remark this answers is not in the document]
moshemorad: can the request reach before the init_checks_app?

Dockerfile.operator
```diff
@@ -0,0 +1,31 @@
...
+COPY holmes_operator/ ./holmes_operator/
+
+# Create non-root user (use holmesop to avoid conflicts)
+RUN useradd -m -u 10001 holmesop && \
+    chown -R holmesop:holmesop /app
+
+# Use numeric UID for Kubernetes runAsNonRoot compatibility
+USER 10001
+
+# Health check
```
arikalon1: can we remove it?

Dockerfile.operator
```diff
@@ -0,0 +1,31 @@
+FROM python:3.11-slim
+
+WORKDIR /app
+
+# Install system dependencies
+RUN apt-get update && apt-get install -y \
+    git \
```
arikalon1: do we really need git? Couldn't find the usage
If not, we can probably remove this part

holmes_operator/config.py
```diff
@@ -0,0 +1,51 @@
+"""Configuration for Holmes Operator."""
+
+import json
+import os
+from dataclasses import dataclass
+from typing import Optional
+
+
+def load_bool(env_var, default: Optional[bool]) -> Optional[bool]:
```
arikalon1, line 9: Should we use the function we already have in common/env_vars.py?

moshemorad: Yeah, wanted to do it as well but than i will need to copy also part of Holmes folder (the operator is in separate module) 
I think we need to start and add more internal lib and reuse everywhere.

arikalon1: works

PR #1532, opened 2026-02-08
Add scheduled initial

moshemorad:
## Summary by CodeRabbit

* **New Features**
  * Introduced ScheduledHealthCheck CRD for cron-based health checks with enable/disable, modes (alert/monitor), destinations, and detailed status (last run, result, active, history).
  * Exposed runtime columns in kubectl for Schedule, Enabled, Last Run, Last Result, Active, Age.
  * Holmes Operator: deployable via Helm (configurable), container image/runtime prepared.
  * Per-destination Slack token/channel support and richer API responses including status enums, model_used, and notifications.

* **Tests**
  * Added unit tests covering HealthCheck execution success, failure, and error paths.

### DIFF
--- a/holmes_operator/scheduler/manager.py
+++ b/holmes_operator/scheduler/manager.py
@@ -134,9 +134,6 @@ class SchedulerManager:
         key = f"{namespace}/{name}"
         logger.info(f"Updating schedule for {key}")
 
-        # Remove old job if exists
-        await self.remove_schedule(name, namespace)
-
         # Add new job
         return await self.add_schedule(name, namespace, cron_expr, spec, scheduled_uid)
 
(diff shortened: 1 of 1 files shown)


holmes_operator/handlers/scheduledhealthcheck.py
```diff
@@ -0,0 +1,317 @@
...
+    )
+
+
+async def _add_scheduledhealthcheck_condition(
+    api, name: str, namespace: str, condition: HealthCheckCondition
+):
+    """Add or update condition in ScheduledHealthCheck status."""
+    try:
+        # Get current resource
+        resource = await asyncio.to_thread(
```
arikalon1, line 272: Should this have some concurrency protection?
Like
Thread A: api.get_namespaced_custom_object
Thread B: api.get_namespaced_custom_object
Thread B: api.patch_namespaced_custom_object_status
Thread A: api.patch_namespaced_custom_object_status - Thread B change is lost

Is that a valid scenario?

It appears in a few places, where we read the resource, update and write it.
I think there might be a race condition where we override updates, but not sure if we want to address it

holmes_operator/scheduler/manager.py
```diff
@@ -0,0 +1,292 @@
...
+                await self.add_schedule(
+                    name=name,
+                    namespace=namespace,
+                    cron_expr=scheduled_spec.schedule,
+                    spec=scheduled_spec,
+                    scheduled_uid=uid,
+                )
+
+            except Exception as e:
+                # Individual schedule failures are logged but don't fail startup
```
arikalon1, line 212: In case we failed to add the schedule, should we update the scheduled check somehow? 
Will we see it somewhere in the k8s events?

holmes_operator/scheduler/manager.py
```diff
@@ -0,0 +1,292 @@
...
+            scheduled_uid: UID of the ScheduledHealthCheck resource
+
+        Returns:
+            New job ID
+        """
+        key = f"{namespace}/{name}"
+        logger.info(f"Updating schedule for {key}")
+
+        # Remove old job if exists
+        await self.remove_schedule(name, namespace)
```
arikalon1: On the `add_schedule` we use `replace_existing=true`
Can we skip the remove here?

holmes_operator/scheduler/manager.py
```diff
@@ -0,0 +1,292 @@
...
+                    cron_expr=scheduled_spec.schedule,
+                    spec=scheduled_spec,
+                    scheduled_uid=uid,
+                )
+
+            except Exception as e:
+                # Individual schedule failures are logged but don't fail startup
+                logger.error(f"Failed to load schedule {namespace}/{name}: {e}")
+
+    async def _check_and_catchup(
```
arikalon1: I think this is not used

holmes_operator/context.py
```diff
@@ -49,17 +59,28 @@ def initialize(cfg: OperatorConfig) -> None:
...
     api_client = HolmesAPIClient(
-        base_url=cfg.holmes_api_url,
-        timeout=cfg.holmes_api_timeout,
+        base_url=config.holmes_api_url,
+        timeout=config.holmes_api_timeout,
     )
 
+    # Initialize and start scheduler manager
+    scheduler_manager = SchedulerManager(
+        timezone_str="UTC",
```
arikalon1, line 68: Since we did the hard work of supporting timezones, should we expose it in a helm value or env var?

holmes_operator/scheduler/manager.py
moshemorad: That is a good point, i will open a ticket for that

holmes_operator/context.py
moshemorad: Initially it was with timezone support. I thought that it will just confuse the user since we only allow cron format. 
We can do it if there will be request from the community.

holmes_operator/handlers/scheduledhealthcheck.py
moshemorad: I don't think it should happen since this flow only runs from kops handlers. If there will be some race it will probably will fail on updating the object before the conditions.

I don't want to add retry logic on set conditions yet. we can do it later on if it will raise.

holmes_operator/scheduler/manager.py
moshemorad: https://robusta-dev.atlassian.net/browse/ROB-3318

PR #1574, opened 2026-02-16
Optimize LLM calls by batching final answer with todo completion

aantn:
## Summary
This PR optimizes the LLM interaction flow by allowing the model to provide its final answer in the same response as marking the last tasks as completed, eliminating unnecessary extra LLM calls.

## Key Changes

- **Early return optimization**: Added `_all_todos_completed()` helper function to detect when all tasks in a TodoWrite call are marked as completed
- **Skip redundant LLM calls**: Modified both `call()` and `call_stream()` methods in `ToolCallingLLM` to return early when:
  - All todos are marked as completed
  - The LLM has already provided a text response in the same turn
  - This prevents an unnecessary extra LLM call that would only be used to generate the final answer
- **Updated prompts**: Enhanced instructions across multiple prompt templates to explicitly encourage batching the final answer with the last TodoWrite call:
  - `_general_instructions.jinja2`
  - `_noflag_general_instructions.jinja2`
  - `investigator_instructions.jinja2`
  - `investigation_procedure.jinja2`

## Implementation Details

The optimization works by:
1. Checking if any TodoWrite tool call in the current batch has all tasks marked as "completed"
2. Verifying that the LLM response includes text content alongside the tool calls
3. Returning the result immediately instead of looping for another LLM call
4. Properly accounting for token usage and metadata before returning

The prompt updates make it clear to the LLM that it should include the final answer text in the same response as the final TodoWrite call, rather than waiting for a separate turn.

https://claude.ai/code/session_01D3bToEw4CfTUkPWhVGsHCN

## Summary by CodeRabbit

* **Refactor**
  * Streamlined finalization flow so the last task’s response includes the complete final answer, removing an extra completion step and reducing response latency.
  * Harmonized efficiency guidance across task systems to consistently handle the final task and improve end-to-end response timing.


aantn:
/eval
branch: master
iterations: 3

aantn:
/eval
iterations: 3

aantn:
/eval
iterations: 3

aantn:
/eval
marker: regression
branch: master

PR #1583, opened 2026-02-18
ROB-3335 Add interrupt support for LLM calls via Escape key

naomi-robusta:
image

## Summary
This PR adds the ability to interrupt long-running LLM calls by pressing the Escape key in interactive mode. The implementation includes terminal control for detecting Escape key presses, thread-based execution of LLM calls, and proper cleanup/rollback of conversation state on interruption.

## Key Changes

- **New `LLMInterruptedError` exception** in `tool_calling_llm.py` to signal user-initiated interruptions
- **Terminal control integration** in `interactive.py`:
  - Detects and handles Escape key presses while respecting escape sequences (arrow keys, etc.)
  - Gracefully falls back to simple thread joining on systems without terminal control or non-TTY environments
  - Properly manages terminal state restoration
  
- **Thread-based LLM execution**:
  - LLM calls now run in a background thread to allow concurrent monitoring for interruption signals
  - Implements `_wait_for_completion_or_escape()` function to monitor for Escape key while thread executes
  - Coordinates terminal access between escape listener and approval UI callbacks
  
- **Cancellation support in LLM call loop**:
  - Added `cancel_event` parameter to `ToolCallingLLM.call()` method
  - Checks for cancellation at multiple points: before iterations, after API responses, and during tool execution
  - Allows graceful cancellation of concurrent tool executions
  
- **Conversation state rollback**:
  - Snapshots message history before LLM call
  - Restores previous state if call is interrupted, allowing user to retry
  - Displays "Interrupted" status message to user

- **User feedback**:
  - Updated "Thinking..." prompt to indicate Escape key can interrupt
  - Handles both direct Escape interruption and `LLMInterruptedError` exceptions

## Implementation Details

- Uses `select` module for non-blocking stdin monitoring (Unix-like systems)
- Disambiguates standalone Escape key from escape sequences by checking for follow-up characters
- Wraps approval callbacks to coordinate terminal state with escape listener
- Thread-safe event signaling via `threading.Event` objects
- Graceful degradation when terminal control unavailable (Windows, non-TTY environments)

https://claude.ai/code/session_01HLFgjJzvw3oHLoY3AGZvHv

## Summary by CodeRabbit

* **New Features**
  * Press Escape to interrupt in-progress AI responses, including tool-invoking/streaming answers.
  * Cooperative cancellation so in-flight AI operations stop promptly when interrupted.
  * Conversation history is rolled back and restored after interruption for seamless continuation.
  * Interactive prompt shows an interrupt hint when your terminal supports it.
  * New configuration options: initial input, include files, toggle tool output, tracing/feedback hooks, and save conversation JSON after each AI response.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -72,7 +72,7 @@ jobs:
 
           echo "should_test=$SHOULD_TEST" >> $GITHUB_OUTPUT
 
-  build:
+  test:
     needs: check-changes
     if: needs.check-changes.outputs.should_test == 'true'
     strategy:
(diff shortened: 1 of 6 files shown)


holmes/core/tool_calling_llm.py
```diff
@@ -450,6 +458,9 @@ def call(  # type: ignore
         i = 0
         metadata: Dict[Any, Any] = {}
         while i < max_steps:
+            if cancel_event and cancel_event.is_set():
```
moshemorad, line 461: The current implementation doesn't break when the actual llm call is being done. 
Is this intentional? I think it would be better to interrupt it also while in the middle of the llm request. if it isn't easy to implement let's make sure that we at least printing that we got the interrupt command and it is in progress. 

WDYT?

moshemorad (approved):

holmes/core/tool_calling_llm.py
naomi-robusta: The response time is super quick (try it)
I don't think we need anything faster

PR #1630, opened 2026-02-23
ROB-3382 Implement lazy initialization for toolset prerequisites in CLI mode

naomi-robusta:
image

## Summary
This PR implements lazy initialization for toolset prerequisites to improve startup performance. Instead of running all prerequisite checks (including slow network/IO operations) at startup, fast config-validity checks are run immediately while slower checks are deferred until the first time a tool from that toolset is used.

## Key Changes

- **New prerequisite prioritization**: Added `_prereq_priority()` function to classify prerequisites by speed:
  - Fast (config-validity): Static flags and environment variables
  - Slow (deferrable): Callable and command prerequisites

- **Split prerequisite checking into two phases**:
  - `check_config_prerequisites()`: Runs only fast checks (static + env vars) at startup
  - `lazy_initialize()`: Runs deferred checks (callable + command) on first tool use
  - `check_prerequisites()`: Existing method unchanged, runs all checks when needed

- **Lazy initialization tracking**: Added private attributes to `Toolset`:
  - `_lazy_init`: Whether toolset has deferred prerequisites
  - `_initialized`: Whether all prerequisites have been checked
  - `needs_initialization` property: Indicates if lazy init is pending

- **ToolExecutor integration**: 
  - Added `ensure_toolset_initialized()` method to trigger lazy initialization before tool invocation
  - Tracks tool-to-toolset mapping for efficient lookup

- **ToolsetManager updates**:
  - Added `_check_config_prerequisites()` for batch config validation
  - Modified `load_toolset_with_status()` to use lazy initialization for cached toolsets
  - MCP servers still use eager initialization (required to load tool definitions)

- **Tool invocation integration**: Modified `_directly_invoke_tool_call()` to ensure toolset initialization before executing tools

## Implementation Details

- Prerequisite ordering is preserved: static checks run before env vars, which run before callables/commands
- Short-circuit evaluation: if a fast check fails, deferred checks are never considered
- MCP toolsets bypass lazy initialization since they need tool definitions at load time
- Lazy initialization is idempotent: calling `lazy_initialize()` multiple times is safe
- Comprehensive test coverage for all lazy initialization scenarios

https://claude.ai/code/session_011HzuYzKJ4Gv8HiE61F4PjQ

## Summary by CodeRabbit

* **New Features**
  * Toolsets now support lazy (on-demand) initialization to defer slow prerequisite checks and improve startup responsiveness.
  * Added a visible indicator and a way to explicitly trigger initialization when a toolset requires it.

* **Bug Fixes**
  * Tool invocation now checks initialization and reports errors before running tools, preventing failed executions.

* **Tests**
  * Added coverage for lazy init, config-only checks, callable/command prerequisites, and failure/edge cases.


holmes/core/tools_utils/tool_executor.py
```diff
@@ -43,13 +44,33 @@ def __init__(self, toolsets: List[Toolset]):
...
+
+        For toolsets loaded from cache without full initialization, this triggers
+        the deferred prerequisite checks (callable and command prerequisites)
+        on first tool use.
+
+        Returns None on success, or an error message string on failure.
+        """
+        toolset = self._tool_to_toolset.get(tool_name)
+        if toolset is None or not toolset.needs_initialization:
+            return None
```
Avi-Robusta, line 66: I suspect there is a bug here if lazy initialization fails

**First call to a tool from a lazily-loaded toolset that fails:**
needs_initialization is True → enters the lazy_initialize() branch
lazy_initialize() fails → returns error message to the LLM

**Second call to any tool from the same toolset:**
needs_initialization is False (because _initialized = True after the first attempt)
Returns None (no error) → the tool proceeds to execute despite the toolset.status == FAILED

can you verify lazy initialization failure doesnt try to reinitialize?

PR #1657, opened 2026-03-04
Handle Prometheus toolset import failure gracefully on Python 3.14+

namcxn:
## Summary
   - Wrap Prometheus toolset import in try/except to handle Python 3.14+ compatibility issues gracefully
   - Log a warning with actionable guidance instead of crashing the application

   ## Test plan
   - [ ] Verify HolmesGPT starts without errors on Python 3.12/3.13
   - [ ] Verify HolmesGPT starts with a warning (not crash) on Python 3.14+

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when optional integrations fail to load so the app logs a warning and continues rather than crashing.
* **Chores**
  * No user-facing API changes.

### DIFF
--- a/holmes/plugins/toolsets/__init__.py
+++ b/holmes/plugins/toolsets/__init__.py
@@ -123,15 +123,15 @@ def load_python_toolsets(
     if not DISABLE_PROMETHEUS_TOOLSET:
         try:
             from holmes.plugins.toolsets.prometheus.prometheus import PrometheusToolset
-
-            toolsets.append(PrometheusToolset())
-        except Exception as e:
+        except (ImportError, ModuleNotFoundError) as e:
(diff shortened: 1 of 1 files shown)


moshemorad:
Hi @namcxn.
Thanks for your contribution.

I tried to reproduce the error with python 3.14 without success on the current master branch.
Can you please add a step to reproduce?

namcxn:
Hi @moshemorad 

I just tested again with Python 3.14 and it appears the issue has been resolved. I no longer need to disable DISABLE_PROMETHEUS_TOOLSET.

Thanks for taking the time to verify this.

moshemorad:
Thanks @namcxn
We appreciate your contribution

PR #1685, opened 2026-03-06
Optimize Anthropic prompt caching for tool definitions

aantn:
## Summary
This change improves prompt caching efficiency when using Anthropic's API by enabling independent caching of tool definitions from system prompts and messages. This is particularly beneficial for scenarios where multiple requests share the same toolset but have different prompts (e.g., scheduled reports).

## Key Changes
- Added `cache_control` with `ephemeral` type to the last tool definition, allowing Anthropic to cache tools independently
- Updated `cache_control_injection_points` configuration to cache both the system prompt prefix and the full prefix up to the last message
- Leverages Anthropic's cache hierarchy (tools → system → messages) to maximize cache hit rates across requests with shared toolsets

## Implementation Details
- Tool caching is applied to the last tool in the tools array to avoid redundant cache control markers
- Two cache control injection points are now configured:
  1. System message level: caches the tools + system prompt prefix independently
  2. Message level: caches the full prefix up to the last message
- This approach ensures that requests with identical tools but different prompts can benefit from cached tool definitions

https://claude.ai/code/session_01DYSeL2XYJGU9jJ4exhjW7p

## Summary by CodeRabbit

* **Refactor**
  * Improved ephemeral caching for LLM completions when tools are used, with more precise cache breakpoints.

* **Tests**
  * Added test infrastructure to warm prompt/tool caches before test runs, reducing cold-start variability.

* **Chores**
  * Updated an LLM client dependency to a newer patch release for improved runtime behavior.


tshtark:
We run 0.40.0 in prod and measured this on real traffic: every fresh `/api/chat` request re-writes a ~43k–58k token prefix (system + 106 tool schemas) because the only breakpoint sits on the last message. With the `role: system` injection point from this PR, a second request with a different ask read all 42,668 prefix tokens from cache (litellm 1.89.0, claude-sonnet-4-5). We are carrying the two `llm.py` hunks as a version-pinned overlay until this lands. Happy to help with the provider gating CodeRabbit asked for if that is what blocks it.

PR #1723, opened 2026-03-10
fix(llm): load MODEL env fallback and improve no-model guidance

### LINKED ISSUE #1722: CLI fails with "No llm models were loaded" when only provider API keys are set. No error or guidance about MODEL env
### What happened?

  - Running holmes ask with only provider API key(s) set (for example GEMINI_API_KEY) and no explicit model caused Holmes to fail with:
      - Exception: No llm models were loaded
  - The error was not actionable and did not explain that a MODEL has not been configured
  - Docs were inconsistent for Gemini CLI setup:
      - TOOL_SCHEMA_NO_PARAM_OBJECT_IF_NO_PARAMS=true and export MODEL=... were not consistently shown in quick-start paths.

  - If MODEL is exported, Holmes should load that model automatically in CLI mode.
  - If no model is configured, Holmes should fail with a clear message explaining exactly how to fix it (--model,
    MODEL, or model list config).
  - Gemini docs should consistently include required env setup and both invocation styles (exported MODEL and per-command --model).

```
unset MODEL
unset OPENAI_API_KEY
unset ANTHROPIC_API_KEY
unset AZURE_API_KEY
# keep only Gemini key set (or any other provider key)
export GEMINI_API_KEY="<your-key>"
```

  2. Run Holmes ask:

`  holmes ask "What pods do I have in my default namespace" --verbose --no-interactive
`
  3. Observe failure:

  - Before fix: generic No llm models were loaded error is displayed
  - No mention of missing MODEL env

_No response_

pavangudiwada:
## Summary

  This PR fixes #1722 

  ### What changed

  - Load `MODEL` from environment when config does not provide a model in CLI flow.
  - Improve the `No llm models were loaded` failure with actionable, provider-agnostic guidance:
    - `--model <provider/model>`
    - `export MODEL=<provider/model>`
    - model list via `MODEL_LIST_FILE_LOCATION` / config
    - clarify that API key alone is not enough
  - Add regression tests:
    - CLI model loading from `MODEL` env var
    - Helpful error path when no models are loaded
  - Update docs to reflect required Gemini CLI setup:
    - `TOOL_SCHEMA_NO_PARAM_OBJECT_IF_NO_PARAMS=true`
    - `export MODEL=...` usage (in addition to `--model`)

  - `holmes/core/llm.py`

  Executed:

  - `poetry run pytest tests/config_class/test_config_get_llm.py -q --no-cov`
  - `poetry run pytest tests/core/test_llm_model_registry_get_model_params.py -q --no-cov`

  Both passed.

## Summary by CodeRabbit

* **New Features**
  * Honor MODEL environment variable as a default model source and support per-command model overrides.

* **Documentation**
  * Updated CLI installation and Gemini provider docs with examples showing MODEL usage, per-command --model, and exporting TOOL_SCHEMA_NO_PARAM_OBJECT_IF_NO_PARAMS before invoking the CLI.

* **Improvements**
  * More helpful error messages when no models are configured, with guidance on how to specify a model.

* **Tests**
  * Added/updated tests covering MODEL env loading, blank MODEL handling, and the no-model error.

### DIFF
--- a/holmes/core/tool_calling_llm.py
+++ b/holmes/core/tool_calling_llm.py
@@ -144,6 +144,7 @@ class LLMCosts(BaseModel):
     cached_tokens: Optional[int] = None
     reasoning_tokens: int = 0
     max_completion_tokens_per_call: int = 0
+    max_prompt_tokens_per_call: int = 0
     num_compactions: int = 0
 
 
@@ -181,6 +182,9 @@ def _process_cost_info(
                 costs.reasoning_tokens += raw.reasoning_tokens
                 costs.max_completion_tokens_per_call = max(
                     costs.max_completion_tokens_per_call, raw.completion_tokens
+                )
+                costs.max_prompt_tokens_per_call = max(
+                    costs.max_prompt_tokens_per_call, raw.prompt_tokens
                 )
         elif raw.cost > 0:
             cost_logger.debug(
@@ -452,6 +456,12 @@ class ToolCallingLLM:
                 costs.prompt_tokens += compaction.prompt_tokens
                 costs.completion_tokens += compaction.completion_tokens
                 costs.total_cost += compaction.cost
+                costs.max_prompt_tokens_per_call = max(
+                    costs.max_prompt_tokens_per_call, compaction.prompt_tokens
+                )
+                costs.max_completion_tokens_per_call = max(
+                    costs.max_completion_tokens_per_call, compaction.completion_tokens
+                )
 
             if (
                 limit_result.conversation_history_compacted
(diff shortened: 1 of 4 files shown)


holmes/core/llm.py
```diff
@@ -539,6 +539,13 @@ def _should_load_config_model(self) -> bool:
                 return False
             return True
 
+        # Load model from environment when config doesn't provide one.
+        # This is used by CLI flows that rely on `export MODEL=...`.
+        model_from_env = os.environ.get("MODEL")
+        if model_from_env:
+            self.config.model = model_from_env
+            return True
```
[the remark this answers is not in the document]
pavangudiwada: Created a PR fixing this

holmes/core/llm.py
```diff
@@ -574,6 +574,15 @@ def _should_load_config_model(self) -> bool:
                 return False
             return True
 
+        # Load model from environment when config doesn't provide one.
+        # This is used by CLI flows that rely on `export MODEL=...`.
+        model_from_env = os.environ.get("MODEL")
```
moshemorad: Can we move it to the load logic in the config.py? 
Config is responsible for loading config and we want to make sure it isn't scattered around.

The reason we have a line below the check fo OPENAI_API_KEY is only since it was for backward compatibility.

moshemorad (changes requested):
Looks good. 
Left one comment.

holmes/core/llm.py
pavangudiwada: Understood, pushed a change to fix it.

Issue #1677, opened 2026-03-06
An error occurred when configuring Prometheus：TypeError: '>' not supported between instances of 'str' and 'int'


shuaiganzi:
### What happened?

[32m2026-03-06 01:05:16.435 INFO     Running tool #1 [bold]execute_prometheus_instant_query[/bold]: Prometheus: Query (Get container memory usage sorted)[0m
[32m2026-03-06 01:05:16.435 INFO     Failed to connect to Prometheus[0m
Traceback (most recent call last):
  File "/app/holmes/plugins/toolsets/prometheus/prometheus.py", line 1366, in _invoke
    if timeout > max_timeout:
       ^^^^^^^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'str' and 'int'

config:
`  custom_toolset.yaml: |-
    toolsets: 
      bash:
        config:
          builtin_allowlist: extended
        enabled: true
      connectivity_check:
        enabled: true
      internet:
        enabled: true
      kubernetes/core:
        enabled: true
      kubernetes/logs:
        enabled: true
      prometheus/metrics:
        enabled: true
        config:
          prometheus_url: 'http://10.43.162.40:8428/'
          query_timeout_seconds_hard_max: 30
          query_timeout_seconds_default: 20
      robusta:
        enabled: true
      runbook:
        enabled: true
    mcp_servers: 
      {}`

### What did you expect to happen?

this why?

### How can we reproduce it (as minimally and precisely as possible)?

no

aantn:
@shuaiganzi is this fixed by https://github.com/HolmesGPT/holmesgpt/pull/1790

PR #1729, opened 2026-03-10
fix(bash,prometheus): handle string timeout and improve unparseable command validation

### LINKED ISSUE #1727: Bash toolset: TypeError when LLM passes timeout as string
## Description

The bash toolset crashes with a `TypeError` when the LLM passes the `timeout` parameter as a string instead of an integer.

## Error

```
File "/app/holmes/plugins/toolsets/bash/bash_toolset.py", line 258, in _invoke
    result = execute_bash_command(cmd=command_str, timeout=timeout)
File "/app/holmes/plugins/toolsets/bash/common/bash.py", line 39, in execute_bash_command
    stdout, _ = process.communicate(timeout=timeout)
File "/usr/local/lib/python3.11/subprocess.py", line 1204, in communicate
    endtime = _time() + timeout
              ~~~~~~~~^~~~~~~~~
TypeError: unsupported operand type(s) for +: 'float' and 'str'
```

## Steps to reproduce

1. Deploy Holmes 0.20.0 in server mode with bash toolset enabled
2. Trigger an investigation that uses bash commands
3. The LLM sometimes passes `timeout` as a string (e.g. `"30"` instead of `30`)
4. Every bash command execution crashes with the TypeError above

## Context

This is the same class of bug as #1677 (Prometheus toolset timeout), but in the bash toolset. `params.get("timeout", 30)` returns whatever the LLM provided — if it's a string, `subprocess.communicate(timeout=...)` fails because it tries `float + str`.

### LINKED ISSUE #1728: Bash toolset: commands with unquoted parentheses always require approval even when all segments are allowed
## Description

When the LLM generates a bash command with unquoted parentheses in arguments (e.g. PromQL queries), `bashlex.parse()` raises a `ParsingError`. The current fallback in `validate_command()` returns `APPROVAL_REQUIRED` for **all** unparseable commands, even when every command segment is in the allow list.

In server mode (e.g. triggered from Keep workflows), there is no user to approve — so these commands silently fail.

## Example

Command generated by LLM:
```bash
curl -sG http://vmselect:8481/select/0/prometheus/api/v1/query --data-urlencode query=topk(10, nvme_smart_log_media_errors) | jq .data.result
```

Both `curl` and `jq` are in the configured allow list, but the command gets `APPROVAL_REQUIRED` because bashlex can't parse unquoted `()`.

Note: if the LLM properly quotes the argument (`'query=topk(10, ...)'`), bashlex parses fine. But LLMs don't always quote consistently.

## Environment

- Holmes 0.20.0, server mode
- Bash toolset with `builtin_allowlist: extended` and `allow: [curl, jq, wget]`

Aladex:
## Related Issues

- Fixes #1727 — Bash toolset: TypeError when LLM passes timeout as string
- Fixes #1728 — Bash toolset: commands with unquoted parentheses always require approval even when all segments are allowed
- Related to #1677 — Prometheus toolset: same timeout TypeError

### 1. Timeout type coercion (bash + prometheus toolsets)

LLMs sometimes pass the `timeout` tool parameter as a string (`"30"`) instead of an integer. This causes:

- **Bash toolset**: `TypeError: unsupported operand type(s) for +: 'float' and 'str'` in `subprocess.communicate(timeout=timeout)`
- **Prometheus toolset**: `TypeError: '>' not supported between instances of 'str' and 'int'` in timeout comparison

Fix: cast `timeout` to `int` with a try/except fallback to the default value.

### 2. shlex-based fallback for unparseable commands (bash toolset)

When `bashlex.parse()` fails (e.g. unquoted parentheses in PromQL queries like `query=topk(10, metric)`), the current code always returns `APPROVAL_REQUIRED`. In server mode (e.g. Keep/Robusta integrations), there is no user to approve, so these commands silently fail.

Added a fallback that uses `shlex.split()` (stdlib, respects shell quoting) to tokenize the command, then groups tokens into segments by shell operators (`|`, `&&`, `;`, `&`) and validates each segment against the allow/deny lists. If all segments are allowed, the command is permitted.

Security is preserved:
- Hardcoded blocks and deny list are checked on the raw string **before** the fallback
- Each segment in the fallback goes through the full `validate_segment()` deny → allow → approval pipeline
- If any segment is not in the allow list, the behavior is unchanged (`APPROVAL_REQUIRED`)
- If `shlex.split()` also fails, falls through to `APPROVAL_REQUIRED`

### Tests

Added 3 test cases to `test_bash_toolset_validation.py`:
- `curl` with parentheses in args piped to `jq` (both allowed) → `ALLOWED`
- `curl` with parentheses piped to non-allowed command → `APPROVAL_REQUIRED`
- `curl` with parentheses piped to denied command → `DENIED`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout parameter handling in Bash and Prometheus tools to gracefully convert string values to integers, preventing type errors.

* **New Features**
  * Added fallback command validation for complex Bash commands that may not parse with the primary parser.

* **Tests**
  * Expanded test coverage for edge cases in command validation with special characters and piped operations.

### DIFF
--- a/holmes/plugins/toolsets/bash/bash_toolset.py
+++ b/holmes/plugins/toolsets/bash/bash_toolset.py
@@ -195,6 +195,11 @@ class RunBashCommand(Tool):
         command_str = params.get("command")
         suggested_prefixes = params.get("suggested_prefixes", [])
         timeout = params.get("timeout", 30)
+        # LLM may pass timeout as string — ensure it's an int for subprocess
+        try:
+            timeout = int(timeout)
+        except (TypeError, ValueError):
+            timeout = 30
 
         # Validate required parameters
         if not command_str:
--- a/holmes/plugins/toolsets/bash/validation.py
+++ b/holmes/plugins/toolsets/bash/validation.py
@@ -7,6 +7,7 @@ against allow/deny lists, with support for composed commands (pipes, &&, etc.).
 
 import logging
 import re
+import shlex
 from dataclasses import dataclass
 from enum import Enum
 from typing import List, Optional, Tuple
@@ -324,6 +325,57 @@ def validate_command(
                 deny_reason=DenyReason.DENY_LIST,
                 message=f"Command matches deny list pattern '{denied}'. This command is blocked by configuration.",
             )
+
+        # Fallback: use shlex to tokenize (respects quoting), then split
+        # tokens into command segments by shell operators (|, &&, ;, etc.).
+        # bashlex may fail on valid commands with special chars in quoted
+        # arguments (e.g. parentheses in PromQL queries passed to curl).
+        SHELL_OPERATORS = {"|", "||", "&&", ";", "&"}
+        try:
+            tokens = shlex.split(command)
+        except ValueError:
+            # shlex can't parse either — fall through to approval
+            return ValidationResult(
+                status=ValidationStatus.APPROVAL_REQUIRED,
+                message="Command contains complex syntax which requires approval.",
+                prefixes_needing_approval=[],
+            )
+
+        # Group tokens into segments separated by shell operators
+        segments_from_shlex: List[str] = []
+        current_segment: List[str] = []
+        for token in tokens:
+            if token in SHELL_OPERATORS:
+                if current_segment:
+                    segments_from_shlex.append(" ".join(current_segment))
+                    current_segment = []
+            else:
+                current_segment.append(token)
+        if current_segment:
+            segments_from_shlex.append(" ".join(current_segment))
+
+        if not segments_from_shlex:
+            return ValidationResult(
+                status=ValidationStatus.APPROVAL_REQUIRED,
+                message="Command contains complex syntax which requires approval.",
+                prefixes_needing_approval=[],
+            )
+
+        all_allowed = True
+        for seg in segments_from_shlex:
+            seg_result = validate_segment(seg, allow_list, deny_list)
+            if seg_result.status == ValidationStatus.DENIED:
+                return seg_result
+            if seg_result.status != ValidationStatus.ALLOWED:
+                all_allowed = False
+                break
+
+        if all_allowed:
+            logger.info(
+                f"Command allowed via fallback validation (bashlex parse failed): {command[:200]}"
+            )
+            return ValidationResult(status=ValidationStatus.ALLOWED)
+
         return ValidationResult(
             status=ValidationStatus.APPROVAL_REQUIRED,
             message="Command contains complex syntax which requires approval.",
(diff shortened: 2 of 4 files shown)


mbadran-c (approved):

Aladex:
@aantn could you take a look when you get a chance? This fixes the timeout TypeError you asked about in #1727 (we're using gpt-5-mini) and also the bashlex parse failure for PromQL queries (#1728) which silently breaks commands in server mode.

aantn:
@Aladex can you try this with https://github.com/HolmesGPT/holmesgpt/pull/1790 and see if it fixes it?
We're trying to address with broader coercion mechanism that isn't tool specific.

Re the bashlex.parse() failure, isn't the generated bash command invalid?

Aladex:
@aantn thanks for the pointer!

**Timeout coercion** — I'll rebase on top of #1790 and drop the per-tool timeout casts from this PR. The universal coercion approach is cleaner.

**bashlex fallback** — the bash command itself is valid. The typical case is something like:

```
curl 'http://prometheus:9090/api/v1/query?query=topk(10, node_cpu_seconds_total)' | jq .status
```

The parentheses are inside single quotes, so bash handles them fine — but `bashlex.parse()` chokes on them. In server mode (Robusta/Keep integrations) there's no human to click "approve", so these commands silently fail even though both `curl` and `jq` are in the allow list.

The shlex fallback only kicks in when bashlex fails, tokenizes by shell operators, and validates each segment against the same allow/deny lists. If any segment isn't explicitly allowed, it still returns `APPROVAL_REQUIRED`.

aantn:
@Aladex are you sure about that?

```
cat > /tmp/test.py << 'EOF'
import bashlex
try:
  bashlex.parse("curl 'query=topk(10,x)' | jq ."); print("quoted: OK")
except:
  print("quoted: FAIL")
try:
  bashlex.parse("curl query=topk(10,x) | jq ."); print("unquoted: OK")
except:
  print("unquoted: FAIL")
EOF
python3 /tmp/test.py
```

By me bashlex handles it just fine:

```
quoted: OK
unquoted: FAIL
```

In the original issue you opened, the example command was unquoted, so I think the LLM was just generating an invalid command.

In any event, can we fix by giving you the option to auto-disapprove all commands on the server flow and continue the investigation?

Aladex:
@aantn you're right — I double-checked and bashlex does handle the quoted version fine, so the fallback wouldn't actually trigger in that case. And the unquoted version is indeed invalid bash.

I need to dig into our production logs a bit more to catch the exact commands the LLM is generating and understand where it's actually breaking. I'll follow up once I have concrete examples.

The auto-disapprove + continue idea for server mode sounds good regardless — happy to help with that if needed.

Aladex:
@aantn you're right on both counts. Confirmed on production — the LLM does quote the arguments properly, so bashlex handles it fine. The shlex fallback isn't needed.

The timeout issue is also confirmed to be a real problem on 0.20.0, but #1790 addresses it with the universal coercion approach, which is the right fix.

Closing this PR in favor of #1790.

aantn:
Awesome, thank you. Adding docs on how to disable approval in server mode - turns out we support it but its not documented!

https://github.com/HolmesGPT/holmesgpt/pull/1800

Let me know if that helps.

PR #1780, opened 2026-03-15
fix: enrich PagerDuty issues with description and alert body details

yakir-shriker:
## Problem

When investigating a PagerDuty incident via `holmes investigate ticket --source pagerduty`, the LLM receives no useful context from the ticket:

```
for issue '[FIRING:1] TestAlert Grafana' with description:'None'
```

The `description` is always `None`, even though the `Issue` model has a `description` field and it is used in the LLM prompt. This means the AI investigation has to rely solely on the incident title.

This is particularly impactful for incidents created by integrations (Grafana, Datadog, Prometheus Alertmanager, etc.) because the real investigative context—labels, annotations, firing details—is stored in the **alert body**, not in the incident title or description.

### 1. `convert_to_issue` never sets the `description` field

The Jira source correctly populates `description`:

```python
# holmes/plugins/sources/jira/__init__.py (existing)
def convert_to_issue(self, jira_issue, description=None):
    description = self.extract_description(jira_issue)
    return Issue(..., description=description, ...)
```

But the PagerDuty source omits it entirely:

```python
# holmes/plugins/sources/pagerduty/__init__.py (before fix)
def convert_to_issue(self, source_issue):
    return Issue(
        id=source_issue["id"],
        name=source_issue["summary"],
        source_type="pagerduty",
        source_instance_id=self.api_url,
        url=f"{source_issue['html_url']}",
        raw=source_issue,
        # description is never set
    )
```

### 2. `fetch_issue` doesn't fetch alert body details

PagerDuty incidents created by integrations store rich context in alert bodies at `GET /incidents/{id}/alerts`. The current code only calls `GET /incidents/{id}`, which returns a bare-bones incident. A real Grafana-created alert body looks like:

```json
{
  "alerts": [{
    "body": {
      "details": {
        "firing": "Value: [no value]\nLabels: alertname = TestAlert, instance = Grafana\nAnnotations:\n - description = fluentd pods are not healthy in us-west-2-non-prod\n - summary = fluentd pods are not healthy",
        "num_firing": "1"
      }
    }
  }]
}
```

This context is critical for the LLM investigation but is never fetched.

## Fix

1. **Set description in `convert_to_issue`**: Add `description=source_issue.get("description")` to the `Issue()` constructor, consistent with the Jira source.

2. **Enrich with alert details in `fetch_issue`**: Add `_enrich_with_alert_details` helper that calls `GET /incidents/{id}/alerts` and appends all non-empty `body.details` entries to the issue description. This runs only for `fetch_issue` (single-issue lookups used by `investigate ticket`), not for `fetch_issues` (bulk listing).

3. **Graceful degradation**: If the alerts API call fails for any reason, the original description is preserved and investigation continues.

## Before / After

**Before:**

```
description: None
```

**After (Grafana-created incident):**

```
Alert triggered from Grafana

firing: Value: [no value]
Labels: alertname = TestAlert, instance = Grafana
Annotations:
 - description = fluentd pods are not healthy in us-west-2-non-prod
 - summary = fluentd pods are not healthy

num_firing: 1
```

## How to Reproduce

```bash
# Create a test alert in Grafana that fires to PagerDuty
# Then investigate it:
holmes investigate ticket \
  --source pagerduty \
  --ticket-api-key "$PD_KEY" \
  --ticket-username user@example.com \
  --ticket-id Q1EOHXXWSUNEKN \
  "Investigate this PagerDuty incident"

# Before fix: LLM sees "description: None"
# After fix: LLM sees full alert details (annotations, labels, firing info)
```

## Tests Added

- `test_sets_description_from_incident` — verifies `convert_to_issue` now populates `description`
- `test_description_none_when_missing` — backward compat when incident has no description field
- `test_basic_fields_are_set` — verifies existing field mapping isn't broken
- `test_enriches_description_with_alert_details` — verifies alert body details are appended
- `test_graceful_on_empty_alerts` — verifies no crash when alerts list is empty
- `test_graceful_on_alerts_api_failure` — verifies graceful degradation on API errors
- `test_alert_details_without_incident_description` — verifies alert details work even when base description is `None`
- `test_fetch_issues_still_works` — backward compat for bulk fetch (no alert enrichment)

## Release Notes

* **New Features**
  * Issue descriptions from PagerDuty incidents now include enriched alert details for better context and visibility.
  * Original incident descriptions are preserved in the issue data.
  * Improved resilience with graceful fallback when alert details cannot be retrieved.

### DIFF
--- a/docs/reference/python-sdk.md
+++ b/docs/reference/python-sdk.md
@@ -262,7 +262,7 @@ Main configuration class (`holmes.config.Config`).
 |-----------|------|---------|-------------|
 | `api_key` | `str` | Auto-detected | LLM API key. Reads from `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, etc. if not provided. |
 | `model` | `str` | *None* | Model identifier, e.g. `"anthropic/claude-sonnet-4-5-20250929"`. Required. |
-| `max_steps` | `int` | `40` | Maximum tool-calling steps per request. |
+| `max_steps` | `int` | `100` | Maximum tool-calling steps per request. |
 | `custom_toolsets` | `list[path]` | *None* | Paths to custom YAML toolset files. |
 | `additional_toolsets` | `list[Toolset]` | *None* | Python `Toolset` instances to load alongside built-in toolsets. |
 | `toolsets` | `dict` | *None* | Inline toolset configuration overrides. |
--- a/examples/custom_llm.py
+++ b/examples/custom_llm.py
@@ -68,7 +68,7 @@ def ask_holmes():
     )
 
     tool_executor = ToolExecutor(load_builtin_toolsets())
-    ai = ToolCallingLLM(tool_executor, max_steps=40, llm=MyCustomLLM())
+    ai = ToolCallingLLM(tool_executor, max_steps=100, llm=MyCustomLLM())
 
     user_prompt = generate_user_prompt(prompt, context={})
     messages = [
--- a/holmes/config.py
+++ b/holmes/config.py
@@ -56,7 +56,7 @@ class Config(RobustaBaseConfig):
     api_base: Optional[str] = None
     api_version: Optional[str] = None
     fast_model: Optional[str] = None
-    max_steps: int = 40
+    max_steps: int = 100
     cluster_name: Optional[str] = None
 
     alertmanager_url: Optional[str] = None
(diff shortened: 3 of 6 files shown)


aantn:
Thank you for the PR! Approved.

yakir-shriker:
@aantn 
Thanks for approving, when can we merge?

aantn:
> @aantn Thanks for approving, when can we merge?

@yakir-shriker it should merge automatically once the tests finish. Note that evals are failing but that's just because PRs from non maintainers dont have credentials to run them. It wont block the PR.

PR #1791, opened 2026-03-16
Holmes update mcp server instructions update

Avi-Robusta:
The first time you created the MCP server in the ConfigMap, Robusta platform synced that config and stored it server-side. That stored copy included the original llm_instructions.

From that point on, every pod restart:

Platform config loads first — Robusta platform sends its stored copy (with the original llm_instructions) → creates RemoteMCPToolset with the old instructions
ConfigMap loads second — your updated config creates a new RemoteMCPToolset with the new instructions
Merge by name — same name exists, so override_with() runs → copies the new llm_instructions string via setattr but never re-renders it
The bug is that override_with() silently dropped the update. But the deeper reason the old value was there in the first place is that the platform cached your original config and always loaded it as the base toolset before the ConfigMap merge happened.

When you renamed the MCP server, the names didn't match — no merge occurred. The ConfigMap toolset was added fresh, bypassing the platform's cached copy entirely. That's why renaming appeared to fix it.

## Summary by CodeRabbit

* **Bug Fixes**
  * LLM instructions containing dynamic templates now properly update and render when configurations change or during system initialization.

* **Tests**
  * Added comprehensive tests for toolset configuration updates and instruction template rendering scenarios.

### DIFF
--- a/holmes/core/tools.py
+++ b/holmes/core/tools.py
@@ -709,6 +709,11 @@ class Toolset(BaseModel):
             if field in self.__class__.model_fields and value not in (None, [], {}, ""):
                 setattr(self, field, value)
 
+        # Re-render llm_instructions after override, since they may contain Jinja2 templates
+        # that reference tool names or config values which could have changed.
+        if self.llm_instructions:
+            self._load_llm_instructions(self.llm_instructions)
+
     @model_validator(mode="before")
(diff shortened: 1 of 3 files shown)


Avi-Robusta:
needs more investigation to root cause of bug

PR #1830, opened 2026-03-22
Refactor toolset enabling

aantn:
Refactoring, not supposed to change behaviour other than edge case toolset has is_default but was explicitly disabled by user setting enabled: false in the config. We want to respect user intent and disable it which should now be the case after this PR.

Note: this also removes is_default which was not actually used to determine what is enabled by default or not (helm values are used for that + passing enabled=true in the constructor)

https://claude.ai/code/session_01U2xED3Qq15dKJKNkHBcXjq

## Summary by CodeRabbit

* **Bug Fixes**
  * Toolsets with required configuration are now reliably reported as missing configuration, regardless of enablement/default state.

* **Behavior Changes**
  * Toolsets are no longer implicitly designated as "default" in several places; enablement is now controlled explicitly.

* **Tests**
  * Updated tests to reflect the new missing-configuration semantics and default-designation behavior.

* **Chores**
  * Removed the checklist item about the default-flag from the PR template.

### DIFF
--- a/tests/test_toolset_auto_enable.py
+++ b/tests/test_toolset_auto_enable.py
@@ -122,10 +122,6 @@ class TestToolsetManagerAutoEnable:
     """Verify that ToolsetManager._list_all_toolsets respects missing_config
     when enable_all_toolsets=True (the CLI path)."""
 
-    def _make_manager_with_toolsets(self, toolsets: List[Toolset]) -> ToolsetManager:
-        manager = ToolsetManager()
-        return manager
-
     @patch("holmes.core.toolset_manager.load_builtin_toolsets")
     def test_auto_enable_skips_missing_config_toolset(self, mock_load):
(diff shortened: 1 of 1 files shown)


aantn:
@claude review

aantn:
@claude review

holmes/plugins/toolsets/robusta/robusta.py
```diff
@@ -519,7 +519,6 @@ def __init__(self, dal: Optional[SupabaseDal]):
             tags=[
                 ToolsetTag.CORE,
             ],
-            is_default=True,
```
moshemorad, line 522: Just to make sure it is intendeed, i see you removed is_default but didn't set enabled=True. we don't want it by default?

aantn: Yes, I think it's fine

PR #1840, opened 2026-03-25
ROB-3570: Add toolset validation and refresh API endpoints

alonelish:
## Summary

- **`POST /api/toolsets/validate`**: Validates toolset/MCP server YAML config by running `check_prerequisites` without deploying. Returns `valid`/`invalid` status per toolset with error details. Supports builtin toolset overrides, custom toolsets, and MCP servers.
- **`POST /api/toolsets/refresh`**: Signals the existing 5-minute refresh thread to re-check all toolsets immediately and sync statuses to Supabase. Fire-and-forget — returns instantly using a `threading.Event` to wake the loop.

## Test plan

- [ ] `POST /api/toolsets/validate` with valid builtin toolset config (e.g. kubernetes) returns `status: "valid"`
- [ ] `POST /api/toolsets/validate` with invalid API URL returns `status: "invalid"` with error
- [ ] `POST /api/toolsets/validate` with MCP server config validates correctly
- [ ] `POST /api/toolsets/validate` with invalid YAML returns 400
- [ ] `POST /api/toolsets/refresh` returns immediately with `{"status": "refresh_triggered"}`
- [ ] Server logs show "Toolset refresh triggered via API call" after calling refresh endpoint
- [ ] Normal 5-minute refresh loop continues to work unchanged
- [ ] `make test-without-llm` passes

## Summary by CodeRabbit

* **New Features**
  * Toolset validation API: submit a YAML configuration string and receive per-toolset results (valid/invalid, error, description).
  * Toolset refresh API: trigger an immediate on-demand refresh of toolset status.

* **Tests**
  * Added API tests for malformed/empty YAML, validation paths (including MCP/custom toolsets), and refresh triggering.

### DIFF
--- a/holmes/toolsets/toolsets_api.py
+++ b/holmes/toolsets/toolsets_api.py
@@ -89,42 +89,39 @@ def validate_toolset(request: ValidateToolsetRequest) -> ValidateToolsetResponse
         toolsets_to_check = []
         results = []
 
-        # 6. Handle builtin overrides — merge user config onto the fresh builtin definition
-        if builtin_overrides:
+        # 6. Handle builtin overrides — load and merge each individually
+        for name, cfg in builtin_overrides.items():
             try:
-                override_toolsets = load_toolsets_from_config(builtin_overrides, strict_check=False)
-                for override in override_toolsets:
-                    if override.name in builtins_by_name:
-                        full_toolset = builtins_by_name[override.name]
-                        full_toolset.override_with(override)
-                        full_toolset.enabled = True
-                        toolsets_to_check.append(full_toolset)
-                        logging.info(f"Merged config for builtin toolset '{override.name}'")
+                loaded = load_toolsets_from_config({name: cfg}, strict_check=False)
+                if loaded and loaded[0].name in builtins_by_name:
+                    full_toolset = builtins_by_name[loaded[0].name]
+                    full_toolset.override_with(loaded[0])
+                    full_toolset.enabled = True
+                    toolsets_to_check.append(full_toolset)
+                    logging.info(f"Merged config for builtin toolset '{name}'")
             except Exception as e:
-                logging.error(f"Failed to load builtin override toolsets: {e}", exc_info=True)
-                for name in builtin_overrides:
-                    results.append(ValidateToolsetResult(
-                        toolset_name=name,
-                        status=ValidationStatus.INVALID,
-                        error=f"Failed to load toolset config: {e}",
-                    ))
+                logging.error(f"Failed to load builtin toolset '{name}': {e}", exc_info=True)
+                results.append(ValidateToolsetResult(
+                    toolset_name=name,
+                    status=ValidationStatus.INVALID,
+                    error=f"Failed to load toolset config: {e}",
+                ))
 
-        # 7. Handle custom/MCP toolsets
-        if custom_dict:
+        # 7. Handle custom/MCP toolsets — load each individually
+        for name, cfg in custom_dict.items():
             try:
-                custom_toolsets = load_toolsets_from_config(custom_dict, strict_check=True)
-                for ts in custom_toolsets:
-                    ts.enabled = True
-                    toolsets_to_check.append(ts)
-                    logging.info(f"Loaded custom/MCP toolset '{ts.name}'")
+                loaded = load_toolsets_from_config({name: cfg}, strict_check=True)
+                if loaded:
+                    loaded[0].enabled = True
+                    toolsets_to_check.append(loaded[0])
+                    logging.info(f"Loaded custom/MCP toolset '{name}'")
             except Exception as e:
-                logging.error(f"Failed to load custom/MCP toolsets: {e}", exc_info=True)
-                for name in custom_dict:
-                    results.append(ValidateToolsetResult(
-                        toolset_name=name,
-                        status=ValidationStatus.INVALID,
-                        error=f"Failed to load toolset config: {e}",
-                    ))
+                logging.error(f"Failed to load custom/MCP toolset '{name}': {e}", exc_info=True)
+                results.append(ValidateToolsetResult(
+                    toolset_name=name,
+                    status=ValidationStatus.INVALID,
+                    error=f"Failed to load toolset config: {e}",
+                ))
 
         # 8. Run prerequisite checks concurrently
         if toolsets_to_check:


server.py
```diff
@@ -429,6 +442,130 @@ def chat(chat_request: ChatRequest, http_request: Request):
 )
 
 
+@app.post("/api/toolsets/validate")
```
moshemorad: Let's move it to different file and app, this file is already very long and it will be helpful to have different app for the toolsets path from befining.

server.py
```diff
@@ -429,6 +442,130 @@ def chat(chat_request: ChatRequest, http_request: Request):
...
+def validate_toolset(request: ValidateToolsetRequest):
```
moshemorad: Please look at `run_toolset_config_tui` and the `run_config_test` Looks like it is implementing the exact logic that we need.

Can we move it to common place and reuse?

server.py
```diff
@@ -429,6 +442,130 @@ def chat(chat_request: ChatRequest, http_request: Request):
...
+    """Validate a toolset configuration by running check_prerequisites without deploying."""
+    try:
```
moshemorad: Please add tests. If you need some brainstorm about it let me know.

tests/test_toolsets_api.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def test_empty_toolsets_returns_400(self, client):
+        response = client.post(
+            "/api/toolsets/validate",
+            json={"yaml_config": "holmes:\n  toolsets: {}"},
+        )
+        assert response.status_code == 400
+        assert "No toolsets or mcp_servers" in response.json()["detail"]
+
+    @patch("holmes.toolsets.toolsets_api.ToolsetManager.check_toolset_prerequisites")
```
moshemorad, line 52: Why we are patching this function and not the Toolset.check_prerequisites function? Lets patch the actual function that do the logic and not the one the uses it. Unless im missing a good reason for that.

holmes/toolsets/toolsets_api.py
```diff
@@ -0,0 +1,164 @@
+import copy
+import logging
+import threading
+from typing import Optional
```
moshemorad: Unused?

holmes/toolsets/toolsets_api.py
```diff
@@ -0,0 +1,164 @@
...
+    _REFRESH_EVENT = refresh_event
+
+    main_app.mount("/api/toolsets", toolsets_app)
+
+
+@toolsets_app.post("/validate")
+def validate_toolset(request: ValidateToolsetRequest) -> ValidateToolsetResponse:
+    """Validate a toolset configuration by running check_prerequisites without deploying."""
+    try:
+        # 1. Parse the YAML string
```
moshemorad: Both step 1 & 2 can be achived with pydantic model, consider it.

holmes/toolsets/toolsets_api.py
```diff
@@ -0,0 +1,164 @@
...
+        loaded_by_name = {t.name: t for t in executor.toolsets} if executor else {}
+        loaded_names = list(loaded_by_name.keys())
+
+        # 5. Split into known (already-loaded) toolsets vs custom/MCP toolsets
+        known_overrides = {}
+        custom_dict = {}
+        for name, cfg in combined.items():
+            if not isinstance(cfg, dict):
+                raise HTTPException(status_code=400, detail=f"Config for '{name}' must be a mapping, got {type(cfg).__name__}")
+            resolved_name = handle_deprecated_toolset_name(name, loaded_names)
```
moshemorad: Are you sure we need it? the deprecated names are old configs. why we will send deprecated name from the UI in that case?

holmes/toolsets/toolsets_api.py
```diff
@@ -0,0 +1,164 @@
...
+            if resolved_name in loaded_by_name:
+                known_overrides[resolved_name] = cfg
+                logging.info(f"Toolset '{name}' (resolved: '{resolved_name}') found in loaded toolsets")
+            else:
+                if cfg.get("type") is None:
+                    cfg["type"] = ToolsetType.CUSTOMIZED.value
```
moshemorad, line 84: Can you explain the reason to mark it as customized? not sure i fully understand it.

holmes/toolsets/toolsets_api.py
```diff
@@ -0,0 +1,164 @@
...
+        toolsets_to_check = []
+        results = []
+
+        # 6. Handle known toolset overrides — deep copy only the matched toolsets, then apply config
+        if known_overrides:
+            try:
+                override_toolsets = load_toolsets_from_config(known_overrides, strict_check=False)
+                for override in override_toolsets:
+                    if override.name in loaded_by_name:
+                        toolset_copy = copy.copy(loaded_by_name[override.name])
```
moshemorad: Please use here deep copy and not copy.
i think we did the same in the termianl so also worth to fix it there (even though that there we are setting the config to new object and not overrider so no harm, but let's be on the safe side.) 

Other option just use model.model_copy(deep=true)

holmes/toolsets/toolsets_api.py
```diff
@@ -0,0 +1,164 @@
...
+                        toolset_copy.override_with(override)
```
moshemorad: why we do the override with and not just updating the config?

holmes/toolsets/toolsets_api.py
```diff
@@ -0,0 +1,164 @@
...
+                logging.error(f"Failed to load toolset overrides: {e}", exc_info=True)
+                for name in known_overrides:
+                    results.append(ValidateToolsetResult(
+                        toolset_name=name,
+                        status="invalid",
+                        error=f"Failed to load toolset config: {e}",
+                    ))
+
+        # 7. Handle custom/MCP toolsets
+        if custom_dict:
```
moshemorad: I think that we need to go over this flow together and redefine it. Not sure i fully understand the flow. 
Let's do it together tomorrow.

holmes/core/models.py
```diff
@@ -193,3 +193,20 @@ class ChatResponse(BaseModel):
...
+
+class ValidateToolsetRequest(BaseModel):
+    yaml_config: str = Field(
+        description="Raw YAML string containing the toolset configuration, including the 'holmes:' wrapper."
+    )
+
+
+class ValidateToolsetResult(BaseModel):
+    toolset_name: str
+    status: str = Field(description="Toolset status: 'valid' or 'invalid'")
```
moshemorad: Please use enum

moshemorad:
@claude review

holmes/toolsets/toolsets_api.py
alonelish: Right, removed "Optional" which is never used.

holmes/toolsets/toolsets_api.py
alonelish: Was not familiar with it. Now added HolmesToolsetConfig  to use the pydantic model

holmes/toolsets/toolsets_api.py
alonelish: I also though about it but thought we might get some old yaml. Thinking about it again this is not likely since the code will be live when the data sources json is already deployed - i will remove it.

holmes/toolsets/toolsets_api.py
alonelish: It's a fallback default  if a toolset isn't a known builtin and doesn't have an explicit type set, it defaults to CUSTOMIZED. This mirrors the logic in ToolsetManager._load_toolsets_from_config

holmes/toolsets/toolsets_api.py
alonelish: There is an issue with deep copy and it turns out there is an issue with copy.copy as well. Lets disucss tomorrow. I think we might have to go back to the load all solution.

holmes/core/models.py
alonelish: done

tests/test_toolsets_api.py
alonelish: I checked this with claude. Its a problem since check_toolset_prerequisites works with a thread pool. Here is the explenation: patching check_prerequisites (the instance method on Toolset) wouldn't work cleanly here because our endpoint calls ToolsetManager.check_toolset_prerequisites (the class method), which internally submits toolset.check_prerequisites to a thread pool. We'd have to patch it on the Toolset class itself, and it would affect all toolset instances globally.

Patching ToolsetManager.check_toolset_prerequisites is the right level - it's the single entry point our endpoint calls, and the mock can directly set status on the toolsets passed in, which is exactly what the real method does.

holmes/toolsets/toolsets_api.py
alonelish: I guess this is safer in case more props are added later and also used in the code in several places where tools are copied today.

PR #1844, opened 2026-03-26
ROB-3586 downgrade litellm version

RoiGlinik:
Downgrade version to avoid a regression.

this is the revert commit , also include the original bug introducing commit .
https://github.com/BerriAI/litellm/pull/20025

## Summary by CodeRabbit

* **Chores**
  * Adjusted a package dependency version: the `litellm` runtime constraint was changed from 1.81.3 to 1.81.0, which may affect underlying library behavior.


RoiGlinik:
@JiwaniZakir , Theres is a regression in the .3 version 

Bug introduced in https://github.com/BerriAI/litellm/pull/19190 (Caching session which breaks when changing models.) There is a revert fix for it , but after litellm security issue in the repo we are waiting with updating. So its temporary.

PR #1871, opened 2026-04-01
fix: change compaction continuation marker role from system to user

basic-system:
## Problem

After conversation history compaction, some OpenAI-compatible LLM providers reject the request with:

```
BadRequestError: System message must be at the beginning.
```

This happens because `compact_conversation_history()` appends a continuation marker with `role=system` at the **end** of the compacted message array:

```
1. system   — original system prompt
2. user     — last user prompt  
3. assistant — compacted conversation summary
4. system   — "...has been compacted. Continue."  ← BREAKS HERE
```

Many LLM providers (vLLM, certain OpenAI-compatible endpoints, custom model servers) enforce that all `system` messages must appear at the beginning of the conversation. The system message in position 4 violates this constraint.

## Fix

Change the continuation marker role from `system` to `user`. This message is just an instruction to continue after compaction — it does not need system-level authority.

## Error trace (from production)

```
LLM BadRequestError on model=openai/emma-coder-v11 (streaming iteration 6):
litellm.BadRequestError: OpenAIException - System message must be at the beginning.

File "/app/holmes/core/tool_calling_llm.py", line 825, in call_stream
    full_response = self.llm.completion(
File "/app/holmes/core/llm.py", line 608, in completion
    result = litellm_to_use.completion(
openai.BadRequestError: Error code: 400 - {"error": {"message": "System message must be at the beginning."}}
```

The error occurs after successful compaction (193527 → 40526 tokens), on the next LLM call with the compacted messages.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of conversational continuation markers so synthetic continuation messages are not mistaken for real prior user prompts, improving accuracy when compacting conversation history and preserving the correct role for continuation messages.
* **Tests**
  * Updated unit tests to cover cases around continuation markers and last-user prompt selection.


moshemorad:
@claude review

basic-system:
Thanks for the thorough review! Both issues raised by @claude are addressed in f810f58d:

1. **Re-compaction regression** — extracted the continuation marker text into a module-level constant `COMPACTION_CONTINUATION_MARKER` and updated `find_last_user_prompt()` to skip messages whose content equals it. This preserves the original user question across multiple compaction cycles, while keeping the marker's role as `user` so providers that require all `system` messages at the beginning of the conversation continue to accept the request. Added a non-Azure regression unit test (`test_find_last_user_prompt_skips_compaction_continuation_marker`) that simulates a post-compaction history and asserts the real user question is returned, not the marker.

2. **Stale Azure-gated assertions** — updated both assertions in `tests/core/truncation/test_compaction.py` from `role == "system"` to `role == "user"` to match the new marker role.

PR #1900, opened 2026-04-12
[ROB-2694] Oauth

Avi-Robusta:
## Summary by CodeRabbit

* **New Features**
  * Full OAuth support for MCP servers: PKCE/browser CLI flows, authorization-code callback, per-user/conversation token caching, background refresh, dynamic tool loading after auth, and token persistence options (memory/disk/DB).

* **Bug Fixes**
  * Improved tool lookup, registration and replacement behavior; clearer tool-change logging.

* **Documentation**
  * Added OAuth setup guide with CLI and Helm examples and browser-accessibility notes.

* **Tests**
  * Extensive unit/integration tests and Kubernetes fixtures for MCP OAuth flows.

* **Chores**
  * Added a `manual` pytest marker and CI-aware skipping of manual tests.


holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -211,7 +257,7 @@ async def get_initialized_mcp_session(
     elif toolset._mcp_config.mode == MCPMode.SSE:
         url = str(toolset._mcp_config.url)
         httpx_factory = create_mcp_http_client_factory(toolset._mcp_config.verify_ssl)
-        rendered_headers = toolset._render_headers(request_context)
+        rendered_headers = _inject_oauth_token(toolset, request_context, toolset._render_headers(request_context))
```
moshemorad, line 260: Why isn't the `toolset._render_headers` calls to `_inject_oauth_token` and append the headers?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -543,6 +720,29 @@ class RemoteMCPToolset(Toolset):
...
+    @property
+    def connect_tool_name(self) -> str:
+        """The name of the OAuth placeholder tool for this MCP server."""
+        return f"{self.name}_connect"
+
+    def get_oauth_config(self) -> Optional[Dict[str, Any]]:
+        """Return OAuth config dict for syncing to DB/frontend, or None if not OAuth-enabled."""
+        if not self.is_oauth_enabled:
+            return None
+        oauth = self._mcp_config.oauth
```
moshemorad, line 736: Why we aren't returning the model itself and created dict? if dict is needed why not just model.model_dump()

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -660,6 +866,139 @@ def prerequisites_callable(self, config) -> Tuple[bool, str]:
...
             )
 
+    def _check_oauth_server_reachable(self) -> Tuple[bool, str]:
+        """For OAuth MCP servers, verify reachability without authenticating.
+
+        If a cached token exists (from a previous request in the same conversation),
+        load the real tools directly. Otherwise, auto-discover OAuth endpoints if needed,
+        then register a placeholder tool that triggers the OAuth flow on first use.
+        """
+        assert isinstance(self._mcp_config, MCPConfig)
```
moshemorad, line 876: Do we really want asserts? it will raise exception.
In which cases this could happen?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -660,6 +866,139 @@ def prerequisites_callable(self, config) -> Tuple[bool, str]:
...
+        assert self._mcp_config.oauth is not None
+        url = str(self._mcp_config.url).rstrip("/")
+
+        # If we already have a cached token (cache → DB → disk), try to load real tools directly
```
moshemorad, line 880: This is to support the cli right? no valid use cases in our server mode?

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -660,6 +866,139 @@ def prerequisites_callable(self, config) -> Tuple[bool, str]:
...
+        oauth_config = self._mcp_config.oauth
+        disk_key = str(self._mcp_config.url)
+
+        if _get_token_manager().get_access_token(oauth_config, None, disk_key=disk_key):
+            try:
+                tools_result = asyncio.run(self._get_server_tools())
```
moshemorad, line 886: Lets not duplicate code please extract and reuse with toolset loading tools

server.py
```diff
@@ -400,6 +430,12 @@ def chat(chat_request: ChatRequest, http_request: Request):
             tracer=server_tracer,
             tool_results_dir=tool_results_dir,
         )
+
+        # Inject authenticated OAuth tools if the user has cached tokens
+        if chat_request.user_id and config._has_oauth_toolsets:
```
moshemorad, line 435: Why this is in the endpoint level and not in the llm/tool executor itself?

server.py
```diff
@@ -274,6 +279,27 @@ async def log_requests(request: Request, call_next):
...
 
+@app.post("/api/oauth/callback")
+def oauth_callback(request: OAuthCallbackRequest) -> OAuthCallbackResponse:
+    logging.info(
+        "OAuth callback: toolset=%s client_id=%s client_secret_present=%s code_present=%s code_verifier_present=%s redirect_uri=%s",
+        request.toolset_name, request.client_id, bool(request.client_secret), bool(request.code),
+        bool(request.code_verifier), request.redirect_uri,
+    )
+    try:
+        toolsets = config.create_tool_executor(dal=dal).toolsets
```
moshemorad, line 290: shouldn't it use the same executor which was already created? why it isn't passing the reuse=true?

holmes/core/tools.py
```diff
@@ -263,9 +263,13 @@ class ToolInvokeContext(BaseModel):
         str
     ] = []  # Bash prefixes approved during this session
     request_context: Optional[Dict[str, Any]] = None
+    tool_executor: Optional[Any] = None  # ToolExecutor reference for dynamic tool registration (e.g., OAuth MCP)
```
moshemorad, line 266: @Avi-Robusta i think that this could be removed when we will move the actual tool loading from oauth from the endpoint level to the get_tool in the llm call function. 

If i get it correctly this is currently here since we are making the tool decisions and we want to update the tools for the confirm oauth but since we are loading the tools in the endpoint before this was called it still contain the old connect tool.

I think that the flow should be run the confirm tool and only than later in the flow refresh the toolsets. WDYT?

holmes/core/tool_calling_llm.py
```diff
@@ -153,6 +154,23 @@ def extract_bash_session_prefixes(messages: List[Dict[str, Any]]) -> List[str]:
...
+    decision_data: Dict[str, Any],
+    request_context: Optional[Dict[str, Any]],
+) -> None:
+    """Process an OAuth callback from a tool approval decision.
+
+    Uses the pending exchange (which holds the PKCE code_verifier generated
+    when Holmes sent the approval request) to exchange the auth code for tokens.
+    """
+    try:
+        payload_json = json.dumps(decision_data)
```
moshemorad, line 168: re dumping it just to load it later on the complete exchange function? im missing something?

holmes/core/supabase_dal.py
```diff
@@ -945,3 +946,59 @@ def finish_scheduled_prompt_run(
...
+        if not self.enabled:
+            return None
+        try:
+            query = (
+                self.client.table(OAUTH_TOKENS_TABLE)
+                .select("*")
+                .eq("account_id", self.account_id)
+                .eq("provider_name", provider_name)
+            )
+            if user_id:
```
moshemorad, line 963: To be honest i don't like it, it think that this case is only valid case in the cli and in server it should be considered as bug. 
e.g. if we didn't pass the user_id in the request for some reason (could be a bug for example) it will mean that we can share context between users and this is a no go. 

Let's rethinkg and maybe separate the token storage for cli from server.

holmes/core/tools_utils/tool_executor.py
```diff
@@ -127,3 +129,35 @@ def get_all_tools_openai_format(
                 continue
             tools.append(tool.get_openai_format())
         return tools
+
+    def with_oauth_tools(
```
moshemorad, line 133: Please use clone_with_extra_tools since most of the logic is duplicate.

holmes/core/tool_calling_llm.py
```diff
@@ -282,17 +300,21 @@ def _execute_tool_decisions(
...
         for tool_call_with_decision in pending_tool_calls:
             tool_call = tool_call_with_decision.tool_call
-            decision = tool_call_with_decision.decision
+            tool_decision = tool_call_with_decision.decision
             tool_result: Optional[ToolCallResult] = None
-            if decision and decision.approved:
+            if tool_decision and tool_decision.approved:
+                # Process OAuth decision if present (auth code from frontend browser OAuth flow)
+                if tool_decision.decision:
+                    _try_process_oauth_decision(tool_call.id, tool_decision.decision, request_context)
```
moshemorad, line 316: We are trying it for each tool decition? we can't do it only for the real ones?
also why it isn't part of the invoke and we are doing it here?

holmes/plugins/toolsets/mcp/oauth_token_manager.py
```diff
@@ -0,0 +1,490 @@
...
+        if manager.has_token(oauth_config, request_context):
+            ...
+
+        # Shutdown background refresh thread
+        manager.shutdown()
+    """
+
+    def __init__(self) -> None:
+        self._cache = OAuthTokenCache()
+        self._disk_store = DiskTokenStore()
```
moshemorad, line 57: We really want to save it on disk in server?

holmes/config.py
```diff
@@ -115,6 +118,7 @@ class Config(RobustaBaseConfig):
     _cached_tool_executor: Optional[ToolExecutor] = None
     _cached_executor_key: Optional[tuple] = None
     _executor_lock: threading.Lock = PrivateAttr(default_factory=threading.Lock)
+    _has_oauth_toolsets: bool = False
```
moshemorad, line 121: Why this is part of the config? make sense to move it to the tool_executor unless im missing specific reason.

moshemorad (changes requested):

moshemorad:
@claude review

holmes/core/models.py
```diff
@@ -195,6 +211,7 @@ class ChatRequestBaseModel(BaseModel):
     trace_span: Optional[Any] = (
         None  # Optional span for tracing and heartbeat callbacks
     )
+    user_id: Optional[str] = None  # User ID from relay session token validation
```
moshemorad, line 214: @Avi-Robusta let's open a tech debt to use user bearer token instead of sending the user id and assume it is authenticated.

holmes/plugins/toolsets/mcp/oauth_token_manager.py
```diff
@@ -0,0 +1,411 @@
...
+            )
+
+        logger.debug(
+            "OAuthTokenManager: token stored (cache_key=%s, expires_in=%s, has_refresh=%s)",
+            cache_key, expires_in, "refresh_token" in token_data,
+        )
+
+    def shutdown(self) -> None:
+        """Stop the background sweep thread."""
+        self._shutdown_event.set()
```
moshemorad, line 243: If its marked as deamon why do we need this?

holmes/plugins/toolsets/mcp/oauth_token_store.py
```diff
@@ -0,0 +1,416 @@
...
+        except Exception:
+            logger.warning("Failed to store token to DB", exc_info=True)
+            return False
+
+    def get_all_for_preload(self) -> List[Dict[str, Any]]:
+        signing_key_hash = self._get_signing_key_hash()
+        if not signing_key_hash:
+            return []
+
+        db_tokens = self._dal.get_all_oauth_tokens_for_cluster(signing_key_hash)
```
moshemorad, line 298: Why we need to fetch all tokens and not only those that are going to be expired?

holmes/plugins/toolsets/mcp/oauth_token_manager.py
```diff
@@ -0,0 +1,411 @@
...
+            self._shutdown_event.wait(timeout=OAUTH_CREDENTIAL_INTERVAL_SECONDS)
+            if self._shutdown_event.is_set():
+                break
+            try:
+                self.preload_from_store()
+                self._sweep_expiring_tokens()
+            except Exception:
+                logger.warning("OAuthTokenManager: sweep failed", exc_info=True)
+
+    def _sweep_expiring_tokens(self) -> None:
```
moshemorad: My intuation when first reading this function name that it is only cleaning the token and not refresh them. Why it is called sweep?

holmes/plugins/toolsets/mcp/oauth_token_store.py
```diff
@@ -0,0 +1,416 @@
...
+
+from holmes.core.config import config_path_dir
+
+logger = logging.getLogger(__name__)
+
+
+# ── In-memory cache ───────────────────────────────────────────────────────
+
+
+class _CachedToken:
```
moshemorad, line 28: What doesn class name starts with _ means?

holmes/plugins/toolsets/mcp/oauth_token_store.py
```diff
@@ -0,0 +1,416 @@
...
+            expiry = None
+            if token_data.get("expires_in"):
+                expiry = (datetime.now(timezone.utc) + timedelta(seconds=token_data["expires_in"])).isoformat()
+
+            self._dal.upsert_oauth_token(
+                provider_name=provider_name or "unknown",
+                encrypted_token=encrypted,
+                signing_key_hash=signing_key_hash,
+                token_expiry=expiry,
+                user_id=user_id,
```
moshemorad, line 272: I don't think it is good idea to enable user_id None to be saved in the db. lets not save user_id none in the db even if we aren't reading it. It means there is bug in that case.

holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -275,6 +379,46 @@ def _invoke(self, params: dict, context: ToolInvokeContext) -> StructuredToolRes
...
+
+            if real_tools:
+                tool_names = [t.name for t in real_tools]
+                logger.info("OAuth MCP %s: loaded %d tools after authentication: %s", self.toolset.name, len(real_tools), tool_names)
+                return StructuredToolResult(
+                    status=StructuredToolResultStatus.SUCCESS,
+                    data=f"Successfully authenticated and discovered {len(real_tools)} tools: {', '.join(tool_names)}. You can now call these tools directly.",
+                    params=params,
+                    invocation=f"OAuth connect to {self.toolset.name}",
+                    oauth_tools=real_tools,
```
moshemorad, line 401: Why we need to pass it using the oauth_tools just to save it in the oauth user tools? what about just saving it in the mcp toolset level instead? any good reason to keep it out side? it will also make it easier to load it as part of the regular flow instead of call it separatly.

holmes/core/supabase_dal.py
```diff
@@ -945,3 +946,88 @@ def finish_scheduled_prompt_run(
...
+            return None
+        if not user_id:
+            return None
+        try:
+            query = (
+                self.client.table(OAUTH_TOKENS_TABLE)
+                .select("*")
+                .eq("account_id", self.account_id)
+                .eq("provider_name", provider_name)
+                .eq("user_id", user_id)
```
moshemorad, line 968: I think that we also should have here the support hash key by this instance no? other wise there can be multiple tokens with different hash key and limit one might result in different token than expected.

holmes/core/tools.py
```diff
@@ -266,6 +268,9 @@ class ToolInvokeContext(BaseModel):
 
     def model_dump(self, **kwargs):
         """Override to exclude sensitive context from serialization"""
+        kwargs.setdefault("exclude", set())
```
moshemorad: Is it still needed?

moshemorad (changes requested):
Couple of comments

holmes/core/models.py
Avi-Robusta: do you mean to open a ticket?

holmes/plugins/toolsets/mcp/oauth_token_manager.py
Avi-Robusta: used for tests, not needed in server

holmes/plugins/toolsets/mcp/oauth_token_store.py
Avi-Robusta: upsert_oauth_token doesnt
```
        if not user_id:
            logging.warning("Cannot upsert OAuth token without user_id (provider=%s)", provider_name)
            return False
```
also it shouldnt be possible to get a token without a user_id here in normal circumstances

holmes/plugins/toolsets/mcp/oauth_token_store.py
Avi-Robusta: the flow only needs to pull the tokens once, no need to pull every day or so when they need to be refreshed

holmes/plugins/toolsets/mcp/oauth_token_store.py
Avi-Robusta: like how underscore in a function or a variable is for internal use only (this class is only used in this file)

holmes/plugins/toolsets/mcp/toolset_mcp.py
Avi-Robusta: > Why we need to pass it using the oauth_tools just to save it in the oauth user tools?

All tools — including user-specific ones — that the LLM uses are managed by the `tool_executor`. We pass `oauth_tools`, and then the `tool_calling_llm` updates the executor with any newly discovered OAuth tools during the flow. (so the next llm call will have the relevant tools)

> what about just saving it in the mcp toolset level instead? any good reason to keep it out side? it will also make it easier to load it as part of the regular flow instead of call it separatly.

If I’m not mistaken, the tool executor doesn’t fetch or refresh tools between tool calls. So if we only add them at the MCP server level, those tools won’t actually be available to the LLM unless we introduce additional logic to refresh the tool set between each call.

Given that, the simplest approach is to pass the tools once upfront, rather than adding complexity to dynamically refresh tools in the executor during an investigation.

Avi-Robusta: open ticket for tech debt

moshemorad (dismissed):

moshemorad (approved):

PR #1912, opened 2026-04-14
ROB-3703: Improve Data source configuration flow

alonelish:
## Changes

- **URL Parsing**: Fix UUID extraction to handle optional dashes and format IDs correctly as dashed UUIDs required by Notion API
- **Authentication**: Correct header access from `toolset.internet_config.additional_headers` instead of `toolset.additional_headers`
- **API Version**: Add required `Notion-Version: 2022-06-28` header to requests
- **Error Handling**: Add JSON validation and better error logging for API responses
- **Block Type Support**: Expand support for multiple Notion block types:
  - Headings (heading_1, heading_2, heading_3)
  - Lists (bulleted_list_item, numbered_list_item)
  - Other blocks (quote, callout, toggle, to_do)
- **Refactoring**: Rename `parse_notion_content()` to `parse_notion_content_from_dict()` and use consolidated parsing logic with block type constants

## Summary by CodeRabbit

* **New Features**
  * Notion import now renders more block types: multi-level headings, numbered lists, quotes, callouts, toggles, and to-dos with checkboxes.
  * Connector/config schemas include richer user-facing metadata (name, description, icon, docs anchor) and return nested metadata per config.
  * Added new managed Prometheus connector options (Coralogix, Google, Grafana Cloud) for alternate endpoints.

* **Bug Fixes**
  * Improved Notion URL handling, request headers, JSON validation, clearer error reporting, and skipping of unsupported/empty blocks.


alonelish:
@claude  review

holmes/plugins/toolsets/prometheus/prometheus.py
```diff
@@ -240,13 +377,52 @@ def get_aws_client(self) -> Optional[AWSPrometheusConnect]:
...
+    _description: ClassVar[Optional[str]] = "Connect to Azure Monitor Managed Prometheus using Azure AD."
+    _icon_url: ClassVar[Optional[str]] = "https://raw.githubusercontent.com/gilbarbara/logos/de2c1f96ff6e74ea7ea979b43202e8d4b863c655/logos/microsoft-azure.svg"
+    _docs_anchor: ClassVar[Optional[str]] = "azure-managed-prometheus"
+    _subtype: ClassVar[Optional[str]] = PrometheusSubtype.AZURE_MANAGED_PROMETHEUS.value
+    # These fields are Optional at the Pydantic level so managed identity
+    # (azure_use_managed_id=True) and the env-var fallback in __init__ keep
+    # working, but the UI form requires them — users configuring this
+    # through the frontend are expected to paste values directly. CLI/Helm
+    # users can omit them and rely on env vars or managed identity.
+    _ui_required_fields: ClassVar[List[str]] = [
```
moshemorad, line 390: I think that this is good example for some ai slop in our code.
If i get it correctly those field are mark as optional since we would like to allow provide them using env var and not only through the config. 

The issue is that we are doing it in the init which happens after the pydantic validation and hence we marked them as optional with default values. 

If we will move those fallbacks to the pre validation of pydantic we will be able to get rid of:
A. Adding this new field
B. Remove all the code to validate that all of those fields which are really required will be provided and not check that as part of the __init__ function but as part of the native pydantic logic.

holmes/plugins/toolsets/confluence/confluence.py
```diff
@@ -19,51 +20,251 @@
...
+    """Confluence Cloud — hosted at yourcompany.atlassian.net with an API token."""
+
+    _name: ClassVar[Optional[str]] = "Confluence Cloud"
+    _description: ClassVar[Optional[str]] = (
+        "Confluence Cloud hosted at <your-company>.atlassian.net, authenticated with an API token."
+    )
+    _icon_url: ClassVar[Optional[str]] = CONFLUENCE_ICON_URL
+    _docs_anchor: ClassVar[Optional[str]] = "confluence-cloud"
+    _subtype: ClassVar[Optional[str]] = ConfluenceSubtype.CLOUD.value
+    _hidden_fields: ClassVar[List[str]] = [
```
moshemorad: This is very confusing, looks like all the Confluence toolsets which inheriting the ConfluenceConfig are hidding those 3 fields. 
If this is the case why we are keeping it in the base config? shouldn't we just remove it?

holmes/plugins/toolsets/datadog/datadog_models.py
```diff
@@ -21,22 +29,26 @@ class DataDogStorageTier(str, Enum):
...
         description="Default maximum number of results to return when a limit is not explicitly provided",
     )
 
 
 class DatadogTracesConfig(DatadogBaseConfig):
     """Configuration for Datadog traces toolset."""
 
+    # Hide list-typed advanced fields from the frontend form and example YAML.
+    # The runtime still accepts them via raw YAML for users who need to override.
+    _hidden_fields: ClassVar[List[str]] = ["indexes"]
```
moshemorad, line 50: Can you explain the use case here? I do see that the config contains the indexes field, why we want to hide it in the UI? to make it simpler?

holmes/plugins/toolsets/confluence/confluence.py
```diff
@@ -19,51 +20,251 @@
...
+        "auth_type",
+        "api_path_prefix",
+        "cloud_id",
+    ]
+    _recommended: ClassVar[bool] = True
+
+    api_url: str = Field(  # type: ignore[assignment]
```
moshemorad, line 81: Just to make sure i get it right, we have this here and also in the base config since we want to override the description?

holmes/core/tools.py
```diff
@@ -1059,14 +1069,60 @@ def get_config_example(self) -> Optional[Dict[str, Any]]:
...
+                # required. De-dup while preserving order.
+                seen = set(existing)
+                for name in ui_required:
+                    if name not in seen:
+                        existing.append(name)
+                        seen.add(name)
+                raw_schema["required"] = existing
+            result[config_cls.__name__] = {
+                "schema": raw_schema,
+                "name": getattr(config_cls, "_name", None) or config_cls.__name__,
```
moshemorad: Why aren't we creating a base class for ToolsetConfig with all this logic and field there?

That way it will be also easier for other contributers to know about those fields and set them.

holmes/utils/pydantic_utils.py
```diff
@@ -52,6 +52,25 @@ class MyToolsetConfig(ToolsetConfig):
     model_config = ConfigDict(extra="allow")
 
     _deprecated_mappings: ClassVar[Dict[str, Optional[str]]] = {}
+    _name: ClassVar[Optional[str]] = None
+    _description: ClassVar[Optional[str]] = None
+    _icon_url: ClassVar[Optional[str]] = None
+    _docs_anchor: ClassVar[Optional[str]] = None
+    _hidden_fields: ClassVar[List[str]] = []
```
moshemorad, line 59: Why aren't we using json_extra for the hidden_fields & ui_required_field?

holmes/utils/pydantic_utils.py
```diff
@@ -52,6 +52,25 @@ class MyToolsetConfig(ToolsetConfig):
...
+    # or runtime fallback populates it later). The UI marks the field as
+    # required; Pydantic stays permissive so CLI / Helm configs with
+    # alternative credential sources keep working.
+    _ui_required_fields: ClassVar[List[str]] = []
+    _recommended: ClassVar[bool] = False
+    # Stable identifier for this variant, used to populate meta.subtype on the
+    # synced HolmesToolsStatus row so the frontend can distinguish catalog
+    # cards that share a backend toolset name (e.g. Prometheus vs VictoriaMetrics,
+    # both backed by prometheus/metrics).
+    _subtype: ClassVar[Optional[str]] = None
```
moshemorad, line 73: Do we really need to have the _subtype field here as well if it is already being set on the toolset class? 
Not sure i fully understand why we need both of them is it for the rendering in the UI?

moshemorad (changes requested):
Awesome work, left couple of comments. 
Let me know if something isn't clear and let's go over it together.

holmes/plugins/toolsets/prometheus/prometheus.py
alonelish: The problem of dropping the ui_required_fields for the Azure Prometheus config is the "azure_use_managed_id" switch which makes all 3 required field, not needed.

**It is explained by Claude below:** 

Moshe's diagnosis is correct: move env-var resolution to @model_validator(mode="before"), which runs against the raw input dict before any field is validated, and the cascade collapses. Pydantic itself enforces presence, the schema's native required array is correct, the __init__ check disappears, and _ui_required_fields is no longer needed — for fields that are unconditionally required.

The catch the proposal doesn't address: conditional requirements
azure_use_managed_id=True means none of the three credentials are required. So a clean "make these fields non-Optional" doesn't work — they're only required when azure_use_managed_id is False.

That has two consequences:

You still need a mode="after" validator for the XOR check (same shape as ServiceNow's api_key XOR (username+password) at servicenow_tables.py:94-108). So the manual presence check moves, it doesn't vanish — but it does become idiomatic Pydantic instead of __init__ magic.
Pydantic's native JSON-schema required array can't express "required only when X". The frontend currently relies on _ui_required_fields to flag the credential fields as required in the form. If you drop _ui_required_fields and leave the fields Optional, the UI form will let users submit empty credentials with azure_use_managed_id=False and only learn about the failure on submit. To do better you'd need either:
A custom __get_pydantic_json_schema__ that emits if/then/required or oneOf, or
Keep _ui_required_fields (or some equivalent) as a UI hint — accepting that it duplicates info the model already has.

holmes/plugins/toolsets/confluence/confluence.py
alonelish: OK, fixed now

holmes/plugins/toolsets/datadog/datadog_models.py
alonelish: Why it's hidden
Two reasons, in order of weight:

Frontend form can't render list[str] cleanly. The schema-driven config form renders scalars (strings, numbers, enums, booleans) and produces sensible inputs. A list[str] either gets a textarea-with-JSON, a "comma-separated" hack, or nothing at all — every option is a worse UX than not showing it. The in-code comment says exactly this: "complex list types don't render as form inputs."

The default ["*"] is right for ~95% of users. Hiding a field whose default works for most people is a reasonable UI policy — it keeps the form short and on-task.

The escape hatch is raw YAML: power users who need to scope to specific indexes drop into the ~/.holmes/config.yaml and add indexes: [...]. The runtime accepts it because it's a real Pydantic field.

holmes/plugins/toolsets/confluence/confluence.py
alonelish: Yes, not only the description, the example url as well. That said, i did remove the defaults in the base class and made it more compact with no default value:

    api_url: str
    user: Optional[str] = None
    api_key: str

holmes/core/tools.py
alonelish: What i did is move the logic in "get_config_schema" to the ToolsetConfig and left it as a thin wrapper in Toolset.

This way the logic of loading all the ToolsetConfig attributes (subtype, hidden_fields etc.) is contained in the same class. Makes sense?

alonelish: Also, in the process had to change HttpToolsetConfig  to extend ToolsetConfig and not BaseModel

holmes/utils/pydantic_utils.py
alonelish: I consulted with Claude about the pros and cons of this idea. 

**Where it gets awkward**

For fields a variant inherits but wants to hide — like ConfluenceDataCenterPATConfig hiding user (which is declared on the base) — there's no field declaration to attach the flag to. The variant would have to redeclare:

class ConfluenceDataCenterPATConfig(ConfluenceConfig):
    user: Optional[str] = Field(  # type: ignore[assignment]
        default=None,
        json_schema_extra={"hidden": True},
    )
…purely to flip the hidden flag. Six lines of boilerplate to replace one list entry. That's the only case in the current codebase where it's awkward.

The suggestion is a clean win for fields a variant declares itself, and a regression for inherited fields.

holmes/utils/pydantic_utils.py
alonelish: We need both because they live on different sides of the same conversation:

**ToolsetConfig._subtype** (ClassVar on each variant config class) — declaration. Says "this variant is called cloud." Read by build_schema_entry and sent to the frontend so the catalog can show one card per variant and know which slug to write back.

**Toolset.subtype** (instance attribute on the Toolset) — utterance. Carries the user's pick from their YAML (subtype: cloud) into the runtime, where prerequisites_callable uses it to look up the matching variant class.

One is "what slug am I" (per class, design-time). The other is "which slug did the user pick" (per instance, load-time). Same value when things work; they meet in the middle.

holmes/plugins/toolsets/datadog/datadog_models.py
moshemorad: Gotcha, so we want to reduce the complexity of setting up this toolset by removing advance field?

holmes/utils/pydantic_utils.py
moshemorad: Maybe this is the initial problem, why we keep having base class that contains field which aren't relevant to the inheriting class?

holmes/plugins/toolsets/prometheus/prometheus.py
moshemorad: @alonelish But now in the UI when im setting the Azure Use Managed Id it is still expecting those field to provided which doesn't make sense

image

holmes/plugins/toolsets/confluence/confluence.py
moshemorad: Just wanted to make sure i understand the case, didn't have any objection to keep the default desc etc on the base class. Maybe worth to keep them so if someone will inheriting it again it will have something even if he didn't added desc etc.

holmes/utils/pydantic_utils.py
moshemorad: I think that this was something we discuss as part of the db task, if we should keep it on the toolset level or the config. we decided to keep it on the toolset level and now anyway we added it to the config level, so we have it twice. 

Maybe we should change it in the future to only be on the config level (Don't do it as part of this task for now)

holmes/plugins/toolsets/prometheus/prometheus.py
alonelish: You are right, this is a limit of the current UI - it does not support conditionally showing elements based on other elements. This would significantly complicate the schema i think and was beyond the scope of this change - that said this is kind of a rare case and the user will have to rely on the docs

holmes/plugins/toolsets/datadog/datadog_models.py
alonelish: yes, from what i understand it is highly unlikely that the user will want to customize it and wanted to keep the YAML minimal.

holmes/plugins/toolsets/confluence/confluence.py
alonelish: Maybe its better that they dont have. it will make them come with their own once they try it out, no?

holmes/utils/pydantic_utils.py
alonelish: But user is used by the 2 other variants as explained below:

Why _hidden_fields instead of just removing the field
user is declared on the base ConfluenceConfig as Optional[str] = None. Cloud and DC_Basic redeclare it as required (user: str). DC_PAT inherits the base's Optional version. There's no way in Pydantic v2 to remove an inherited field from a subclass — you can override its type or default, but you can't delete it.

Since the runtime uses _hidden_fields to strip entries from the JSON schema's properties and required arrays before the schema reaches the frontend, _hidden_fields = ["user"] is the clean way to say:

"DC PAT inherits this field for type/Pydantic reasons, but the form-filling user shouldn't be asked about it."

Maybe we could add a parent class for the 2 that use it to make it nicer. Not sure i would do it now.

alonelish: Moved the "user" from the base class to the 2 classes that actually use it and removed it from hidden_fields

moshemorad (approved):

Issue #1907, opened 2026-04-13
RabbitMQ credentials are not set from env variable


nayanparegi:
### What happened?

RabbitMQ - credentials are nested inside a list item under clusters:
Holmes resolves {{ env.VAR }} in flat config fields but not recursively inside list items

### What did you expect to happen?

It should able to resolve credential from env variable so I can set those in the secret instead of directly putting those in configmap

### How can we reproduce it (as minimally and precisely as possible)?

try to set credential of RaibitMQ toolset using Kubernetes secret

moshemorad:
Hi @nayanparegi,
Thanks for reporting.

Not sure i fully understand the issue, there are 2 points in the issue.

1. `Holmes resolves {{ env.VAR }} in flat config fields but not recursively inside list items` - Can you give example? what does it mean?
2. It should able to resolve credential from env variable so I can set those in the secret instead of directly putting those in configmap - you can reference secret as env var in k8s more info here: https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#define-container-environment-variables-using-secret-data

dattjigneshdixit-arch:
Hi @moshemorad   
Here's a  example of the issue:                                                                                            
  
   Works  for {{ env.VAR }} in a flat config map (e.g., Elasticsearch toolset):                                                          
   toolsets:     
     elasticsearch/data:                                                                                                              
       config:                                                                                                                        
         username: "{{ env.ELASTICSEARCH_USERNAME }}"   #  resolved
         password: "{{ env.ELASTICSEARCH_PASSWORD }}"   #  resolved                                                                  
                                                                                                                                      
   Does NOT work  for {{ env.VAR }} inside a list item under config.clusters (RabbitMQ toolset):                                         
   toolsets:                                                                                                                          
     rabbitmq/core:                                                                                                                   
       config:   
         clusters:
           - id: rabbitmq
             username: "{{ env.RABBITMQ_USERNAME }}"   #  NOT resolved                                                               
             password: "{{ env.RABBITMQ_PASSWORD }}"   #  NOT resolved
             api_url: "{{ env.RABBITMQ_API_URL }}"             #  NOT resolved                                                               
                                                                                                                                      
The env vars (RABBITMQ_USERNAME, RABBITMQ_PASSWORD, RABBITMQ_API_URL) are correctly set on the Holmes pod via additionalEnvVars with secretKeyRef. We verified the env vars are present in the container.                                                          
[...]

PR #1938, opened 2026-04-22
override_with reverted env vars resolution for nested arrays

arikalon1:
## Summary by CodeRabbit

* **Bug Fixes**
  * Overrides now only apply explicitly set values, respect false/zero values, and ignore empty values (empty string/empty dict).
  * Identity fields (e.g., name) are preserved and cannot be overwritten by overrides.
  * Environment-variable templates in overrides remain unresolved until runtime.
* **Tests**
  * New tests added to validate override behavior, edge cases, and full load/override pipeline.


arikalon1:
fix #1907

moshemorad (approved):

PR #1615, opened 2026-02-22
Improve jq filtering

aantn:
https://claude.ai/code/session_01N5rWgtZrghkJVJtL3xdEGZ

## Summary by CodeRabbit

* **Tests**
  * Added a new jq-filter test marker and applied it across Elasticsearch, Grafana, Prometheus and HTTP test cases.
* **Improvements**
  * JSON filtering failures now surface structured error details plus a truncated preview of the original response to aid debugging and reduce noisy output.

PR #1947, opened 2026-04-24
fix(toolsets): fail-closed on max_depth<=0 to prevent silent truncation

### LINKED ISSUE #1946: elasticsearch_list_indices silently returns "no indices" when LLM picks max_depth=0 (affects all 6 JsonFilterMixin tools)
### TL;DR

The shared `JsonFilterMixin` used by `elasticsearch_list_indices`, `elasticsearch_mappings`, `http_request`, `grafana_get_dashboard_by_uid`, `grafana_get_home_dashboard`, and `list_prometheus_rules` **silently destroys response bodies** when the LLM chooses `max_depth=0`, while still returning `status=SUCCESS`. The LLM has no signal to retry and confidently produces **false-negative answers** ("nothing found" when data is abundant). In production this burns tokens and — more dangerously — makes HolmesGPT lie to the user without any error surface.

**HolmesGPT version:** `0.24.4`.

### How we're hitting this in production right now

Our setup:

- HolmesGPT `0.24.4`, deployed to Kubernetes (pod `holmes-holmes-5ccbb5db6c-pc7jc`).
- LLM backend: Azure AI Foundry / Azure OpenAI via LiteLLM (deployment `holmesgpt-test`).
- Toolsets enabled: `elasticsearch/cluster` and `elasticsearch/data` (default built-in toolsets).
- Target cluster: ECK-managed Elasticsearch 8.x, 4 nodes (3 data), **422 primary shards / 572 total, status green**, running APM + Filebeat + OTel collector. Hundreds of `.ds-filebeat-*`, `.ds-logs-apm.*`, `.ds-metrics-apm.*`, `.ds-traces-apm-*` data streams.

What we asked HolmesGPT:

> *"Check which of these Elasticsearch index patterns exist on the cluster and tell me their doc counts: `filebeat-*`, `logstash-*`, `.ds-logs-*`, `.ds-metrics-*`, `metricbeat-*`, `apm-*`, `.ds-traces-*`."*

What we expected: a list of patterns with matching indices and doc counts (most of those patterns match hundreds of our indices).

What actually happened:

- The LLM fired **7 parallel `elasticsearch_list_indices` calls**, one per pattern.
- Every call was sent with `max_depth: 0`.
- Every call's `tool_calls[].result.status` came back as `"success"`.
- Every call's `tool_calls[].result.data` came back as the literal string `"...truncated at depth 0"`.
- The LLM saw seven `success`-with-empty-looking-data results and concluded:

  > *"None of the Elasticsearch index patterns you asked about currently exist on the cluster."*

- Total conversation cost: **~20,000 tokens** for zero useful output.
- No error surface. No warning log. No retry signal. The user got a confident, wrong answer.

This is not a single-query glitch; it reproduces every time the LLM defaults to `max_depth=0`, which it does consistently in our deployment.

### Why it's dangerous

HolmesGPT's project contract (`CLAUDE.md`) says:

> *"All toolsets MUST return detailed error messages from underlying APIs to enable LLM self-correction."*

A silent `status=success` with a destroyed payload is the worst failure mode an agentic tool can have — it looks like truth to the downstream LLM. There is no retry, no `truncated: true` flag, no status other than `SUCCESS`. The LLM treats an empty shell as authoritative. Every caller downstream (end users, automated runbooks, Slack notifications) gets a lie dressed as a success.

### Reproducer

Against a HolmesGPT deployment with any live Elasticsearch cluster reachable by the `elasticsearch/data` toolset:

**Step 1 — baseline (toolset works):**

```bash
curl -sS http://localhost:5050/api/chat \
  -H 'Content-Type: application/json' \
  -d '{"ask":"Summarize Elasticsearch cluster health"}' | jq .
```

Returns real cluster health. Proves auth/TLS/LLM wiring.

**Step 2 — trigger the bug:**

```bash
curl -sS http://localhost:5050/api/chat \
  -H 'Content-Type: application/json' \
  -d '{"ask":"Check which of these Elasticsearch index patterns exist on the cluster and tell me their doc counts: filebeat-*, logstash-*, .ds-logs-*, .ds-metrics-*, metricbeat-*, apm-*, .ds-traces-*"}' | jq .
```

Every `tool_calls[].result.data == "...truncated at depth 0"`. The LLM's `analysis` falsely claims none exist.

**Step 3 — prove the data is there:**

```bash
curl -sS http://localhost:5050/api/chat \
  -H 'Content-Type: application/json' \
  -d '{"ask":"Use elasticsearch_list_indices with pattern=* and max_depth=5 to retrieve ALL index names. Then tell me the first 30 index names alphabetically. Do NOT use max_depth=0 as it truncates the output."}' | jq .
```

Returns 200+ indices. Same tool, same pattern, only `max_depth` differs → opposite conclusions. This is the smoking gun: the bug is purely in the filter layer, not in Elasticsearch or auth.

### Root cause

Three defects combine, all in `holmes/plugins/toolsets/json_filter_mixin.py`:

1. **Destructive truncation at depth 0** — `_truncate_to_depth(value, max_depth, current_depth=0)` at lines 21–41. At `current_depth=0` and `max_depth=0`, the guard `current_depth >= max_depth` fires immediately on the root payload. For any `dict` or `list` it is wholesale replaced with the literal string `"...truncated at depth 0"`. `_cat/indices` returns a list, so the entire index list is gone.

   Reproduced in isolation:
   ```python
   _truncate_to_depth([{"index": "foo"}, {"index": "bar"}], 0)
   # → '...truncated at depth 0'
   ```

2. **Silent status preservation** — `filter_result()` at lines 108–139 unconditionally sets `base_result.data = filtered_data` without inspecting whether the filter destroyed the payload. The upstream `SUCCESS` status from `_make_request` (HTTP 200 with real data) survives to the LLM alongside a payload that is now just a string. No `truncated` flag, no `ERROR`, no `NO_DATA`.

3. **Adversarial tool-schema description** — lines 60–64:

   ```python
   "max_depth": ToolParameter(
       description="Maximum nesting depth to return from the JSON "
                   "(0 returns only top-level keys). Leave empty for full response.",
       type="integer", required=False,
   ),
   ```

   This description is factually wrong. For a **list** response (`_cat/indices`) there are no "top-level keys" to return; `max_depth=0` replaces the whole list with a string. For a **dict** response, `max_depth=0` does not preserve keys either — it replaces the dict with a string. This is the causal trigger: the LLM reads *"0 returns only top-level keys"*, reasonably picks 0 as the cheap-summary option, and hits defects (1) and (2).

### Blast radius

Six tools inherit `JsonFilterMixin` and are all vulnerable:

| Toolset | Tool | Call site |
|---|---|---|
| elasticsearch/data | `elasticsearch_list_indices` | `holmes/plugins/toolsets/elasticsearch/elasticsearch.py:860` |
| elasticsearch/data | `elasticsearch_mappings` | `holmes/plugins/toolsets/elasticsearch/elasticsearch.py:614` |
| http | `http_request` | `holmes/plugins/toolsets/http/http_toolset.py:555` |
| grafana/dashboards | `grafana_get_dashboard_by_uid` | `holmes/plugins/toolsets/grafana/toolset_grafana.py:378` |
| grafana/dashboards | `grafana_get_home_dashboard` | `holmes/plugins/toolsets/grafana/toolset_grafana.py:404` |
| prometheus | `list_prometheus_rules` | `holmes/plugins/toolsets/prometheus/prometheus.py:747` |

A fix in the mixin protects all six simultaneously.

### Related open PRs — why the blast radius is growing

There are two open PRs that touch `JsonFilterMixin` but do **not** address this bug; they are complementary to this fix:

- **#1695 — "Add JSON filtering capability to multiple toolsets"** (open, author @aantn, last updated March 2026) extends `JsonFilterMixin` to 7 additional toolsets (Datadog, ServiceNow, MongoDB Atlas, RabbitMQ, Coralogix, New Relic, plus further Elasticsearch tools). It does **not** modify the mixin core, does **not** change `max_depth` defaults or description, and does **not** introduce a truncation-state signal. **If #1695 merges before this fix lands, every newly-covered tool inherits the same silent-truncation bug** — the blast radius roughly doubles. Strong argument for landing this fix first, or at minimum in parallel.

- **#1615 — "Improve jq filtering"** (open, author @aantn, last updated February 2026) adds a structured error payload (`jq_error`, `jq_expression`, `raw_response_preview`) when a `jq` expression fails. It touches `json_filter_mixin.py` in a different region (jq error path) from this fix (the `filter_result` guard + `max_depth` description), so any merge conflict is mechanical, not semantic.

No open issue currently tracks this bug class; `max_depth` / silent-truncation returns zero results in the HolmesGPT/holmesgpt issue search.

### Proposed fix (PR incoming)

Single file changed (plus tests): `holmes/plugins/toolsets/json_filter_mixin.py`.

1. **Rewrite the `max_depth` tool-schema description** — state the valid range (`>= 1`), point the LLM at `jq` for precise extraction, explicitly warn against `0` and negative values. Removes the lure.

2. **Fail-closed in `filter_result()` on `max_depth <= 0`** — if `params["max_depth"]` is a non-None int `<= 0` **and** the upstream status is `SUCCESS`, return an `ERROR` result with a self-corrective message ("max_depth must be >= 1 or omitted. … Retry with max_depth>=1 … or omit the parameter … or use the jq parameter …"). The guard intentionally checks `status == SUCCESS`, so genuine upstream errors (e.g. "HTTP 503: cluster unreachable") pass through verbatim — we do not clobber real failures with a parameter error.

3. **Regression tests** (in `tests/plugins/toolsets/test_json_filter_mixin.py`):
    - `max_depth=0` on SUCCESS → ERROR with self-corrective message
    - `max_depth=-1` on SUCCESS → ERROR (closes the undocumented negative-means-full escape hatch)
    - `max_depth=0` when upstream was already ERROR → upstream error preserved verbatim
    - `max_depth` omitted → full response unchanged
    - Description-wording regression: the string *"0 returns only top-level keys"* must never come back, and the description must mention `>= 1`.

No in-tree caller currently passes `max_depth <= 0` (grep-verified), so this change has zero blast radius on existing callers. It only tightens the LLM-facing contract.

### Known remaining gap (deliberately out of scope for this PR)

`max_depth=1` on a **list-of-dicts** response (the shape `_cat/indices` returns) still produces `[sentinel, sentinel, …]` with `status=SUCCESS` — same silent-truncation class of bug, one level in. Fixing that properly requires a response-envelope redesign (e.g. `{truncated: true, max_depth_used: 1, data, hint}`) that changes the response shape for all 6 consumer tools and their tests. It is a strictly larger change that deserves its own review and rollback surface.

I will open a follow-up issue specifically for the envelope redesign, and link it from this one when the PR for the `max_depth<=0` fix merges.

### Environment

- HolmesGPT: `0.24.4`
- Python: 3.11
- LLM provider: Azure AI Foundry / Azure OpenAI (via LiteLLM), deployment `holmesgpt-test`
- Elasticsearch: 8.x (ECK-managed), 4 nodes, 572 shards, green
- Deployment: Kubernetes (Helm chart)
- Toolsets enabled: `elasticsearch/cluster`, `elasticsearch/data` (built-in defaults)

### Related

- `CLAUDE.md` § "Toolset Architecture": *"All toolsets MUST return detailed error messages from underlying APIs to enable LLM self-correction."* — this bug violates that contract.
- `CLAUDE.md` § "Server-Side Filtering is Critical" — motivates the existence of `JsonFilterMixin`; this issue is about hardening that mixin against silent failure.

sebv004:
## Summary

Fixes a silent-truncation bug in `JsonFilterMixin` that made six toolset tools produce false-negative answers (e.g. `elasticsearch_list_indices` reporting "no indices exist" against a cluster with hundreds of matching indices). Full context, reproducer, and live-cluster symptom in #1946.

Root cause: when the LLM calls a mixin-backed tool with `max_depth=0` (a plausible value given the old description *"0 returns only top-level keys"*), `_truncate_to_depth` replaces the entire response with the literal string `"...truncated at depth 0"`, but `filter_result()` preserves `status=SUCCESS`. The LLM sees success + empty-looking data and confidently reports the wrong answer with no retry signal.

This PR applies the minimal fix — two small edits in one file — plus regression tests:

- **Rewrites the `max_depth` tool-schema description** so the LLM stops choosing 0. States the valid range (`>= 1`), points at `jq` for precise extraction, explicitly warns against `0` and negative values.
- **Fail-closed in `filter_result()` on `max_depth <= 0`** with a self-corrective `ERROR` message the LLM can act on. The guard only fires when the upstream call succeeded (`status == SUCCESS`), so genuine upstream errors (e.g. HTTP 503 "cluster unreachable") are preserved verbatim — no clobbering of real failures with a parameter error.
- Protects all six current consumer tools in a single mixin-level change: `elasticsearch_list_indices`, `elasticsearch_mappings`, `http_request`, `grafana_get_dashboard_by_uid`, `grafana_get_home_dashboard`, `list_prometheus_rules`. Also preemptively protects every toolset being added by #1695 (Datadog, ServiceNow, MongoDB Atlas, RabbitMQ, Coralogix, New Relic, and more) once that PR merges.

## Test plan

- [x] `poetry run pytest tests/plugins/toolsets/test_json_filter_mixin.py -v --no-cov` → **9/9 passing** (4 pre-existing tests unchanged, 5 new regression tests)
- [x] Standalone sanity check of `_truncate_to_depth` against an Elasticsearch `_cat/indices`–shaped payload: confirms bug at depth 0, confirms known remaining gap at depth 1 on list-of-dicts (see below), confirms real data returned at depth ≥ 2 and `None`
- [x] `max_depth<=0` test: returns `ERROR` with self-corrective message
- [x] `max_depth=-1` test: same (closes the undocumented negative-means-full escape hatch)
- [x] Upstream-error-preservation test: when the mocked call returns `ERROR` and the LLM (hypothetically) passes `max_depth=0`, the upstream error string survives verbatim — no clobber
- [x] `max_depth` omitted: full response unchanged (no regression on the happy path)
- [x] Description-wording regression test: the string *"0 returns only top-level keys"* can never come back, and the description must mention `>= 1`
- [ ] Maintainer to verify against a real Elasticsearch cluster using the three-step reproducer in #1946

## Known remaining gap (deliberately out of scope — follow-up issue welcome)

`max_depth=1` on a **list-of-dicts** response (the shape `_cat/indices` returns) still produces `[sentinel, sentinel, …]` with `status=SUCCESS` — the same silent-truncation class of bug, one level in. Fixing that properly requires a response-envelope redesign (e.g. `{truncated: true, max_depth_used, data, hint}`) that changes the response shape for all six consumer tools and their tests. That is a strictly larger change and deserves its own review and rollback surface, so it is deliberately out of scope here. Documented in #1946 under "Known remaining gap".

## Related PRs

- #1695 extends `JsonFilterMixin` to seven additional toolsets but does **not** touch the mixin core. If it merges before this fix, every newly-covered toolset inherits the silent-truncation bug. Complementary to this PR.
- #1615 improves `jq` error messaging in the same file but a different region. No behavioral overlap; any conflict is mechanical.

## References

Closes #1946

## Release Notes

* **Bug Fixes**
  * Added runtime validation for the depth parameter; values of 0 or below now return an error with guidance instead of being processed.

* **Documentation**
  * Updated depth parameter description to clarify that values must be >= 1; omit the parameter for a complete, untruncated response.


moshemorad (approved):
LGTM!
Thanks for your contribution.

PR #1963, opened 2026-04-29
Add retry logic to conversation worker and fix previous comments

naomi-robusta:
## Summary

This PR adds a background thread to periodically flush pending conversation events to Supabase, and implements bounded retry logic with exponential backoff for transient post failures. These changes ensure events are pushed promptly even when the consumer thread is blocked on slow LLM/tool calls, and improve resilience to temporary network issues.

## Key Changes

- **Background event flusher thread**: Added `_flusher_loop()` that runs in a daemon thread and flushes pending events at regular intervals (at least every `batch_interval_seconds`). This prevents events from being buffered for extended periods when the consumer thread is idle or blocked.

- **Retry logic with tenacity**: Extracted post logic into `_post_with_retry()` that retries transient errors (network, 5xx, DNS) up to 3 times with exponential backoff (0.5s–4s). Mismatch errors (assignee/request_sequence/status) are still surfaced immediately as `ConversationReassignedError` without retry.

- **Dual locking strategy**: 
  - `_lock` guards state (`_pending_events`, `_pending_compact`, `_last_*_time`)
  - `_flush_lock` serializes concurrent flushes between consumer and background threads to prevent overlapping network posts

- **Graceful shutdown**: The consumer thread stops the background flusher before the final drain to avoid races on the closing flush.

- **Improved DAL retry**: Refactored `get_conversation_events()` in `supabase_dal.py` to use tenacity for cleaner retry logic.

- **ConversationTask hydration**: Converted `user_message_data` and `conversation_history` to private attributes with properties to clarify they are post-construction hydrations, not part of the validated row schema.

- **Status validation**: Added `ConversationStatus.updatable_values()` class method to centralize valid status values for `update_conversation_status()`.

- **Test documentation**: Added comprehensive README documenting unit and integration test structure, prerequisites, and how to run them.

## Implementation Details

- The background flusher floors its wait interval at 50ms to prevent tight spinning in tests with zero/negative `batch_interval_seconds`.
- Transient post errors are wrapped in a new `_TransientPostError` exception to drive tenacity retries while allowing `ConversationReassignedError` to bypass retry logic.
- The consumer thread now acquires `_lock` only to snapshot state and check timing conditions, releasing it before the network call to allow the background flusher to proceed.
- Added `user_id` to `ChatRequest` construction and passed as `request_context` to `call_stream()` to ensure per-user OAuth tools resolve correctly.

https://claude.ai/code/session_018Dd2SUABJjyHSJ2vpdYu1k

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event delivery with automatic retries, exponential backoff, and preservation of pending events when transient failures occur.
  * More resilient event fetching that tolerates transient backend failures and returns safely on persistent errors.

* **Improvements**
  * Streaming now carries per-conversation user context so tools/auth resolve correctly.
  * Worker explicitly fails conversations when streaming yields no terminal result.
  * Centralized validation for conversation status updates to prevent invalid transitions.

* **Documentation**
  * Added developer guide for running unit and integration tests.


holmes/core/conversations_worker/worker.py
```diff
@@ -620,19 +622,39 @@ def _run_chat_and_publish(
...
 
                 terminal = publisher.consume(stream)
-                status = self._terminal_to_status(terminal)
-                if status in ("completed", "failed"):
+                if terminal is None:
+                    # The stream ended without a terminal event (or the
+                    # terminal batch could not be saved). Post an explanatory
+                    # error event before marking the conversation failed so
+                    # the UI shows why instead of an unexplained status flip.
+                    logging.warning(
```
moshemorad: Consider make it error so we will get sentry event and know if it happens. WDYT?

moshemorad (dismissed):

moshemorad (approved):

PR #1969, opened 2026-04-29
ROB-3618: Add AI usage tracking via HolmesUsageEvents

alonelish:
## Summary

Wires every LLM-consuming entry point through a new shared **usage recorder** so each request lands as one row in a new `HolmesUsageEvents` Supabase table. Enables per-account / per-cluster / per-user / per-feature cost reporting plus thumbs-up/down feedback on individual chat turns — none of which is answerable today (Holmes persists no AI metrics).

Driven by the AI Metrics requirements doc (Hypatos / Adobe Firefly / Playtika / Cynet asks).

## What's covered

| Entry point | Tracked? |
|---|---|
| `POST /api/chat` (direct) | ✅ |
| ConversationWorker → `chat()` | ✅ |
| ScheduledPromptsExecutor → `chat()` | ✅ (`request_type='scheduled_prompt'`) |
| `POST /api/agui/chat` (experimental) | ✅ (`request_type='agui_chat'`) |
| `POST /api/checks/execute` | ✅ (`request_type='health_check'`) |
| `POST /api/feedback` | ✅ (new endpoint, UPDATEs feedback_* columns) |
| CLI (`holmes ask`, `holmes check`, `holmes interactive`) | ❌ by design — no FastAPI server, no DAL hook |

## How it works

- **`holmes/core/usage_recorder.py`** (new) — `UsageRecorderState` dataclass + `stream_with_usage_recording` (streaming) + `record_from_llm_result` (non-streaming) + `record_error`. Single source of truth used by all four call sites. Fire-and-forget on a daemon thread so telemetry never blocks the response.
- **`SupabaseDal.record_usage_event` / `record_feedback`** write through the existing `self.client` (same email+password user-JWT pattern as `upsert_holmes_status`). **No new env vars, no service_role key.** RLS mirrors `HolmesStatus`'s `FOR ALL TO public USING (account_id IN get_user_accounts() OR is_support())` policy.
- **`ChatRequest`** gains 6 optional fields: `request_type`, `request_source`, `source_ref`, `conversation_id`, `conversation_source`, `meta`. All additive — old clients keep working.
- **`LLMResult`** gains a `finish_reason` field captured from the last LLM iteration's `response.choices[0].finish_reason`.
- **`POST /api/feedback`** accepts `{request_id, sentiment, category?, comment?}` and UPDATEs the matching event row.

## Schema (separately applied to Supabase)

`HolmesUsageEvents` covers identity (`account_id` uuid, `cluster_id`, `user_id` uuid, `conversation_id`, `conversation_source` discriminator, `request_id`), classification (`request_type`, `request_source`, `source_ref`, `model`, `provider`, `is_robusta_model`), the full `RequestStats` set (token counts, costs, peaks, compactions), shape/outcome (`iterations`, `tool_call_count`, `duration_ms`, `is_streaming`, `finish_reason`), `feedback_*` columns, and a `meta` JSONB for forward-compat.

Optional `HolmesUsageEventCalls` per-LLM-call child table is gated by `HOLMES_RECORD_LLM_CALLS=true` — table created with RLS but emission **not enabled in v1** (deferred — each check is one LLM call so child rows would near-duplicate the parent).

## Test plan

- [x] **Unit tests for the helper** — `tests/core/test_usage_recorder.py` (15 tests covering streaming/non-streaming/error/approval paths, disabled-DAL no-op, fire-and-forget threading).
- [x] **Unit tests for SupabaseDal** — `tests/core/test_supabase_dal_usage.py` (11 tests covering payload shape, account-scoped UPDATE, NULL handling, error swallowing).
- [x] **Existing tool_calling_llm tests** — all 41 still pass after adding `finish_reason` capture (one regression I introduced and fixed: `isinstance(str)` guard so `MagicMock`'d test responses don't break pydantic validation).
- [x] **169 passed / 0 failed** on the combined run of all directly-affected test files. Other failures in the broader suite are pre-existing Windows-env issues (cp1252 codec, missing `/bin/sh`, tempfile permission semantics) — verified by checking that `holmes/main.py` and other files I never touched have the same pre-existing failures.

## Deployment notes

1. Apply the Supabase migration first (`HolmesUsageEvents` + RLS).
2. Deploy this PR. Recorder writes through the existing user-JWT `self.client`, no new env vars to provision.
3. Frontend coordination (separate PR on the FE side):
   - Send `request_source` / `source_ref` / `user_id` / `conversation_id` on `/api/chat` body.
   - Same fields in the JSONB `data` of `user_message` events written to `ConversationEvents` for the worker path.
   - Reuse the same `conversation_id` the FE writes to its `ChatHistory` row so `LEFT JOIN ChatHistory` works.

## Out of scope (per plan)

- Daily / hourly rollup tables (defer until volume justifies)
- Storing prompts / completions text (PII, opt-in flag in the future)
- CLI tracking
- Per-LLM-call rows for health checks (each check is one LLM call; child row would duplicate the parent)
- RBAC-gated SELECT policy (placeholder commented in the plan, only needed if a `DB_HOLMES_USAGE_EVENTS_SELECT` permission action gets minted)

## Summary by CodeRabbit

* **New Features**
  * In-app feedback endpoint with sentiment validation and persistent feedback storage.
  * Global usage recording for AI calls (streaming and non-streaming) with request classification, conversation linkage, model/provider info, tool-call counts, finish reasons, and background persistence.

* **Bug Fixes**
  * Ensure usage events are recorded on errors, exceptions, and premature stream termination; classify rate-limited errors.

* **Tests**
  * Added comprehensive tests for usage recording and feedback persistence.


holmes/checks/checks_api.py
```diff
@@ -13,8 +13,21 @@
...
+from holmes.core.usage_recorder import UsageRecorderState
 from holmes.plugins.destinations.slack.plugin import SlackDestination
 
+
+def _resolve_provider(model: Optional[str]) -> str:
+    """Best-effort: return the canonical litellm provider for `model`."""
+    if not model:
+        return "unknown"
+    try:
+        import litellm
```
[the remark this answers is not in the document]
alonelish: Fixed in 45676569 — moved `import litellm` to module scope alongside `from litellm.exceptions import AuthenticationError`.

holmes/core/supabase_dal.py
```diff
@@ -817,6 +819,115 @@ def sync_toolsets(self, toolsets: list[dict], cluster_name: str) -> None:
...
+            self.client.table(HOLMES_USAGE_EVENTS_TABLE).update(
+                {
+                    "feedback_sentiment": sentiment,
+                    "feedback_category": category,
+                    "feedback_comment": comment,
+                    "feedback_at": datetime.now().isoformat(),
+                }
+            ).eq("account_id", self.account_id).eq(
+                "request_id", request_id
+            ).execute()
```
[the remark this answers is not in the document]
alonelish: Good catch — fixed in 45676569.

The v1 auth model assumes rater == asker (we explicitly designed feedback for the not-yet-shared-chats case), so in practice the user_id on the row already matches the user posting the thumbs. But you're right that a leaked request_id is a real attack vector and the extra `.eq("user_id", user_id)` is essentially free.

The one nuance: I gated it on `if user_id:` rather than always applying. v1's `user_id` always comes from the relay query param so it should be present, but skipping the filter when None keeps the door open for future system / scheduled flows that might want to record feedback (e.g., automated quality scoring) without breaking them. Inline comment in the diff explains the rationale.

alonelish:
Walked through all three CodeRabbit reviews — addressed in commit 45676569. Summary:

## Fixed (5)

| Comment | File | Resolution |
|---|---|---|
| Move `import litellm` to module scope | `holmes/checks/checks_api.py` | ✅ Moved alongside `from litellm.exceptions import AuthenticationError` |
| Scope `record_feedback` UPDATE by `user_id` | `holmes/core/supabase_dal.py` | ✅ Added `.eq("user_id", user_id)` when `user_id` is provided. Gated on `if user_id:` so future system/scheduled-feedback flows aren't broken when `user_id` is None |
| `body_user_id` misleading name + unnecessary try/except | `server.py` `/api/feedback` | ✅ Simplified to `user_id: Optional[str] = http_request.query_params.get("user_id")` — `.get()` doesn't raise so try/except is unneeded |
| Silent `except` hides errors in agui user_id extraction | `experimental/ag-ui/server-agui.py` | ✅ Narrowed to `(AttributeError, TypeError)` and logs at debug. Also added `isinstance(ctx, dict)` guard since `RunAgentInput.context` is opaque per AG-UI |
| `data.get("num_llm_calls", x) or x` overwrites legitimate 0 | `holmes/core/usage_recorder.py` | ✅ Replaced with explicit `is not None` check |

## Not addressed (intentional)

**`__all__` not alphabetically sorted (RUF022)** in `holmes/core/usage_recorder.py`. CodeRabbit itself flagged this as "💤 Low value" / "purely cosmetic." The current order is intentional:

1. `UsageRecorderState` — the public class everyone imports first
2. `stream_with_usage_recording` — most common helper (streaming chat path)
3. `record_from_llm_result` — non-streaming alternative
4. `record_error` — error-path helper

That groups them by usage frequency / typical reading order, which I find easier to scan than alphabetical. Happy to flip if maintainers prefer the lint-clean version — I'll add a `# noqa: RUF022` if we keep the current order.

All 79 tests still pass. Plan file lives at `~/.claude/plans/does-holmes-save-any-sunny-globe.md` (outside the repo) for reference.

holmes/core/usage_recorder.py
```diff
@@ -0,0 +1,242 @@
...
+                state.status = "error"
+                saw_terminal = True
+            yield msg
+    except Exception:
+        if not saw_terminal:
+            state.status = "error"
+        raise
+    finally:
+        _fire(state)
+
```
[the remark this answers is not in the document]
alonelish: Fixed in 0f8bd384 — `stream_with_usage_recording` now downgrades `state.status` from the default `success` to `aborted` in the `finally` block when no terminal event was seen. Updated the existing test `test_stream_without_terminal_event_still_records_as_aborted` to assert the downgrade, plus added `test_terminal_event_keeps_its_explicit_status` so the downgrade doesn't accidentally fire when a real terminal arrived.

holmes/core/supabase_dal.py
```diff
@@ -817,6 +819,125 @@ def sync_toolsets(self, toolsets: list[dict], cluster_name: str) -> None:
...
+        """Update the feedback_* columns on a HolmesUsageEvents row.
+
+        Best-effort. The UPDATE is account-scoped via WHERE so RLS and an
+        explicit predicate both prevent cross-account writes if request_id
+        ever collides.
+
+        ``user_id`` is accepted for symmetry / future use (e.g. a
+        feedback_user_id column when chats become shareable) but is not
+        currently written, since v1's auth model assumes the rater is the
+        asker (events.user_id == feedback giver).
```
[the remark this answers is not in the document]
alonelish: Fixed in 0f8bd384 — docstring updated to describe the actual scoping behavior (optional `.eq("user_id", user_id)` predicate) and remove the stale 'not currently written' wording.

server.py
```diff
@@ -580,6 +671,37 @@ def chat(chat_request: ChatRequest, http_request: Request):
...
+    # params (e.g. when posted by the Robusta relay). `.get()` returns None
+    # for missing keys without raising, so no try/except needed.
+    user_id: Optional[str] = http_request.query_params.get("user_id")
+    dal.record_feedback(
+        request_id=req.request_id,
+        sentiment=req.sentiment,
+        category=req.category,
+        comment=req.comment,
+        user_id=user_id,
+    )
```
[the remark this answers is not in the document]
alonelish: Fixed in 0f8bd384 — `/api/feedback` now returns `400 missing user_id` when the query param is absent. The DAL keeps the `if user_id:` guard so future system/scheduled-feedback flows can still call `record_feedback` without a user_id, but the FE-facing endpoint must always provide it.

holmes/core/usage_recorder.py
```diff
@@ -0,0 +1,242 @@
...
+    """Background-thread the dal write so the response path never blocks."""
+    if state.dal is None or not getattr(state.dal, "enabled", False):
+        return
+    try:
+        threading.Thread(
+            target=state.dal.record_usage_event,
+            kwargs=state.to_kwargs(),
+            daemon=True,
+            name="usage-recorder",
+        ).start()
```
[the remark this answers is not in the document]
alonelish: Acknowledged but not changing for v1 — daemon thread per event is fine at Holmes' expected volume:

- Peak chat rate: ~100/min sustained = ~1.7/sec
- Each thread: ~8KB stack, runs one Supabase INSERT (~50–200ms typical)
- Steady state: 0–1 threads in flight
- Worst case (60s Supabase timeout, peak load): ~100 threads × 8KB ≈ 800KB total, gone after the timeout fires

A bounded `ThreadPoolExecutor` becomes worth the complexity when sustained traffic exceeds ~1000/min OR Supabase outages are routine. Neither holds for v1.

It's also documented in the design plan (`~/.claude/plans/does-holmes-save-any-sunny-globe.md`) under "What's deliberately out of scope" as a known nice-to-have. Easy to switch later — `_fire` is a single function and the executor would be a drop-in. Re-opening this when one of the volume thresholds is approached.

server.py
```diff
@@ -596,6 +629,38 @@ def chat(chat_request: ChatRequest, http_request: Request):
...
+    """Body for POST /api/feedback. Defined inline to avoid bloating models.py
+    with an endpoint-specific schema."""
+    request_id: str
+    sentiment: str  # 'thumbs_up' | 'thumbs_down'
+    category: Optional[str] = None
+    comment: Optional[str] = None
+
+
+@app.post("/api/feedback")
+def feedback(req: FeedbackRequest, http_request: Request) -> dict:
```
moshemorad: There is a good reason to keep the feedback update as part of holmes? 
If no let's move it out to rpc on the db.

holmes/core/usage_recorder.py
```diff
@@ -0,0 +1,375 @@
...
+# Extracted into a shared regex so both the direct /api/chat path (server.py)
+# and the worker path (conversations_worker/worker.py) can run the same
+# detection. Heuristic — fragile if the runner format changes.
+_SLACK_ASK_PREFIX_RE = re.compile(
+    r"^\*\*@user_(?P<slack_user_id>U[A-Z0-9]+)\*\*\s*•\s*"
+    r"(?P<slack_triggered_at>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z)",
+)
+
+
+def detect_slack_origin(ask: Optional[str]) -> Optional[Dict[str, Any]]:
```
moshemorad, line 59: What about teams? this will also work for teams? we know how to diffrentiate between them?
Is it part of the task scope?

holmes/core/usage_recorder.py
```diff
@@ -0,0 +1,375 @@
...
+    """All the data needed to write one HolmesUsageEvents row.
+
+    Identity / classification fields are set by the entry point at
+    construction time. Mutable runtime fields (`stats`, `iterations`,
+    `tool_call_count`, `finish_reason`, `status`) are filled by the
+    stream wrapper or by `record_from_llm_result` before firing.
+    """
+
+    # required identity / classification — set by the entry point
+    dal: Any  # SupabaseDal; typed Any to avoid circular import at runtime
```
moshemorad: Please add description for each field so it will be clear what each one is mean. There are a lot of field here and it isn't always easy to understand what to set where.

holmes/core/usage_recorder.py
```diff
@@ -0,0 +1,375 @@
...
+    is_internal: bool = False
+    meta: Dict[str, Any] = field(default_factory=dict)
+
+    # mutable — filled during the call by the wrapper or recorder
+    t_start: float = field(default_factory=time.monotonic)
+    stats: Optional[RequestStats] = None
+    iterations: int = 0
+    tool_call_count: int = 0
+    finish_reason: Optional[str] = None
+    status: str = "success"
```
moshemorad: Use enum

holmes/core/supabase_dal.py
```diff
@@ -69,6 +69,8 @@
 CONVERSATIONS_TABLE = "Conversations"
 CONVERSATION_EVENTS_TABLE = "ConversationEvents"
 OAUTH_TOKENS_TABLE = "OAuthTokens"
+HOLMES_USAGE_EVENTS_TABLE = "HolmesUsageEvents"
+HOLMES_USAGE_EVENT_CALLS_TABLE = "HolmesUsageEventCalls"
```
moshemorad: Unused.

holmes/core/supabase_dal.py
```diff
@@ -820,6 +822,136 @@ def sync_toolsets(self, toolsets: list[dict], cluster_name: str) -> None:
                 f"An error occurred during toolset synchronization: {e}", exc_info=True
             )
 
+    def record_usage_event(
```
moshemorad: Consider, instead of all those params just get the UsageRecorderState object and send relevant data to the db.
It will also make the to_kwarg function redundant.

moshemorad (changes requested):
Great work, left couple of comments let me know if something isn't clear.

server.py
alonelish: Makes sense, will move to RPC

alonelish: Moved. From runner as well

holmes/core/usage_recorder.py
alonelish: You are correct, this is not complete. I had to make new changes to the relay repo to support both teams and slack in the right way - this is just a fallback "hack"

holmes/core/usage_recorder.py
alonelish: Added comments

holmes/core/supabase_dal.py
alonelish: Removed HOLMES_USAGE_EVENT_CALLS_TABLE

holmes/core/supabase_dal.py
alonelish: Removed to_kwarg

Avi-Robusta (dismissed):
Looked at moshes comments and your changes since then
LGTM
I saw some git actions here are failing, 
Check if they are the ones that normally fail for you or if this code introduced new issues there

holmes/core/usage_recorder.py
```diff
@@ -225,48 +433,64 @@ def stream_with_usage_recording(
...
-        # `state.status` would still be the constructor default "success".
-        # That's wrong — mark such cases as "aborted" so dashboards can
+        # `state.status` would still be the constructor default SUCCESS.
+        # That's wrong — mark such cases as ABORTED so dashboards can
         # filter incomplete runs out of "successful chat" metrics.
-        if not saw_terminal and state.status == "success":
-            state.status = "aborted"
-        _fire(state)
+        if not saw_terminal and state.status == RequestStatus.SUCCESS:
+            state.status = RequestStatus.ABORTED
```
moshemorad, line 511: I think that im asked about it also in the original pr, why we mark it as aborted? aborted mean that something was stopped intentionally this is the case here?

holmes/core/usage_recorder.py
```diff
@@ -148,67 +149,274 @@ def build_chat_recorder_state(
...
+        self.finish_reason = (
+            metadata.get("finish_reason") or self.finish_reason
+        )
+
+    def _fire(self) -> None:
+        """Background-thread the dal write so the response path never blocks."""
+        if self.dal is None or not getattr(self.dal, "enabled", False):
+            return
+        try:
+            threading.Thread(
```
moshemorad: Please change it to ThreadPool i still don't see a good reason to start a new thread every time.

moshemorad (changes requested):

holmes/core/usage_recorder.py
alonelish: The realistic cause of this branch is client disconnect (browser tab closed, request canceled) -  the user intentionally stopped consuming the stream, even if Holmes didn't intentionally stop producing it. Picked "aborted" rather than "error" because tagging client-disconnects as errors would inflate Holmes' error-rate dashboards with user-driven hangups, which we want to keep separable.

alonelish: BTW, i fixed the error case to collect token usage if we have an error after a few LLM iterations. Previously that would result in 0 tokens - in this PR version we fixed it.

holmes/core/usage_recorder.py
alonelish: Changed to ThreadPool and tested

holmes/core/usage_recorder.py
moshemorad: Are you sure that this happens due to client closing tab? I think i saw it happens also in other cases can you double check and report back? e.g. start chat close the tab and see if it is indeed reporting aborted.

alonelish: It does not happen on browser exit or tab close, at least not usually. That said, it is currently used for internal stats and we can leave it to see if there are any cases where it does happen. I know it happens when runner restarts during llm session.

moshemorad (approved):

PR #1974, opened 2026-04-30
Flush events eagerly on TOKEN_COUNT to prevent memory buildup

naomi-robusta:
## Summary
Modified the event publisher to flush pending events immediately when a `TOOL_RESULT` event is encountered, preventing events from sitting in memory during subsequent LLM calls.

## Key Changes
- Added `StreamEvents.TOOL_RESULT` to the `_FLUSH_IMMEDIATELY_EVENTS` set in `event_publisher.py`
- Updated the comment explaining the rationale: TOOL_RESULT events precede another (typically slow) LLM call, so immediate flushing ensures subscribers see events promptly rather than waiting for the LLM call to complete
- Added comprehensive test `test_publisher_flushes_eagerly_on_tool_result()` that verifies:
  - Events are split into two flush calls when TOOL_RESULT is present
  - First flush contains AI_MESSAGE + START_TOOL + TOOL_RESULT (triggered by TOOL_RESULT)
  - Second flush contains TOKEN_COUNT + ANSWER_END (triggered by ANSWER_END)
  - Compact flags are set correctly for each flush

## Implementation Details
The change treats TOOL_RESULT similarly to terminal events (ANSWER_END, etc.) in terms of triggering immediate flushes, but with different semantics: rather than ending a turn, TOOL_RESULT marks a transition point where the next operation (another LLM call) may take significant time. This prevents a buildup of pending events in memory during that latency window.

https://claude.ai/code/session_01LjFK8k38XDifb4Q3LbuXcu

## Summary by CodeRabbit

* **Bug Fixes**
  * Token-count markers now trigger immediate flushes for more timely event delivery; tool-result events no longer force intermediate flushes.

* **Tests**
  * Updated tests to verify eager flushing on token-count markers and that intermediate flushes are non-compacting while terminal flushes compact results.


yuryrudey (approved):

holmes/core/conversations_worker/event_publisher.py
```diff
@@ -23,9 +23,13 @@
...
 # "history snapshot + prior events superseded" semantics so it's flushed +
-# compacted with the same logic.
+# compacted with the same logic.  TOOL_RESULT events are flushed eagerly
+# because the next step is another LLM call which typically takes >1s — if
+# we don't flush now, pending events sit in memory for the duration of the
+# LLM call and subscribers see nothing.
 _FLUSH_IMMEDIATELY_EVENTS = _TERMINAL_EVENTS | {
-    StreamEvents.CONVERSATION_HISTORY_COMPACTED
+    StreamEvents.CONVERSATION_HISTORY_COMPACTED,
+    StreamEvents.TOOL_RESULT,
```
moshemorad: Most of our events are probably tools calls, i like that before we were batching those and inserting into the db. 
Is there a better solution for this also with keeping the batching? maybe some timer that will flush every second or something like that? 

I don't think it is a most but we are kinda abusing our db every where (in runner we are doing the same, no batching) 

I don't want to push back the release so for me it is still a green light but let's remember in the future to try and improve that.

Thanks.

moshemorad (dismissed):

moshemorad (approved):

PR #1989, opened 2026-05-03
Add Supabase Realtime verification before starting conversation worker

naomi-robusta:
## Summary

This PR adds a verification mechanism to the ConversationWorker that checks whether Supabase Realtime is actually enabled before spinning up polling and subscription infrastructure. The worker now starts a background verifier thread that calls an `is_realtime_enabled()` RPC, only proceeding with active workers once Supabase confirms realtime support is available.

## Key Changes

- **New `_realtime_verify_loop()` method**: Background thread that repeatedly calls `is_realtime_enabled()` RPC until getting a definitive answer:
  - On `True`: Updates HolmesStatus and starts active workers (executor, claim loop, realtime manager)
  - On `False`: Logs warning and shuts down the worker without starting active components
  - On connectivity errors (`None`): Retries with exponential backoff (5s initial, up to 120s max)

- **New `is_realtime_enabled()` method in SupabaseDal**: Calls the RPC and handles various response formats and error cases:
  - Returns `True`/`False` for definitive answers
  - Returns `None` for connectivity/transport errors (allows retry)
  - Treats missing RPC (PGRST202) as `False` (definitive)

- **Modified `start()` method**: Now marks worker as running but defers starting executor/claim-thread/realtime-manager until verifier confirms realtime is enabled

- **New `_start_active_workers()` method**: Extracted the executor/claim-loop startup logic into a separate idempotent method called by the verifier

- **Updated `stop()` method**: Now properly handles the verify thread, skipping self-join when called from within the verify thread itself

- **Updated HolmesStatus**: Added `realtime_available` parameter to `update_holmes_status_in_db()` so realtime-related status fields only flip to env-var values after verification succeeds

- **New environment variables**: `CONVERSATION_WORKER_REALTIME_VERIFY_INITIAL_BACKOFF_SECONDS` and `CONVERSATION_WORKER_REALTIME_VERIFY_MAX_BACKOFF_SECONDS` for controlling retry backoff

## Implementation Details

- The verify thread is daemon-threaded and uses a `threading.Event` for graceful shutdown signaling
- Comprehensive test coverage including unit tests for the verify loop behavior and end-to-end integration tests
- Avoids wasted load against projects that don't support realtime by deferring worker startup
- Distinguishes between "server said no" (definitive False) and "couldn't reach server" (None) to prevent premature shutdown on transient network issues

https://claude.ai/code/session_01DfrnBQwebpiDjDrDd87HmT

## Summary by CodeRabbit

* **New Features**
  * Worker startup now verifies realtime availability before activating realtime workflows; activation is deferred or aborted based on definitive checks.
  * Status updates now reflect realtime availability (realtime-related fields remain disabled until availability is confirmed).

* **Chores**
  * Added two configurable backoff settings for realtime verification (defaults added; no runtime clamping/validation included).

* **Tests**
  * Expanded tests for realtime verification paths, start/stop lifecycles, and realtime-status checks.


holmes/core/supabase_dal.py
```diff
@@ -916,6 +916,82 @@ def update_run_status(
...
+            ``None``  — Could not determine (connectivity error, auth failure,
+                       or any other transport-level issue). The caller should
+                       NOT take destructive action in this case.
+
+        We deliberately distinguish "definitive answer from server" from
+        "couldn't reach the server" so the conversation worker only disables
+        itself when Supabase has actually told us realtime is off.
+        """
+        if not self.enabled:
+            return None
```
moshemorad, line 937: Shouldn't this be False? if dal isn't enable?

moshemorad (approved):

holmes/core/supabase_dal.py
naomi-robusta: not really. this shouldn't happen, it's just defensive code, we shouldn't reach it if it's actually none, so I made it to match the other connectivity errors.

PR #2023, opened 2026-05-11
Propagate inbound trace context

IceHamster:
## Summary

Continue inbound distributed trace context for `/api/chat` so Holmes investigation spans join the caller's trace when a standard `traceparent` header is present.

## What changed

- added an HTTP middleware in `server.py` that:
  - extracts OpenTelemetry/W3C trace context from inbound request headers
  - attaches that context for the lifetime of the request
  - detaches the context in `finally` to avoid leaking it into later requests
- added `tests/test_server_trace_context.py` to verify that:
  - `holmesgpt.investigation` reuses the upstream trace ID and parent span when `traceparent` is provided
  - trace context is cleaned up after the request, so a later request without trace headers starts a fresh trace

## Why

Without this, Holmes always starts a new root trace for `/api/chat`, which breaks trace continuity between upstream callers and the Holmes investigation spans they trigger.

With this change, Holmes traces can be correlated with the originating request in OTel backends while keeping request-local context isolated.

## Testing

```bash
.venv/bin/python -m pytest tests/test_server_trace_context.py tests/test_server_endpoints.py --no-cov -n 0 -q
```

## Summary by CodeRabbit

* **New Features**
  * Distributed trace context propagation now supported for HTTP requests, enabling seamless tracing across services.

* **Tests**
  * Added tests validating trace context propagation functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HolmesGPT/holmesgpt/pull/2023)

### DIFF
--- a/server.py
+++ b/server.py
@@ -22,6 +22,7 @@ import litellm
 from holmes.core.oauth_config import OAuthConfigLookupError, OAuthTokenExchangeError
 from holmes.core.oauth_server_callbacks import get_toolset_oauth_config, process_oauth_callback
 from holmes.core.oauth_utils import _get_token_manager
+from opentelemetry import context as otel_context, propagate
 import sentry_sdk
 import uvicorn
 from fastapi import FastAPI, HTTPException, Request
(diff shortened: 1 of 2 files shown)


IceHamster:
Closed due to CLA

PR #2042, opened 2026-05-14
Improve 'close-matches' naming behaviour

aantn:
## Summary
Updated the Elasticsearch disaster recovery test case to include explicit instructions for service field validation, ensuring the LLM only considers logs from the `companyopswebjob` service.

## Changes
- Added clarifying instructions to the user prompt in the test case that:
  - Require exact matching of the `service` field to `companyopswebjob` (case-insensitive)
  - Prohibit substitution or inference from logs of other services with similar names
  - Mandate explicit acknowledgment if no logs for `companyopswebjob` are found
  - Prevent proposing root causes from unrelated services
  - Recommend obtaining logs directly from `companyopswebjob` if unavailable

## Details
This change improves the test case by making the service filtering requirement explicit in the prompt, which helps validate that the LLM correctly handles cases where it must distinguish between services and avoid making assumptions based on partial name matches or logs from different services.

https://claude.ai/code/session_015scgyQbqJD23heqct31Ger

## Summary by CodeRabbit

* **New Features**
  * Assistant now explicitly states when no data exists for the exact requested entity, may surface findings from similarly named/related entities only when clearly labeled as different, and prompts users to confirm or provide the intended entity. It also flags cluster/source mismatches when relevant.

* **Tests**
  * Updated and added test fixtures to enforce these transparency rules, including a new wrong-cluster scenario and added a regression tag to relevant tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HolmesGPT/holmesgpt/pull/2042)


aantn:
@coderabbitai you were right that the prescriptive instructions in the test prompt were biasing toward instruction-following — that change has been reverted in f90dcd2.

The fix is now on the Holmes side instead: `holmes/plugins/prompts/generic_ask.jinja2` now instructs the model to be both transparent AND useful when it can't find the exact entity but finds similarly-named ones (state no data for the exact name, report what was found in the related entities, clearly label them as different services).

The test prompt is back to its original form. Expected_output now checks for both behaviors (transparency about the missing name + useful reporting on adjacent services, clearly labeled). 254 passes 9/9 across Sonnet 4.5, Opus 4.6, and Opus 4.7 with that change. Eval 19 was also updated from "either behavior" to "both behaviors" for consistency.

---

aantn:
@coderabbitai re: the suggestion to add `"Must recommend obtaining logs directly from companyopswebjob..."` — that bullet was deliberately removed from `expected_output` earlier in this PR. The "ideal behavior" intentionally requires only (1) transparency about no exact-match data and (2) useful adjacent reporting clearly labeled — not a recommendation to fetch the named service's logs.

The header comment did still contain that phrasing, which is what created the inconsistency you spotted. I've removed it in b2282ab so the comment now matches the criteria.

---

aantn:
@coderabbitai re: moving `set -e` before `source ../../shared/es_test_utils.sh` / `es_setup` — skipping this in 259 because it would diverge from the convention used across all 21 existing ES-based test_case.yaml fixtures in `tests/llm/fixtures/test_ask_holmes/` (e.g. 183a–g, 184–193, 235, 245, 254). They all place `set -e` *after* the source + es_setup calls. Worth doing as a separate cleanup across all 21 files rather than in this PR.

---

PR #2058, opened 2026-05-17
fix(operator): forward HOLMES_API_KEY in HealthCheck requests (#2030)

### LINKED ISSUE #2030: Using API key for holmes api when deploying via helm breaks health checks due to authenticaiton error
### What happened?

  After enabling API key authentication on the Holmes API (HOLMES_API_KEY env var on the holmes server) via the Helm chart, the Holmes operator can no longer execute health checks. Every HealthCheck
  resource fails with a 401 Unauthorized from the operator's call to /api/checks/execute, because the operator's HTTP client does not send the API key.

  The operator's HolmesAPIClient (holmes_operator/client/holmes_api_client.py) constructs an httpx.AsyncClient with no auth headers, and holmes_operator/config.py does not read any HOLMES_API_KEY env
  var, so there is no way to configure the operator to authenticate against an API-key-protected Holmes server.

  HealthCheck status from the cluster:

  Conditions:
      Last Transition Time:  2026-05-12T23:35:40.619007+00:00
      Message:               Operator failed to execute check: Client error '401 Unauthorized' for url 'http://holmesgpt-holmes/api/checks/execute'
      Reason:                OperatorError
      Status:                True
      Type:                  Failed
    Error:     Client error '401 Unauthorized' for url 'http://holmesgpt-holmes/api/checks/execute'
    Phase:     Running
    Result:    error

  Events:
    Type     Reason         Age   From  Message
    ----     ------         ----  ----  -------
    Normal   Logging        68s   kopf  Executing check holmesgpt/incident-health-check via Holmes API
    Normal   Logging        68s   kopf  Creating HealthCheck: holmesgpt/incident-health-check
    Warning  OperatorError  64s   kopf  Failed to execute health check: Client error '401 Unauthorized' for url 'http://holmesgpt-holmes/api/checks/execute'
    Error    Logging        64s   kopf  Handler 'on_healthcheck_create' failed with an exception. Will retry.

    File "/app/holmes_operator/client/holmes_api_client.py", line 97, in execute_check
      response.raise_for_status()
    File "/usr/local/lib/python3.11/site-packages/httpx/_models.py", line 829, in raise_for_status
      raise HTTPStatusError(message, request=request, response=self)
  httpx.HTTPStatusError: Client error '401 Unauthorized' for url 'http://holmesgpt-holmes/api/checks/execute'

 When HOLMES_API_KEY is configured on the Holmes API server via the Helm chart, the operator deployed by the same chart should be able to authenticate to it out of the box — for example, by exposing a
  value such as operator.holmesApiKey (or a secretRef) that wires the key into the operator pod, and having HolmesAPIClient automatically send it as X-API-Key (or Authorization: Bearer …) on every
  request to /api/checks/execute. HealthChecks should then succeed exactly as they do when API key auth is disabled.

  1. Install the HolmesGPT Helm chart with operator.enabled=true and set HOLMES_API_KEY on the holmes API pod (via additionalEnvVars referencing a Secret).
  2. Confirm the server enforces auth — the holmes pod logs "API key authentication enabled (HOLMES_API_KEY is set)".
  3. Apply any HealthCheck CR in the cluster.
  4. kubectl describe healthcheck <name> shows OperatorError: 401 Unauthorized for url 'http://<release>-holmes/api/checks/execute'.
  5. Unset HOLMES_API_KEY and the same HealthCheck succeeds, confirming the operator never sends the key.

_No response_

net0pyr:
Fixes #2030.

## Problem

When `HOLMES_API_KEY` is set on the Holmes API server via the Helm chart, the Holmes operator cannot authenticate against the API server and every `HealthCheck` CR fails with `401 Unauthorized` from `/api/checks/execute`. The operator never had a code path that forwarded an API key, so users who enable auth on the server effectively break the operator.

## Server-side auth contract (confirmed before coding)

- Header: `X-API-Key` (checked first), with `Authorization: Bearer <key>` as a fallback (`holmes/utils/auth.py`).
- Scope: global FastAPI middleware in `server.py:278`, applied to every path except `/healthz` and `/readyz`. `/api/checks/execute` is gated.
- Comparison: plain `==`, not `hmac.compare_digest`. Out of scope for this PR — see "Follow-ups" below.

The operator now sends the same `X-API-Key` header the server already accepts. No new convention introduced; no server-side changes.

## What changed

**Operator config (`holmes_operator/config.py`)**
- New `HOLMES_API_KEY` env var, exposed as `OperatorConfig.holmes_api_key` (`Optional[str]`, default `None`). Default `None` preserves prior behavior byte-for-byte when the env var is unset.

**Operator HTTP client (`holmes_operator/client/holmes_api_client.py`)**
- `HolmesAPIClient.__init__` accepts an optional `api_key` argument. When non-empty, `X-API-Key` is attached to the underlying `httpx.AsyncClient` **once at construction**, so it applies to every current and future endpoint automatically (no per-call risk of forgetting on a new endpoint).
- Only `bool(self._api_key)` is logged — the key value never appears in logs, error messages, or `__repr__`.

**Operator context (`holmes_operator/context.py`)**
- `initialize()` passes `config.holmes_api_key` to the client.

**Helm chart (`helm/holmes/values.yaml`, `helm/holmes/templates/operator-deployment.yaml`)**
- Two mutually-exclusive operator-side options:
  - `operator.holmesApiKey` — plain value (convenient for local testing).
  - `operator.holmesApiKeyExistingSecret.{name,key}` — `secretKeyRef`, recommended for production. Takes precedence when both are set.
- If neither is set, no `HOLMES_API_KEY` env var is rendered into the operator Deployment — current behavior, no regression for users who don't run with auth.
- `values.yaml` notes that the natural pattern is to point both the server and the operator at the same Secret/key. Not enforced.

**Tests (`tests/holmes_operator/test_holmes_api_client_auth.py`)**
- `HolmesAPIClient(api_key=None)` → outgoing request has no auth header.
- `HolmesAPIClient(api_key="secret")` → outgoing request to `/api/checks/execute` carries `X-API-Key: secret`.
- `HolmesAPIClient(api_key="secret")` → header is present on every endpoint, asserted at the `httpx.AsyncClient` level so future endpoints inherit it automatically.
- Constructing the client never logs the key value.
- Uses `respx` and pytest-asyncio to match the existing operator test idiom — no new dependencies.

## Summary by CodeRabbit

* **New Features**
  * Added API key configuration for Holmes API integration in operator deployments.
  * Helm values support either a direct API key or a reference to an existing Kubernetes Secret (secret reference takes precedence).
  * Operator deployment now exposes the API key to the runtime container when configured.

* **Tests**
  * Added tests validating authentication header behavior, request handling, and that API keys are not leaked in logs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HolmesGPT/holmesgpt/pull/2058?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/helm/holmes/templates/operator-deployment.yaml
+++ b/helm/holmes/templates/operator-deployment.yaml
@@ -59,6 +59,16 @@ spec:
           value: {{ .Values.operator.cleanupCompletedChecks | default false | quote }}
         - name: COMPLETED_CHECK_TTL_HOURS
           value: {{ .Values.operator.completedCheckTTLHours | default 24 | quote }}
+        {{- if and .Values.operator.holmesApiKeyExistingSecret .Values.operator.holmesApiKeyExistingSecret.name .Values.operator.holmesApiKeyExistingSecret.key }}
+        - name: HOLMES_API_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.operator.holmesApiKeyExistingSecret.name | quote }}
+              key: {{ .Values.operator.holmesApiKeyExistingSecret.key | quote }}
+        {{- else if .Values.operator.holmesApiKey }}
+        - name: HOLMES_API_KEY
+          value: {{ .Values.operator.holmesApiKey | quote }}
+        {{- end }}
         {{- if .Values.operator.additionalEnvVars }}
         {{- toYaml .Values.operator.additionalEnvVars | nindent 8 }}
         {{- end }}
--- a/helm/holmes/values.yaml
+++ b/helm/holmes/values.yaml
@@ -879,6 +879,18 @@ operator:
   holmesApiUrl: ""  # Defaults to "http://<release-name>-holmes:80"
   holmesApiTimeout: 300  # seconds
 
+  # Authenticate the operator's calls to the Holmes API server.
+  # Required when the API server has HOLMES_API_KEY set (otherwise HealthChecks
+  # fail with 401 Unauthorized from /api/checks/execute).
+  # Natural pattern: point both the server and the operator at the same Secret/key.
+  # Prefer holmesApiKeyExistingSecret for production deployments;
+  # holmesApiKey (plain value) is convenient for local testing only.
+  # If both are set, holmesApiKeyExistingSecret takes precedence.
+  holmesApiKey: ""
+  holmesApiKeyExistingSecret:
+    name: ""   # Name of an existing Secret containing the API key
+    key: ""    # Key within the Secret that holds the API key value
+
   # Logging
   logLevel: INFO
 
--- a/holmes_operator/client/holmes_api_client.py
+++ b/holmes_operator/client/holmes_api_client.py
@@ -22,21 +22,35 @@ class HolmesAPIClient:
     - Error handling and logging
     """
 
-    def __init__(self, base_url: str, timeout: int = 300):
+    def __init__(
+        self,
+        base_url: str,
+        timeout: int = 300,
+        api_key: Optional[str] = None,
+    ):
         """
         Initialize the Holmes API client.
 
         Args:
             base_url: Base URL of the Holmes API server (e.g., "http://holmes-api:80")
             timeout: Default timeout for requests in seconds
+            api_key: Optional API key. When set, sent as the X-API-Key header on
+                every request to authenticate against a Holmes API server that has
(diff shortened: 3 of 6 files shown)


gueux:
Any updates? Seems like fix is ready and just need to be merged in main.

PR #2103, opened 2026-05-29
fix(datadog): convert v1/events time params to UNIX timestamps

mubarak-j:
## Problem

When HolmesGPT investigates a Kubernetes issue and calls the Datadog v1 Events API (`/api/v1/events`), it consistently fails with a **400 Bad Request**:

```
HTTP error: 400 - {"errors":["The value provided for parameter 'end' is invalid"]}
```

Full trace observed in production:

```
2026-05-29 21:05:52.344 INFO     DatadogAPIGet Tool Invocation:
2026-05-29 21:05:52.344 INFO       Endpoint: /api/v1/events
2026-05-29 21:05:52.344 INFO       Query Params: {
  "start": -86400,
  "end": "now",
  "tags": "integration:kubernetes",
  "priority": "normal"
}
2026-05-29 21:05:52.353 INFO     Converted relative time fields: end: 'now' -> '2026-05-29T21:05:52.352289Z'
2026-05-29 21:05:52.872 ERROR    HTTP error: 400 - {"errors":["The value provided for parameter 'end' is invalid"]}
```

## Root Cause

There are two bugs in `preprocess_time_fields` in `datadog_api.py`:

### Bug 1 — Wrong format for v1 endpoints

`convert_relative_time` always returns RFC3339 strings (e.g. `"2026-05-29T21:05:52Z"`). This is correct for Datadog v2 API endpoints, but the **v1 Events API requires UNIX integer timestamps**. Sending an ISO string for `end` is what triggers the 400.

### Bug 2 — Integer relative offsets are silently ignored

The LLM (following the prompt in `datadog_general_instructions.jinja2`) sometimes produces `start` as a negative integer (e.g. `-86400` meaning "86400 seconds ago"). `preprocess_time_fields` only processes values where `isinstance(value, str)`, so an integer `-86400` passes through untouched. The Datadog v1 Events API then receives a nonsensical negative number instead of a valid UNIX timestamp, which would also cause a 400.

## Fix

Two additions to `datadog_api.py`:

1. **`_V1_UNIX_TIMESTAMP_ENDPOINTS`** — a set of known v1 endpoints that require UNIX integer timestamps (`api/v1/events`, `api/v1/metrics`).

2. **`_resolve_to_unix_timestamp(value)`** — converts any relative time representation to an absolute UNIX timestamp:
   - Negative integer → `now + value` (e.g. `-86400` → `now - 86400`)
   - String `"now"` → `int(time.time())`
   - Unit strings like `"-24h"`, `"-7d"` → resolved via existing `convert_relative_time`, then converted to UNIX int
   - Already-absolute UNIX timestamp → returns `None` (no conversion needed)

3. **`preprocess_time_fields`** — when the target endpoint is in `_V1_UNIX_TIMESTAMP_ENDPOINTS`, it applies `_resolve_to_unix_timestamp` to time fields instead of the RFC3339 path, and now handles both `str` and `int`/`float` values (not just strings).

Non-v1 endpoints are unaffected — they continue to receive RFC3339 strings as before.

## Testing

Smoke-tested with inline assertions covering:
- `_resolve_to_unix_timestamp(-86400)` → absolute UNIX timestamp ~24h ago
- `_resolve_to_unix_timestamp("now")` → current UNIX timestamp
- `_resolve_to_unix_timestamp(1748556352)` → `None` (already absolute, no change)
- `preprocess_time_fields({"start": -86400, "end": "now"}, "api/v1/events")` → both fields become UNIX ints
- `preprocess_time_fields({"filter": {"from": "-24h", "to": "now"}}, "api/v2/logs/events/search")` → v2 endpoint unchanged, still receives ISO strings

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Datadog v1 compatibility by resolving varied time inputs ("now", negative offsets, relative durations, RFC3339-like strings, numeric offsets) into UNIX integer timestamps for v1 endpoints (events, metrics, query). Falls back to prior relative-time handling when resolution fails; conversions are logged.

* **Tests**
  * Added tests verifying time resolution behavior, v1/v2 distinctions, skipping None/missing fields, and that original payloads are not mutated.


moshemorad:
@claude review

mubarak-j:
@claude review

PR #2108, opened 2026-05-31
feat: opt-in client-side tool search (load_tools) to cut tool-schema tokens

yakir-shriker:
## Problem

With several MCP servers enabled, every tool's full JSON schema is injected into the system prompt on every turn — before the user sends anything. In a multi-server setup this is ~16–50k tokens of tool definitions on a trivial `hi`, inflating cost, latency, and the KV cache (the "MCP tax").

## This PR: opt-in client-side tool search (`load_tools`)

Behind `HOLMES_TOOL_SEARCH_ENABLED` (default **off**):

- Heavy tool schemas (MCP toolsets) are **held back** from the default tool list; the model gets a single `load_tools(query)` **function tool**.
- The model calls `load_tools` to discover what it needs; HolmesGPT runs the search itself (`ToolExecutor.search_deferred_tools` — regex/substring over tool name/description/toolset) and marks matches loaded.
- The existing tool-refetch loop in `call_stream` exposes the loaded tools on the next step; the model calls them normally.

`load_tools` is an ordinary function tool that HolmesGPT executes, so it works identically across the Anthropic API, Bedrock, OpenAI, and any LiteLLM-fronted gateway — no beta headers, no server tools.

## Changes
- `holmes/core/tool_search.py` — `load_tools` tool definition.
- `holmes/core/tools_utils/tool_executor.py` — `get_visible_tools_openai_format(loaded_names)` (core tools + `load_tools` + already-loaded MCP tools) and `search_deferred_tools(query)`.
- `holmes/core/tool_calling_llm.py` — per-conversation `_loaded_tool_names`; `_get_tools` serves the visible set when enabled; `load_tools` is intercepted in `_invoke_llm_tool_call` and never reaches the `ToolExecutor`.
- `holmes/common/env_vars.py` — `HOLMES_TOOL_SEARCH_ENABLED` flag.
- Tests for the visible listing, search, `_get_tools` routing, and the interception.

Only MCP-type toolsets are deferred (built-ins like kubernetes/bash stay loaded so common ops don't pay a search hop); the policy is a single `DEFERRABLE_TOOLSET_TYPES` set, easy to extend.

## Verification
- Unit tests added; existing suite green; `ruff` clean.
- End-to-end against a LiteLLM gateway (Claude Opus): a `hi` turn and a real AWS investigation both run cleanly — `load_tools` → tool loads → real tool call → answer, with no broken `tool_use`/`tool_result`, and ~24–89% fewer prompt tokens depending on how much of the catalog is in play.

Flag is off by default, so this is a no-op unless explicitly enabled. Happy to adjust naming, the deferral policy, or add docs.

Signed-off-by: Yakir Shriker <yakirshr@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Added a configuration toggle to enable client-side progressive tool search, exposing heavy toolsets only when requested
  * Introduced a user-invokable "load tools" action to discover and load tools on demand, reducing initial tool-schema context

* **Tests**
  * Added tests for progressive disclosure, search behavior, load flow, and state-reset handling

### DIFF
--- a/holmes/common/env_vars.py
+++ b/holmes/common/env_vars.py
@@ -62,6 +62,10 @@ SENTRY_TRACES_SAMPLE_RATE = float(os.environ.get("SENTRY_TRACES_SAMPLE_RATE", "0
 
 EXTRA_HEADERS = os.environ.get("EXTRA_HEADERS", "")
 THINKING = os.environ.get("THINKING", "")
+# Client-side tool search. When enabled, heavy tool schemas (MCP toolsets) are held
+# back from the default tool list and a load_tools meta-tool is exposed so the model
+# loads them on demand instead of putting the whole catalog in context every turn.
+TOOL_SEARCH_ENABLED = load_bool("HOLMES_TOOL_SEARCH_ENABLED", False)
 REASONING_EFFORT = os.environ.get("REASONING_EFFORT", "").strip().lower()
 
 
--- a/holmes/core/tool_calling_llm.py
+++ b/holmes/core/tool_calling_llm.py
@@ -23,9 +23,11 @@ from holmes.common.env_vars import (
     LOG_LLM_USAGE_RESPONSE,
     RESET_REPEATED_TOOL_CALL_CHECK_AFTER_COMPACTION,
     TEMPERATURE,
+    TOOL_SEARCH_ENABLED,
     load_bool,
 )
 from holmes.core.llm import LLM
+from holmes.core.tool_search import LOAD_TOOLS_NAME
 from holmes.core.llm_usage import RequestStats
 from holmes.core.models import (
     FrontendToolResult,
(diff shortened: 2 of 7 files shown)


yakir-shriker:
Thanks @coderabbitai — addressed both:

1. **Reset loaded tools between conversations**: `reset_interaction_state()` now also clears `_loaded_tool_names` (+ a regression test `test_reset_interaction_state_clears_loaded_tools`).
2. **Invalid-regex fallback test**: added `test_search_deferred_tools_falls_back_for_invalid_regex` covering `"["` / `"("`.

Squashed into the single commit; 57 tests pass, ruff clean.

yakir-shriker:
Added both integration tests in the latest push:

- `test_load_tools_intercepted_during_invocation` — drives the full loop: the model emits a `load_tools` tool call, `_invoke_llm_tool_call` intercepts it (never routed to the ToolExecutor), `search_deferred_tools` runs, `_loaded_tool_names` is updated, and the loop continues to a final answer.
- `test_loaded_tool_names_preserved_across_clone` — `with_executor()` carries `_loaded_tool_names` as an independent copy.

Also rebased onto latest master (temperature change from #2109 is already there), so this is a single clean commit. 62 tests pass, ruff clean.

ranyhb:
i need it !!!!

PR #2123, opened 2026-06-04
fix: redact API keys and tokens from logs (#2010)

### LINKED ISSUE #2010: BUG: Gemini API key exposure in error logs / exception messages
### What happened?

When a request to a Gemini model fails (e.g. with a 400 BadRequestError), the raw GEMINI_API_KEY is leaked in plaintext as part of the exception traceback that HolmesGPT writes to stdout/stderr.                                                                                       
                                                                                                                                                                              
The key is appended to the request URL by LiteLLM's Gemini handler:                                                                                                       

https://generativelanguage.googleapis.com/v1alpha/models/<model>:generateContent?key=AXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX

When httpx raises HTTPStatusError, the URL (and therefore the API key) is included in the error message. HolmesGPT then logs the full exception via logger.error(...) from:                                                   
  - holmes/core/tool_calling_llm.py → call_stream / call                                                                                                                      
  - server.py chat handler (Error in /api/chat: ...)
                                                                                      
The leaked key flows into whatever log backend is configured (in our case Loki / Grafana, and from there potentially Slack alerts), making it trivially harvestable by anyone with read access to logs.                                                                                                                                            
                                                                                                                                                                              
Excerpt from a real log line (key redacted):                                                                                                                                

```
httpx.HTTPStatusError: Client error '400 Bad Request' for url                                                                                                               
  'https://generativelanguage.googleapis.com/v1alpha/models/gemini-3.1-flash-lite-preview:generateContent?key=AIza****REDACTED****'
```

The API key should never appear in any log output, exception message, or error response — regardless of whether the underlying HTTP call succeeds or fails.                 

Expected behavior:                                                                                                                                                          
- Sanitize/redact the key=... query parameter (and any Authorization, x-goog-api-key, etc. headers) before logging the exception.
- HolmesGPT's logging layer should treat upstream LiteLLM/httpx exception strings as untrusted and scrub known credential patterns (key=, api_key=, Bearer ..., AWS keys, etc.) before writing them.
- Ideally also report this upstream to LiteLLM so the URL is masked at the source.

1. Configure HolmesGPT with a Gemini model and a real GEMINI_API_KEY:
```                                       
  modelList:
    gemini-flash:                                                                                                                                                             
      api_key: "{{ env.GEMINI_API_KEY }}"                                                                        
      model: gemini/gemini-3.1-flash-lite-preview
  config:                                                                                                                                                                     
    model: "gemini-flash"
```
 
2. Trigger any request to /api/chat that causes the upstream Gemini API to return a non-2xx response.
 The simplest way:                                                     
    - Use an invalid API key, or                                                                                                                                              
    - Send a tool-calling conversation long enough to trigger Gemini implicit caching (which currently returns 400 CachedContent can not be used with GenerateContent request setting system_instruction, tools or tool_config).
3. Inspect the HolmesGPT pod logs:                                                                                                                                          
  kubectl logs -n holmesgpt deploy/holmesgpt | grep 'key='                                                                                                                    4. The full Gemini API key is visible in plaintext inside the httpx.HTTPStatusError traceback.

- Affected environment: HolmesGPT running on EKS (image tag: <fill in>), LiteLLM version: <fill in from pip freeze inside the pod>.                                         
- Severity: High — the key is a long-lived static credential and ends up in any downstream log sink (Loki/Grafana, Slack channel notifications, S3 log archives, etc.).
- Mitigation we applied locally: rotated the leaked GEMINI_API_KEY via Google AI Studio and updated AWS Secrets Manager.
- Related upstream behavior: LiteLLM's vertex_and_google_ai_studio_gemini.py puts the API key in the URL query string rather than a header, which is what makes the         
  exception leak the secret. Worth filing a parallel issue against BerriAI/litellm to either (a) move the key to the x-goog-api-key header, or (b) sanitize the URL in raised 
  exceptions.                                                                                                                                                                 
- Suggested fix locations in HolmesGPT:                                                                                                                                     
- holmes/core/llm.py (completion wrapper) and holmes/core/tool_calling_llm.py — wrap LiteLLM exceptions and scrub key=[^&\s]+ (and similar credential patterns) before    
  re-raising or logging.
- Central logging filter / formatter that redacts known secret patterns, applied to all loggers.

ricky9408:
## What

Fixes #2010 — Gemini (and other provider) API keys were being exposed in plaintext in logs.

## Why

LiteLLM appends the API key to the request URL as a `?key=...` query parameter. When httpx receives a non-2xx response and raises `HTTPStatusError`, the full URL — credential included — becomes part of the exception message. HolmesGPT logs that via `logging.error(..., exc_info=True)` (e.g. the LLM-call error handlers in `holmes/core/tool_calling_llm.py` and the `/api/chat` handler in `server.py`), so the raw key lands in stdout/stderr and flows downstream into log backends (Loki, Grafana, Slack alerts). Static, long-lived credentials then become harvestable by anyone with log access.

## How

- Add `SecretRedactingFilter` and `redact_secrets()` in `holmes/utils/log.py`. The filter scrubs credentials from **every** log record before it reaches a handler — both the message/args **and** the formatted exception traceback (cached on the record so downstream handlers reuse the redacted text).
- Redaction targets:
  - Credential-bearing query params and key/value pairs: `key`, `api_key`, `token`, `access_token`, `password`, `secret`, `authorization`, etc. (URLs, query strings, headers, JSON bodies).
  - `Bearer` / `Basic` authorization values.
  - Well-known provider key formats by value (Google `AIza…`, OpenAI `sk-…`, Anthropic `sk-ant-…`), so a leaked key is masked even without a recognizable key name.
- Install the filter on the root logger handlers in both logging entry points: `server.py` `init_logging()` and the CLI `init_logging()` in `holmes/utils/console/logging.py`. Installation is idempotent.

Non-secret query params and ordinary log text are preserved.

## Tests

Added `tests/utils/test_log_redaction.py` (18 cases) covering query-param, header, JSON-body, known-key-format and full-traceback redaction, idempotent install, and a non-secret pass-through case. All pass.

## Summary by CodeRabbit

* **New Features**
  * Automatic secret redaction applied to all logging output — messages, formatted arguments, and tracebacks — and enabled during logging initialization; installation is idempotent and preserves non-secret content.
* **Tests**
  * Added comprehensive tests covering redaction in URLs, tokens, known key formats, JSON-like bodies, exception tracebacks, interpolation cases, and idempotent installation.

### DIFF
--- a/holmes/utils/console/logging.py
+++ b/holmes/utils/console/logging.py
@@ -6,6 +6,8 @@ from typing import List, Optional
 from rich.console import Console
 from rich.logging import RichHandler
 
+from holmes.utils.log import install_secret_redaction
+
 
 class Verbosity(Enum):
     NORMAL = 0
@@ -97,6 +99,10 @@ def init_logging(verbose_flags: Optional[List[bool]] = None, log_costs: bool = F
             ],
         )
         suppress_noisy_logs()
+
+    # Strip API keys / tokens from all log records (messages and tracebacks)
+    # before they are printed or shipped to a backend.
+    install_secret_redaction()
 
     logging.debug(f"verbosity is {verbosity}")
 
--- a/holmes/utils/log.py
+++ b/holmes/utils/log.py
@@ -1,7 +1,106 @@
 """Logging utilities for Holmes."""
 
 import logging
-from typing import Any
+import re
+from typing import Any, Optional
+
+REDACTED = "[REDACTED]"
+
+# Credential-bearing key names that show up in URLs, query strings, headers and
+# JSON bodies (e.g. LiteLLM appends `?key=<gemini-key>` to the request URL, and
+# httpx echoes that URL back inside HTTPStatusError messages / tracebacks).
+_SECRET_KEY_NAMES = (
+    r"api[_-]?key|apikey|access[_-]?token|refresh[_-]?token|token|"
+    r"client[_-]?secret|secret|password|passwd|pwd|authorization|auth|key"
+)
+
+# `?key=...` / `&api_key=...` in URLs and query strings. Stops at the next query
+# separator or whitespace/quote so only the credential value is masked.
+_QUERY_PARAM_RE = re.compile(
+    rf"(?i)([?&](?:{_SECRET_KEY_NAMES})=)[^&\s\"']+"
+)
+
+# `key=value`, `key: value`, `"key": "value"` in headers / JSON / kwargs dumps.
+_KEY_VALUE_RE = re.compile(
+    rf"(?i)([\"']?\b(?:{_SECRET_KEY_NAMES})\b[\"']?\s*[:=]\s*[\"']?)"
+    r"[^\s,&\"'}{]+"
+)
+
+# `Bearer <token>` / `Basic <token>` authorization header values.
+_BEARER_RE = re.compile(r"(?i)\b(bearer|basic)\s+[A-Za-z0-9._\-+/=]+")
+
+# Well-known provider key formats, redacted by value regardless of context so a
+# leaked key is masked even when it appears without a recognizable key name.
+_KNOWN_KEY_FORMATS = [
+    re.compile(r"AIza[0-9A-Za-z_\-]{35}"),  # Google / Gemini
+    re.compile(r"sk-ant-[0-9A-Za-z_\-]{20,}"),  # Anthropic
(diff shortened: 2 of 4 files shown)


ricky9408:
I can see every recent PRs are stuck by 'License Compliance'.
May I ask what is this check?
I will fix it if something is wrong, thanks :)

PR #2152, opened 2026-06-07
Switch holmes image to Alpine base

### LINKED ISSUE #2146: Bug: Pinned LiteLLM version blocks GitHub Copilot endpoint overrides via environment variables
### What happened?

The current LiteLLM pin in this project prevents documented GitHub Copilot endpoint overrides from working for all required endpoints.
With the current pinned version, the package falls back to hardcoded defaults for several URLs instead of honoring environment variables.

When these environment variables are set, LiteLLM should use them instead of hardcoded defaults:

- GITHUB_COPILOT_API_BASE
- GITHUB_COPILOT_DEVICE_CODE_URL
- GITHUB_COPILOT_ACCESS_TOKEN_URL
- GITHUB_COPILOT_API_KEY_URL
This matches LiteLLM provider documentation for GitHub Copilot custom endpoints:
https://docs.litellm.ai/docs/providers/github_copilot

This blocks documented enterprise/custom GitHub Copilot endpoint configuration and can cause integration failures where non-default endpoints are required.
https://holmesgpt.dev/latest/ai-providers/github-copilot/?h=github+co

With the currently pinned version, only API base is present while the other endpoint URLs remain hardcoded, so enterprise/custom endpoint setups are not fully overridable through environment variables.

- Make sure ~/.holmes/model_list.yaml is configured as required for co-pilot
- Set the the env var endpoints as valid alternative enterprise or custom urls
- Run a holmes "ask" operation with a model entry configured in ~/.holmes/model_list.yaml
- Validate output shows no override via env vars has taken place and the hardcoded API base is returned during device auth (GITHUB_COPILOT_DEVICE_CODE_URL is the most visible for initial device auth)

### Version validation performed:

Tested versions in the 1.83.x line:

- 1.83.7: only API base present; device code/access token/api key override vars missing
- 1.83.8: same as 1.83.7
- 1.83.9: same as 1.83.7
- 1.83.10: all four override vars present and wired
- 1.83.11: all four override vars present and wired

I validated that version 1.83.10 is the first version in the 1.83.x line where all required override variables are implemented.

moshemorad:
## Summary

Eliminates all HIGH and CRITICAL CVEs from the holmes Docker image by switching the base from `python:3.11-slim-bookworm` to `python:3.11-alpine`, refreshing the bundled CLI binaries to CVE-patched builds, and bumping vulnerable Python dependencies.

**Scan results (trivy, all severities):**

| Image | Critical | High | Medium | Low |
|---|---|---|---|---|
| holmes:0.31.1 (before) | 4 | 21 | many | many |
| bookworm + patches | 4 | 13 | — | — |
| **Alpine (this PR)** | **0** | **0** | **2 → 0** | **0** |

The Debian base carried 17 HIGH/CRITICAL findings with **no fix available in any Debian release** (bookworm/trixie/sid), including the perl group (CVE-2026-42496 Critical + 5 High) which is structurally unremovable on Debian: `perl-base` is Essential and `git` depends on `perl`. Alpine's git has no perl dependency, so the entire group is absent.

> The 2 Medium aiohttp findings (CVE-2026-34993/47265) from the original scan were subsequently cleared by the dependency bumps below (aiohttp 3.14.1).

## What's in the PR

**Base image**

- Both the builder and final stages now use `python:3.11-alpine`.

**Bundled CLI binaries**

- **kubectl** — official `dl.k8s.io` release binary (**v1.36.2**), pulled at build time with upstream SHA-256 verification (pinned via the `KUBECTL_VERSION` build arg). v1.36.2 is the first kubectl built with Go ≥ 1.26.3 (go1.26.4), so it carries the stdlib CVE fixes natively — no from-source rebuild needed.
- **argocd / helm / kube-lineage** — still rebuilt from source with a pinned Go 1.26.4 toolchain plus dependency `replace` pins (see `scripts/build_go_binaries.sh`), because no upstream release fixes their CVEs yet: argocd v3.3.11 (go-git 5.19.1, go-billy 5.9.0, slack-go 0.23.1), helm v3.21.0 (containerd 1.7.32), kube-lineage v2.2.5 (grpc 1.79.3, spdystream 0.5.1, containerd 1.7.32, embedded helm 3.20.2).

**Python dependencies**

- litellm pinned to **1.89.0** (CRITICAL CVE-2026-49468 auth-bypass; also relaxes the aiohttp pin). Python capped at `<3.14` (litellm ≥1.84 declares `Requires-Python <3.14`).
- Security floors expressed as `>=` (not `^`) so they never cap a newer patched release: aiohttp ≥3.14.1 (CVE-2026-34993/47265), cryptography ≥48.0.1 (GHSA-537c-gmf6-5ccf), starlette ≥1.3.1 (CVE-2026-54283/54282), pydantic-settings ≥2.14.2, python-multipart ≥0.0.31 (CVE-2026-53540).

**Microsoft ODBC** (azure/sql toolset) — download + GPG-signature-verified apk.

## ⚠️ Behavioral changes to review

- **azure/sql toolset is amd64-only on Alpine**: Microsoft ships no working aarch64 Alpine ODBC package (their `_arm64.apk` contains x86_64 metadata). arm64 images skip the driver gracefully.
- **confluent-kafka is source-built** against `librdkafka-dev` from Alpine **edge** (stable ships 2.12.1; the binding needs ≥ 2.14.0). Consider pinning or building librdkafka from source before release.
- **kubectl bumps v1.34.8 → v1.36.2** and is now the official upstream binary (no longer an apt package or a from-source rebuild). kubectl's version-skew policy supports ±1 minor from the API server; it's used here as a read-only troubleshooting client, so wider skew is low-risk.

## Follow-ups (out of scope for this PR)

- **`Dockerfile.operator`** is still `python:3.11-slim` (Debian) — tracked in **ROB-516** (scan, then switch to Alpine if it carries unfixable HIGH/CRITICALs).
- **Non-root `USER`** for the main image (DS-0002) — deferred; needs `CAP_NET_RAW` for tcpdump, a writable HOME for the kubectl/helm/argocd caches, and Helm-chart `securityContext` coordination.

## Validation

- trivy full-severity scan on the Alpine image: 0 Critical / 0 High.
- 2439 unit tests pass (7 pre-existing env-related failures, identical on clean master).
- Bundled binaries smoke-tested on amd64 + arm64 (argocd, helm, kubectl version string, kube-lineage, git, jq; litellm/pyodbc/confluent_kafka imports). kubectl v1.36.2 toolchain confirmed go1.26.4 via `go version`.
- Deployed to an AKS cluster: full toolset init + live LLM chat with kubectl/log tool-calling verified end-to-end.
- Image size: 386 MB → 319 MB compressed.

## Summary by CodeRabbit

* **Infrastructure & Performance**
  * Migrated container build/runtime base images to Alpine, optimizing image size and runtime behavior.
* **Security & Tooling**
  * Updated bundled Kubernetes/Helm/Argo tooling to CVE-patched binaries with refreshed per-architecture checksums and client validation.
  * Improved Microsoft ODBC driver installation to use verified package handling.
  * Rebuilt bundled Go binaries with an updated pinned toolchain and component version pins.
* **Dependencies & Compatibility**
  * Expanded/adjusted supported Python range and refreshed key Python dependencies for security.
* **Documentation**
  * Updated Bash toolset allowlist command lists and archive flag variants.


scripts/build_go_binaries.sh
```diff
@@ -1,39 +1,61 @@
...
 CURRENT_GO_VERSION="$(go env GOVERSION 2>/dev/null | sed 's/^go//')"
 if [ -z "$CURRENT_GO_VERSION" ]; then
-  echo "Go is not installed or not on PATH. Go ${MIN_GO_VERSION}+ is required." >&2
+  echo "Go is not installed or not on PATH. Go 1.21+ is required (GOTOOLCHAIN downloads ${GOTOOLCHAIN#go})." >&2
   exit 1
 fi
 if ! printf '%s\n%s\n' "$MIN_GO_VERSION" "$CURRENT_GO_VERSION" | sort -V -C; then
-  echo "Go ${MIN_GO_VERSION}+ is required (found ${CURRENT_GO_VERSION}). Please upgrade." >&2
+  echo "Go ${MIN_GO_VERSION}+ is required (found ${CURRENT_GO_VERSION}). GOTOOLCHAIN switch failed?" >&2
   exit 1
```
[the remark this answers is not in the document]
moshemorad: @claude WDYT?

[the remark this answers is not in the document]
moshemorad: Applied the portable fix in `scripts/build_go_binaries.sh` (pushed to `claude/hopeful-volta-o1gjj1`):

- Check `command -v go` **before** the `go env GOVERSION` substitution, so a missing `go` doesn't abort the script under `set -e` before we can print the helpful message.
- Replaced the GNU-only `sort -V -C` with a portable `sort -t. -k1,1n -k2,2n -k3,3n | head -n1` comparison that works on BSD/macOS too. Verified against `1.26.3` vs `1.26.2/1.26.3/1.26.4/1.27.0/2.0.0`.

I also picked up the other review items while here:
- **Dead `. /venv/bin/activate`** in the Dockerfile venv RUN — removed (the venv is already selected via `VIRTUAL_ENV`/`PATH`).
- **Non-root USER (DS-0002)** — deferring to a follow-up, as you and CodeRabbit both noted; it needs `CAP_NET_RAW` for tcpdump, a writable HOME for the kubectl/helm/argocd caches, and coordination with the Helm chart's `securityContext`, which is out of scope for this CVE pass.

Separately, a fresh OSV scan of the locked deps surfaced advisories published **after** the original docker scout run, including a **CRITICAL** in litellm. Patched in the two latest commits: litellm `1.83.14→1.84.9` (CVE-2026-49468 auth-bypass; also relaxes the `aiohttp==3.13.4` pin so we take aiohttp `3.14.1`, clearing CVE-2026-34993/47265 + the 3.14.x hardening CVEs), cryptography `48.0.0→48.0.1` (HIGH OpenSSL-in-wheels), starlette `1.2.1→1.3.1` (HIGH CVE-2026-54283), pydantic-settings `2.14.1→2.14.2`, python-multipart `0.0.30→0.0.31`. Only 6 packages change in the lock; the lone remaining OSV hit is `pytest` (dev-only group, not in the image).

---

galvesLoy:
@moshemorad do you have any idea when this PR will be deployed ? 🙏 

Thank you

pyproject.toml
```diff
@@ -65,9 +70,18 @@ kopf = "^1.37.0"
...
+# Starlette >= 1.3.1 fixes CVE-2026-54283 (HIGH: request.form() size limits
+# silently ignored) and CVE-2026-54282; >= 1.0.1 already fixed CVE-2026-48710 /
+# GHSA-86qp-5c8j-p5mr (Host header validation when reconstructing request.url).
+starlette = "^1.3.1"
+# cryptography >= 48.0.1 ships a patched OpenSSL in its wheels (GHSA-537c-gmf6-5ccf, HIGH).
+cryptography = "^48.0.1"
+# pydantic-settings >= 2.14.2 fixes GHSA-4xgf-cpjx-pc3j (symlink escape out of secrets_dir).
+pydantic-settings = "^2.14.2"
+# python-multipart >= 0.0.31 fixes CVE-2026-53540 (negative Content-Length buffers the whole body).
+python-multipart = "^0.0.31"
```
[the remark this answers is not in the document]
moshemorad: Fixed in the merge commit. Changed `python-multipart = "^0.0.31"` → `">=0.0.31"`. You're right that Poetry reads `^0.0.31` as `==0.0.31` for `0.0.x` — and after the change the lock now resolves to **0.0.32**, which the caret would have frozen out. I applied the same caret→`>=` fix to the other security-floor pins (starlette, cryptography, pydantic-settings, aiohttp), since one of them (`cryptography ^48.0.1`) was actively downgrading master's already-resolved 49.0.0.

---

docs/data-sources/builtin-toolsets/bash.md
```diff
@@ -60,16 +60,16 @@ The `builtin_allowlist` field controls which commands are pre-approved:
...
 
 **`extended`** (Helm default) - adds these on top of `core`:
 
 | Category | Commands | Why container-only |
 |----------|----------|--------------------|
-| File reading | `cat`, `echo`, `base64` | Can read sensitive files (~/.ssh, ~/.aws) on local machines |
+| File reading | `cat`, `base64` | Can read sensitive files (~/.ssh, ~/.aws) on local machines |
 | Filesystem | `ls`, `find`, `stat`, `du`, `df` | Exposes local filesystem structure |
-| Archives | `tar -tf`, `gzip -l`, `zcat`, `zgrep` | Can inspect local archives |
+| Archives | `tar -tf`, `tar -tvf`, `tar -tfv`, `tar -ftv`, `gzip -l`, `zcat`, `zgrep` | Can inspect local archives |
```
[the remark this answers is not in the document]
moshemorad: Good catch — fixed. Removed `tar -tfv` and `tar -ftv` from both the docs **and** the implementation (`default_lists.py`), keeping `tar -tf` / `tar -tvf`. As you note, in a short-option cluster `-f` consumes the rest of the cluster as its archive-name argument, so those forms could only ever match broken `tar` commands. They were dead entries in the allowlist, so dropping them loses nothing.

---

moshemorad:
Hi @galvesLoy We will push it in the next couple of days.
Running last checks and if all good we will proceed with it

scripts/build_go_binaries.sh
```diff
@@ -1,39 +1,70 @@
...
+# Pin the build toolchain: Go >= 1.26.3 fixes stdlib
+# CVE-2026-42499/33814/39836/33811/39820. Go auto-downloads it if the locally
+# installed go is older (requires local go >= 1.21).
+export GOTOOLCHAIN=go1.26.4
+
+MIN_GO_VERSION="1.26.3"
+# Check for the go binary first: under `set -e` the command substitution below
+# would otherwise abort the script before the empty-string check could run.
+if ! command -v go >/dev/null 2>&1; then
+  echo "Go is not installed or not on PATH. Go 1.21+ is required (GOTOOLCHAIN downloads ${GOTOOLCHAIN#go})." >&2
```
naomi-robusta: use MIN_GO_VERSION instead of 1.21+

scripts/build_go_binaries.sh
```diff
@@ -125,13 +170,40 @@ echo "==> Building kube-lineage for linux/arm64..."
 CGO_ENABLED=0 GOOS=linux GOARCH=arm64 GOFLAGS=-mod=mod go build \
   -o "$OUTDIR/arm64/kube-lineage" ./cmd/kube-lineage
 
+echo "==> Cloning kubernetes $KUBECTL_VERSION (kubectl)..."
+git clone --depth 1 --branch "$KUBECTL_VERSION" https://github.com/kubernetes/kubernetes.git "$TMPDIR/kubernetes"
```
naomi-robusta: we're building kubectl from source all of the sudden? This feels risky.

scripts/build_go_binaries.sh
```diff
@@ -87,13 +128,12 @@ CGO_ENABLED=0 GOOS=linux GOARCH=arm64 GOFLAGS=-mod=mod go build \
 echo "==> Cloning Helm $HELM_VERSION..."
 git clone --depth 1 --branch "$HELM_VERSION" https://github.com/helm/helm.git "$TMPDIR/helm"
```
naomi-robusta, line 134: 1. we should delete this folder on cleanup
2. we shouldn't use a deterministic path. better use:

```
BUILD_TMP_DIR=$(mktemp -d) || exit 1
cleanup() {
    rm -rf "$MY_TMP_DIR"
}
trap cleanup EXIT
git clone --depth 1 --branch "$HELM_VERSION" https://github.com/helm/helm.git "$BUILD_TMP_DIR/helm"
pushd "$BUILD_TMP_DIR/helm" 2>&/dev/null
```

scripts/build_go_binaries.sh
```diff
@@ -87,13 +128,12 @@ CGO_ENABLED=0 GOOS=linux GOARCH=arm64 GOFLAGS=-mod=mod go build \
...
 
-echo "==> Pinning grpc to $GRPC_PATCHED_VERSION (CVE-2026-33186)..."
 cd "$TMPDIR/helm"
```
naomi-robusta, line 136: use pushd & popd instead of cd in scripts

scripts/build_go_binaries.sh
naomi-robusta: also use $BUILD_TMP_DIR here
and pushd & popd

scripts/build_go_binaries.sh
```diff
@@ -125,13 +170,40 @@ echo "==> Building kube-lineage for linux/arm64..."
 CGO_ENABLED=0 GOOS=linux GOARCH=arm64 GOFLAGS=-mod=mod go build \
   -o "$OUTDIR/arm64/kube-lineage" ./cmd/kube-lineage
 
+echo "==> Cloning kubernetes $KUBECTL_VERSION (kubectl)..."
```
naomi-robusta: You're cloning kubectl master here, not $KUBECTL_VERSION

scripts/build_go_binaries.sh
```diff
@@ -125,13 +170,40 @@ echo "==> Building kube-lineage for linux/arm64..."
...
+git clone --depth 1 --branch "$KUBECTL_VERSION" https://github.com/kubernetes/kubernetes.git "$TMPDIR/kubernetes"
+cd "$TMPDIR/kubernetes"
+
+KUBECTL_MAJOR="$(echo "${KUBECTL_VERSION#v}" | cut -d. -f1)"
+KUBECTL_MINOR="$(echo "${KUBECTL_VERSION#v}" | cut -d. -f2)"
+KUBECTL_LDFLAGS="-w -s \
+  -X k8s.io/component-base/version.gitVersion=$KUBECTL_VERSION \
+  -X k8s.io/component-base/version.gitMajor=$KUBECTL_MAJOR \
+  -X k8s.io/component-base/version.gitMinor=$KUBECTL_MINOR \
+  -X k8s.io/component-base/version.gitTreeState=clean"
```
naomi-robusta: same issue, you're assuming the kubectl version, although you cloned master here. Maybe grab the version from README.md or from CHANGELOG folder?

Dockerfile
```diff
@@ -1,64 +1,74 @@
...
-# Create and activate virtual environment.
-# Upgrade wheel to >= 0.46.2 to fix CVE-2026-24049 (path traversal); the version
-# pulled in by --upgrade-deps (0.45.1) is vulnerable.
+# Create venv; upgrade wheel (CVE-2026-24049) and pip (CVE-2026-3219/6357, CVE-2025-8869).
+# The venv is selected via VIRTUAL_ENV/PATH below (sourcing activate in a RUN has
+# no effect — the shell exits when the layer finishes).
 RUN python -m venv /venv --upgrade-deps && \
-    /venv/bin/pip install --upgrade 'wheel>=0.46.2' && \
-    . /venv/bin/activate
+    /venv/bin/pip install --upgrade 'wheel>=0.46.2' 'pip>=26.1'
```
naomi-robusta: nit: extract versions to the env vars at the top

Dockerfile
```diff
@@ -147,20 +166,13 @@ ARG AWS_REGION
...
-RUN rm -rf /usr/local/lib/python3.11/ensurepip/_bundled/setuptools-65.5.0-py3-none-any.whl
-
-# Upgrade wheel + setuptools in the base image's system Python to fix CVE-2026-24049
-# (wheel 0.45.1 path traversal). The venv at /venv was already upgraded in the builder stage,
-# but the base image's system Python still has the vulnerable copy.
-RUN /usr/local/bin/pip install --upgrade --no-cache-dir 'wheel>=0.46.2' 'setuptools>=80.0.0' \
-    && rm -rf /usr/local/lib/python3.11/site-packages/wheel-0.45.1.dist-info
+# Upgrade base-image system Python's wheel/setuptools/pip (CVE-2026-24049,
+# CVE-2026-3219/6357, CVE-2025-8869, CVE-2026-1703).
+RUN /usr/local/bin/pip install --upgrade --no-cache-dir 'wheel>=0.46.2' 'setuptools>=80.0.0' 'pip>=26.1'
```
naomi-robusta: duplicate versions as before, should be an env var

Dockerfile
```diff
@@ -1,64 +1,74 @@
+# Alpine-based image (switched from Debian bookworm to drop unfixable perl
+# CVEs; Alpine git has no perl dependency).
+
 # Build stage
-FROM python:3.11-slim-bookworm as builder
+FROM python:3.11-alpine AS builder
```
naomi-robusta, line 14: what about Dockerfile.operator?

naomi-robusta (changes requested):

naomi-robusta:
> confluent-kafka is source-built against librdkafka-dev from Alpine edge (stable ships 2.12.1; the binding needs ≥ 2.14.0). Consider pinning or building librdkafka from source before release.
Should this be pinned?

scripts/build_go_binaries.sh
moshemorad: Agreed — and it's no longer needed, so I've removed the kubectl-from-source build entirely.

The rebuild only existed to get a patched Go toolchain: at the time, every published kubectl release (through v1.36.1) was compiled with go1.26.2, vulnerable to the stdlib CVEs. That's no longer true — **kubectl v1.36.2 ships built with go1.26.4**, the exact toolchain our other binaries use, so the official binary now carries the same fixes. I verified directly:

| kubectl | Go toolchain | stdlib CVEs fixed (need ≥1.26.3) |
|---|---|---|
| v1.36.2 (latest) | go1.26.4 | ✅ |
| v1.36.1 / v1.36.0 | go1.26.2 | ❌ |
| v1.35.6 / v1.34.9 | go1.25.11 | ❌ |

So the Dockerfile now pulls the official **v1.36.2** from dl.k8s.io with upstream SHA-256 verification (pinned via `KUBECTL_VERSION`), and `build_go_binaries.sh` no longer builds kubectl. I also removed the committed `bin/go-cve-rebuild/*/kubectl.gz` artifacts (~34 MB). argocd/helm/kube-lineage are still rebuilt — no upstream release fixes their CVEs yet.

This also resolves your other notes on this block (cloning the wrong ref / assuming the version from a master clone) — that code is gone.

One trade-off worth a conscious call: this bumps the bundled kubectl **v1.34.8 → v1.36.2**. Only the 1.36 line has the go1.26.4 build; v1.34.9 and v1.35.6 are still on go1.25.11, so staying on 1.34/1.35 would mean keeping the rebuild. kubectl's skew policy supports ±1 minor from the API server, and it's used here as a read-only troubleshooting client, so the wider skew should be low-risk — but flagging it in case you'd prefer a different pin.

---

scripts/build_go_binaries.sh
moshemorad: Extracted into a named constant. Note I kept it as a *separate* `MIN_BOOTSTRAP_GO_VERSION="1.21"` rather than reusing `MIN_GO_VERSION` — they're genuinely two different floors: `MIN_GO_VERSION=1.26.3` is the build toolchain (enforced below), while `1.21` is the minimum *local* Go that can bootstrap the `GOTOOLCHAIN` auto-download (GOTOOLCHAIN landed in Go 1.21, and the pinned toolchain is fetched automatically). The message now interpolates the constant instead of a bare literal.

---

Dockerfile
moshemorad: Done — hoisted into `PIP_MIN_VERSION` / `WHEEL_MIN_VERSION` / `SETUPTOOLS_MIN_VERSION` ARGs at the top of the file (with the CVE rationale), re-declared per stage and referenced via `${...}`.

---

Dockerfile
moshemorad: Done — the duplicated `wheel`/`pip`/`setuptools` floors now come from the shared ARGs at the top (`PIP_MIN_VERSION` / `WHEEL_MIN_VERSION` / `SETUPTOOLS_MIN_VERSION`), so the builder and final stages no longer repeat the literals. (Docker ARGs don't cross `FROM`, so each stage re-declares the bare `ARG` to pick up the global default — that's the minimal way to dedup here.)

---

scripts/build_go_binaries.sh
moshemorad: This is already in place at the top of the script — the temp dir is non-deterministic and cleaned up on exit:

```bash
TMPDIR=$(mktemp -d)
trap "rm -rf $TMPDIR" EXIT
```

All clones go under `$TMPDIR/...` (the `/helm`, `/argo-cd` leaves sit inside that random parent, so there's no fixed path or collision risk), and the `trap` removes the whole tree on any exit. So both points are covered.

---

scripts/build_go_binaries.sh
moshemorad: Each section `cd`s into a fresh **absolute** path (`$TMPDIR/helm`, `$TMPDIR/argo-cd`, …) and never relies on returning to a previous directory, so `cd` is safe here and `pushd`/`popd` wouldn't change behavior. Happy to switch to `pushd`/`popd` if you'd prefer it as a house-style convention — let me know.

---

Dockerfile
moshemorad: Good call — out of scope for this CVE pass, so I've tracked it as a follow-up: **[ROB-516](https://linear.app/robusta/issue/ROB-516)** (assigned to @moshemorad). It covers scanning `Dockerfile.operator` (still `python:3.11-slim`) and switching it to Alpine if it carries unfixable HIGH/CRITICALs like the main image did.

---

naomi-robusta (approved):

Dockerfile
```diff
@@ -1,64 +1,93 @@
...
+# verification. v1.36.2 is the first kubectl built with Go >= 1.26.3 (go1.26.4),
+# so it already carries the stdlib CVE fixes (CVE-2026-42499/33814/39836/33811/
+# 39820/39823/39825/39826/42504) that previously forced a from-source rebuild --
+# it is now built with the exact toolchain our other bundled binaries use.
+# argocd/helm/kube-lineage are still rebuilt (see scripts/build_go_binaries.sh)
+# because no upstream release fixes their CVEs yet. Bump KUBECTL_VERSION as newer
+# releases ship (the 1.34/1.35 lines are still on a vulnerable Go toolchain).
 ARG TARGETARCH
+ARG KUBECTL_VERSION=v1.36.2
+RUN cd /tmp \
```
naomi-robusta, line 57: should this be /tmp or $TMPDIR?
There's a lot uf usage of /tmp in the rest of the file as well

naomi-robusta (approved):

PR #2181, opened 2026-06-11
Send explicit max_tokens on every LLM completion request

moshemorad:
## Problem

A customer reported Holmes responses getting cut off mid-answer. Their chat metadata showed:

```json
"max_completion_tokens_per_call": 4096,
"finish_reason": "length",
"max_output_tokens": 64000,
"max_tokens": 1000000
```

`completion_tokens` landed at exactly 4096 with `finish_reason: "length"` — the model hit a hard 4096 output cap, even though Holmes computed (and reported) a 64000-token output budget.

**Root cause:** `get_maximum_output_token()` is used to reserve output space during input budgeting and compaction (`input_context_window_limiter.py`, `compaction.py`), but it was never sent on the actual request — `DefaultLLM.completion()` passed no `max_tokens` to litellm. litellm then falls back to provider defaults. For Anthropic-family models, litellm resolves the default from its cost map; when the model name isn't in the map (proxy aliases, custom gateways — the same situation that makes users configure `max_context_size` by hand), it falls back to `DEFAULT_ANTHROPIC_CHAT_MAX_TOKENS = 4096`. Long answers get silently truncated while the metadata claims a 64000 budget.

## Fix

`DefaultLLM.completion()` now always sends an explicit `max_tokens`, using the same value input budgeting already reserves. Precedence:

1. Explicit `max_tokens` / `max_completion_tokens` in model args — always wins (and a user-set `max_completion_tokens` blocks injection so no conflicting pair is sent).
2. `OVERRIDE_MAX_OUTPUT_TOKEN` env var — now actually reaches the request instead of only affecting compaction math.
3. Computed: `min(64000, context_window / 5)`, capped by the model's `max_output_tokens` from litellm's cost map when known.

`max_tokens: null` / `max_completion_tokens: null` config sentinels are stripped, mirroring the existing `temperature: null` handling (PR #698 semantics).

Provider safety: litellm 1.83.7 translates `max_tokens` per provider — `max_completion_tokens` for OpenAI o-series/gpt-5 reasoning models, `maxTokens` for Bedrock converse, `maxOutputTokens` for Gemini — so sending it is safe across providers (verified against the pinned litellm).

## Changes

- `holmes/core/llm.py` — inject `max_tokens` in `DefaultLLM.completion()` (all three call sites benefit: agentic loop, compaction, fast-model summarization)
- `tests/core/test_llm_completion_max_tokens.py` — new behavior-matrix tests, including a reproduction of the customer scenario (unknown model + `max_context_size: 1000000` → `max_tokens: 64000`, not 4096)
- `tests/core/test_llm_completion_temperature.py`, `tests/core/test_llm_completion_cache_control.py` — test helpers that bypass `__init__` now set `max_context_size`
- `docs/reference/context-management.md` — document the output token limit and its resolution order

## Testing

- `tests/core/test_llm_completion_max_tokens.py` — 8 new tests, all pass
- Full non-LLM suite: **2487 passed, 91 skipped** (skips are missing-credential environment skips, pre-existing)

## Note for operators

Models unknown to litellm's cost map now receive a computed `max_tokens` instead of none. If the computed value exceeds what the upstream model actually supports, the provider may reject the request with an explicit error instead of silently truncating at its default — set `OVERRIDE_MAX_OUTPUT_TOKEN` or `max_tokens` in model args to the correct value for the model.

https://claude.ai/code/session_01LPu4dG5LMjRgQgpBUsWhKw

---
_Generated by [Claude Code](https://claude.ai/code/session_01LPu4dG5LMjRgQgpBUsWhKw)_

## Release Notes

* **Bug Fixes**
  * LLM completion requests now consistently include an explicit output-token budget to help prevent mid-response truncation.
  * Output-token cap resolution is updated with clearer precedence and correct behavior when `max_tokens` and `max_completion_tokens` are both present.

* **Documentation**
  * Context management docs now include an “Output Token Limit” section explaining the enforced cap and its resolution order.

* **Tests**
  * Added coverage to verify output-token limit injection, stripping, environment overrides, and model-specific capping.


holmes/core/llm.py
```diff
@@ -762,7 +777,13 @@ def completion(
...
 
     def get_maximum_output_token(self) -> int:
-        max_output_tokens = floor(min(64000, self.get_context_window_size() / 5))
+        # Reserve output budget = max(64k, 12% of the context window). The 64k
+        # floor keeps small and unknown models usable (the 200k fallback window
+        # gives 12% = 24k, so they stay at 64k), while large windows scale up:
+        # a 1M-context model reserves 120k. The crossover is ~533k. This value
+        # is still capped below to the model's real max_output_tokens when the
+        # model is known to litellm.
+        max_output_tokens = max(64000, self.get_context_window_size() * 12 // 100)
```
naomi-robusta, line 786: nit: 65536

naomi-robusta (approved):

PR #2199, opened 2026-06-17
Fix chat input multiline handling

kayahk:
## Summary
- replace the HolmesGPT chat text input with a multiline textarea
- make plain Enter insert a newline instead of submitting, including responsive/mobile keyboards
- keep keyboard submit available via Cmd/Ctrl+Enter and preserve history navigation for single-line prompts
- preserve submitted prompt formatting in the chat transcript, including line breaks and inline `code` styling, with long prompts collapsible
- add a chat font-size toggle for more readable output
- add a dark mode toggle that applies across the portal services and chat UI

## Verification
- `git diff --check`
- `CI=false yarn build` (passes; existing ESLint warnings remain)
- `CI=true yarn build` previously failed because pre-existing warnings are treated as errors: hook dependency and unused variable warnings in ChatAssistant/MainContent

## Release Notes

* **New Features**
  * Chat input now supports multiline prompts using a textarea.
  * Send is triggered by **Ctrl+Enter** (**Cmd+Enter** on Mac) instead of plain Enter.
  * Message history navigation with arrow keys is limited to single-line prompts.
  * Prompts can auto-collapse when long, with a toggle to expand/collapse.
  * Added **Large text** and **Dark mode** toggles.

* **UI Improvements**
  * Updated prompt/chat styling (including code formatting), header controls, and hover/disabled behavior.
  * Improved markdown rendering in assistant messages.

### DIFF
--- a/experimental/ag-ui/front-end/src/App.css
+++ b/experimental/ag-ui/front-end/src/App.css
@@ -87,6 +87,25 @@ html, body {
   font-size: 1.1rem;
 }
 
+.sidebar-theme-control {
+  padding: 0 1rem 1rem;
+}
+
+.theme-toggle-button {
+  width: 100%;
+  padding: 0.6rem 0.75rem;
+  border-radius: 999px;
+  border: 1px solid rgba(255, 255, 255, 0.25);
+  background-color: rgba(255, 255, 255, 0.1);
+  color: white;
+  cursor: pointer;
+  font-weight: 600;
+}
+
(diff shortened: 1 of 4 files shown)


kayahk:
Addressed the CodeRabbit IME composition concern in `b09902aef` by returning early from chat history shortcut handling while `nativeEvent.isComposing` is true. Re-ran `CI=false yarn build` and `git diff --check` successfully.

PR #2200, opened 2026-06-17
ROB-4017: Harden Supabase DAL connection (HTTP/1.1, CA bundle, RemoteProtocolError retry)

moshemorad:
## Summary (ROB-4017)

Hardens the Supabase DAL client against intermittent connection failures (dropped `HolmesStatus` upserts, dropped conversation claims) caused by a shared client under concurrency and by Supabase's edge closing idle keep-alive connections.

This is a **single PR** consolidating what were two stacked branches for the same ticket — the base client-wiring and the `RemoteProtocolError` retry follow-on. It now targets `master` directly and supersedes #2122 (which can be closed).

## Changes

- **Force HTTP/1.1** for the Supabase client. httpcore's *sync* HTTP/2 connection isn't thread-safe, and one `SupabaseDal` client is shared across the conversation worker, realtime callbacks and request threads — under concurrency the HTTP/2 framing corrupts and calls fail with `RemoteProtocolError: Server disconnected`.
- **Honor the environment CA bundle** (`SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE`) via an explicit `SSLContext`, rather than letting our own httpx client fall back to certifi (which breaks TLS verification behind an intercepting proxy). Passes an `SSLContext`, not a path string, since httpx has deprecated `verify=<str>`.
- **Retry `RemoteProtocolError` at the transport** (`SupabaseRetryTransport`, an `httpx.HTTPTransport` subclass using tenacity's `@retry`). Even on HTTP/1.1, Supabase's edge closes idle keep-alives, so a reused connection fails with "Server disconnected" *before* the request reaches Supabase — safe to replay on a fresh connection. Hardening at the transport covers every sub-client (postgrest, auth, storage, realtime) uniformly, with no backoff (a reaped socket just needs a fresh connection).
- **`CLAUDE.md`**: documents the convention to prefer `tenacity` for retries over hand-rolled loops.

## Tests

Deterministic, no network: `test_supabase_dal_retry.py` pins the retry contract (retry-then-succeed, budget exhaustion, non-retryable passthrough); `test_supabase_dal_transport.py` pins the client wiring (transport built, `http2` off, `verify` as `SSLContext`, client forwarded to postgrest).

Mirrors the relay-side fix (ROB-4012).

https://claude.ai/code/session_015thi2RsgXWp8BzuDCfxNzG

## Release Notes

* **Bug Fixes**
  * Improved Supabase connectivity by retrying transient remote-protocol disconnects (with retry attempt logging) while avoiding retries for non-transient failures.
  * Unified HTTP client and TLS certificate verification behavior across the data layer.
* **Tests**
  * Added coverage for retry success, retry exhaustion, and immediate failure for non-retriable errors.
  * Expanded tests to confirm correct transport and TLS configuration.
* **Documentation**
  * Updated contribution guidance to prefer tenacity-based retries over custom retry loops.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -217,6 +217,7 @@ For the complete eval CLI reference (flags, env vars, model comparison, debuggin
 - Type hints required (mypy configuration in pyproject.toml)
 - Pre-commit hooks enforce quality checks in CI
 - **ALWAYS place Python imports at the top of the file**, not inside functions or methods
+- **Use `tenacity` for retries**, not hand-rolled retry loops — the `@retry(...)` decorator normally, or the `Retrying(...)` iterator form when the attempt budget/params must come from per-instance or dynamic values. Only hand-roll a retry when there's a real reason to, and document why.
 - **NEVER run `pre-commit`, `ruff`, or `mypy` unless the user explicitly asks you to**. These tools are triggered by commit hooks which are not installed on all machines, and running them causes widespread formatting/type changes to files unrelated to your task. Only lint/format files you are actively editing, and only if asked.
 
 **Documentation Examples**:
--- a/holmes/core/supabase_dal.py
+++ b/holmes/core/supabase_dal.py
@@ -4,12 +4,14 @@ import gzip
 import json
 import logging
 import os
+import ssl
 import threading
 from datetime import datetime, timedelta
 from enum import Enum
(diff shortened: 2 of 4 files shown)


naomi-robusta (approved):
LGTM. Just make sure this works in on-prem

PR #2208, opened 2026-06-19
Fix/stream disconnect thread hang

gesila1073:
## Summary by CodeRabbit

* **New Features**
  * Improved cancellation for in-progress LLM streaming, including while tool calls are executing.
  * Streaming now responds to client disconnects by stopping downstream work.
* **Bug Fixes**
  * Interrupted LLM streams are handled gracefully, avoiding unintended error responses in the stream.
* **Tests**
  * Added tests for interruption handling and end-to-end streaming cleanup on both normal completion and exceptions.


gesila1073:
@aantn We are actively using HolmesGPT in our production ops workflow for AI-assisted incident investigation via the streaming API.

We found a critical issue in real usage: when clients cancel streaming investigations, especially under concurrent traffic, subsequent requests can hang because in-flight resources are not released promptly.

Without this fix, client-cancel under concurrency can cause subsequent investigations to hang, which directly blocks new troubleshooting sessions and impacts incident response.

This PR fixes the cancellation path so a client cancel can terminate in-flight investigation work cleanly and reliably, preventing follow-up requests from getting stuck.

This is important for production stability on our side, and we would really appreciate including this in an official release so we can safely upgrade our application.

Thank you for reviewing.

gesila1073:
@aantn hello cloud you merge it？

PR #2210, opened 2026-06-21
ROB-320: Signed approval tickets (JWT)

RoiGlinik:
Every tool call marked pending_approval gets a server-minted HS256 JWT bound to {tool_call_id, tool_name, args_hash}. The resume path refuses to execute any pending_approval whose ticket doesn't verify.

Surface:
- holmes/utils/approval_tickets.py — JWT mint/verify, canonical args hash, signing-key loader. HOLMES_APPROVAL_SIGNING_KEY env var (used as-is, no decoding); ephemeral fallback when unset.
- holmes/core/models.py — PendingToolApproval gains approval_ticket.
- holmes/core/tool_calling_llm.py — mint at the pending_approval site; verify before executing tool_decisions; on failure synthesize a denial decision so the existing deny pipeline produces a TOOL_RESULT with ERROR status and the LLM explains the rejection to the user in chat. No new SSE event types, no relay shim, no FE-side wire changes.
- server.py — one-line INFO log when the env var is set.
- docs/reference/environment-variables.md — new HOLMES_APPROVAL_SIGNING_KEY section with openssl key-gen snippet and 30-day TTL note.

Tests:
- tests/test_approval_tickets.py — mint/verify round-trip, expiry, tampering, alg=none rejection, canonicalization, key-loader fallback.
- tests/test_approval_ticket_security.py — security regressions for forged pending_approval, tampered args, cross-call ticket reuse, and the happy-path round-trip.
- tests/test_edit_command_removed.py — adapted to attach a real ticket to the synthetic pending_approval message.

Replay protection (jti + redemption cache) and multi-replica caveats are intentionally out of scope here.

## Summary by CodeRabbit

* **New Features**
  * Added signed approval tokens to gate resuming tool execution, binding approvals to the specific tool call and its arguments for stronger tamper protection.
  * Approval tokens automatically expire after 30 days.
* **Bug Fixes**
  * Improved rejection handling to ensure invalid or tampered approvals are safely denied and never executed; orphaned pending approvals are cleaned up.
* **Documentation**
  * Documented `HOLMES_APPROVAL_SIGNING_KEY` for configuring the HMAC signing key used for tool approvals.
* **Tests**
  * Added security and unit test coverage for token signing, verification failures, and replay/tampering scenarios.


moshemorad:
@claude review

docs/reference/environment-variables.md
```diff
@@ -89,6 +89,41 @@ docker run -d \
   ...
 ```
 
+### HOLMES_APPROVAL_SIGNING_KEY
+**Default:** not set (an ephemeral key is generated per process at startup)
+
+HMAC signing key for tool-approval tickets. Holmes mints a short-lived JWT
```
moshemorad: What is ticket? why we are using this terminology? is that a known terminology when signing?

holmes/utils/approval_tickets.py
```diff
@@ -0,0 +1,106 @@
...
+    guessable string silently weakens HMAC; that's an operator-trust call
+    (see docs for `HOLMES_APPROVAL_SIGNING_KEY`).
+    """
+    raw = os.environ.get("HOLMES_APPROVAL_SIGNING_KEY", "").strip()
+    if raw:
+        return raw, True
+    return secrets.token_bytes(32), False
+
+
+SIGNING_KEY, SIGNING_KEY_FROM_ENV = _load_signing_key()
```
moshemorad: We are using SIGNING_KEY_FROM_ENV only for the logging? can't we just add the log to the function itself instead? we plan to use it in other cases?

holmes/utils/approval_tickets.py
```diff
@@ -0,0 +1,106 @@
...
+
+def verify_ticket(
+    ticket: Optional[str],
+    tool_call_id: str,
+    tool_name: str,
+    args_json: Optional[str],
+) -> None:
+    """Verify a ticket. Raises `ApprovalTicketError` on any failure."""
+    if not ticket:
+        raise ApprovalTicketError()
```
moshemorad: Can we add relevant message for each of those so we will know what was the problem in the verification?

moshemorad (changes requested):

docs/reference/environment-variables.md
RoiGlinik: changed to token

holmes/utils/approval_tickets.py
RoiGlinik: yea just some left over from old behavior removing that

moshemorad (approved):

PR #2214, opened 2026-06-22
ROB-424 ROB-425 ROB-665: fix compaction on Bedrock Converse gateways (tools-attached summarization, user-role summary)

moshemorad:
## Summary

Fixes three compaction failures on Bedrock and gateways that translate to Bedrock Converse (Kong AI Gateway, LiteLLM proxy):

- Compaction summarization call sent tool-call history without `tools`, so Converse-translating gateways rejected it: `"The toolConfig field must be defined when using toolUse and toolResult content blocks."`
- With extended thinking enabled, the assistant-role summary carried a signed thinking block; the trailing `role:"system"` "Continue." sentinel was hoisted by litellm, fusing the summary into the next assistant turn → Bedrock 400: `"thinking or redacted_thinking blocks in the latest assistant message cannot be modified"` (reproduced live byte-for-byte, `messages.1.content.1`).

## Changes

**Summarization request (primary path)**
- Keeps the conversation in its native shape (tool blocks included) and attaches the agentic loop's `tools` — the same request shape as every agentic call, so `toolConfig` is always defined and the request extends the previous call's prompt-cache prefix (this is the largest request Holmes makes).
- The compaction prompt now starts with a no-tools preamble instructing the model to reply with plain text only.

**Fallback (compatibility net)**
- If the primary attempt fails or the model responds with a tool call, retry once with `_flatten_tool_messages_for_compaction()` — tool calls/results rewritten as plain text, no `tools` attached. Accepted by every OpenAI-compatible endpoint. (In eval runs this also rescued histories whose recorded tool arguments were invalid JSON that Bedrock conversion can't parse.)
- **Fallback usage is observable**: `CompactionResult` records `fallback_used` / `fallback_reason`, surfaced in the `conversation_history_compacted` event metadata — so an unexpected fallback (which forfeits the primary request shape) is diagnosable from conversation events alone.

**Compacted history shape** (was `[system?, last_user, assistant summary, trailing system "Continue."]`)
- Now `[system?, user summary (with embedded continue directive), last_user]` — the summary is a **user** message built from response text only, so thinking blocks are never replayed and no non-leading system message remains. Matches Claude Code's post-compaction shape.
- `CompactionResult` gains a `summary` field; `compact_if_necessary` passes `tools` down and reads the summary from it.

## Verification

- Live against Bedrock with thinking enabled and `litellm.modify_params=False`: the previously-failing post-compaction iteration now passes.
- Through a vanilla LiteLLM proxy (no `modify_params`): tools-attached primary succeeds with no fallback; the old tool-less shape reproduces the customer error and is rescued by the fallback.
- Live compaction runs against Bedrock via real `/api/chat` conversations confirm the primary path executes with `fallback_used: false`.
- All 7 compaction evals pass through the proxy topology; unit tests cover request/output shapes, both fallback triggers, and `fallback_used`/`fallback_reason` reporting. Full `tests/core` suite: 655 passed.
- Eval fixture repair: 5 recorded tool calls had invalid JSON escapes (lost escaping level on `grep 'a\|b'` commands); fixed so arguments parse while keeping commands byte-identical.


holmes/plugins/prompts/conversation_history_compaction.jinja2
```diff
@@ -1,3 +1,5 @@
+IMPORTANT: Do not use or call any tools for this response, even if tools are available. Respond only with plain text containing the analysis and summary requested below.
```
Avi-Robusta, line 1: Is it possible to exclude the tool definitions from the compaction LLM completion while still including the tool results?
Removing it could reduce input tokens and potentially avoid this issue.
Or does Bedrock tend to hallucinate tool calls when no tools are provided?

moshemorad: Unfortunately no — that combination (tool results in history, no tool definitions attached) is exactly what the old code sent, and it's the thing Bedrock Converse hard-rejects with the customer-reported 400: `"The toolConfig field must be defined when using toolUse and toolResult content blocks."` It's an API-level validation on Converse (and on gateways translating to it — Kong, LiteLLM proxy), not a hallucination concern, so there are only two self-consistent shapes:

1. **Native tool blocks + tools attached** — the primary path in this PR.
2. **No tools + tool blocks flattened to plain text** — the PR's fallback (`_flatten_tool_messages_for_compaction`), accepted by every OpenAI-compatible endpoint.

On the token cost of attaching the definitions: it's cheaper than it looks. The summarization request then has the *same prefix* (system + tools + conversation) as the previous agentic call, so the tool definitions are prompt-cache reads (~10% of input price) rather than fresh tokens — while the flattened no-tools shape shares no prefix with anything and re-tokenizes the entire history at full price. Verified this live against Bedrock: the tools-attached primary runs with `fallback_used: false`.

Re hallucinated tool calls with tools attached: it happens occasionally, and it's handled — the prompt starts with a no-tools preamble, and if the model calls a tool anyway we detect it (`fallback_reason: "model responded with a tool call instead of a summary"`) and retry once via the flattened fallback.

---

Avi-Robusta: why is toolUse and toolResult passed to compaction, itsnt it all flattened?

Avi-Robusta: Ah got it, we pass everytthing by default

moshemorad:
/eval
tags: compaction

holmes/core/truncation/compaction.py
```diff
@@ -127,47 +251,86 @@ def compact_conversation_history(
...
-    ):
-        response_message = response.choices[0].message  # type:ignore
-    else:
+    fallback_reason: Optional[str] = None
+    try:
+        if tools:
+            response: Optional[ModelResponse] = llm.completion(
+                messages=conversation_history + [instructions_message],
+                tools=tools,
+                tool_choice="auto",
```
Avi-Robusta, line 264: this should be tool_choice="none"
we don't want tool calling here correct?

moshemorad: Correct that we don't want tool calls here — but `tool_choice="none"` can't be enforced on the platform this PR targets. Bedrock Converse's `toolChoice` only supports `auto` / `any` / `tool` (specific); there is no `none`. litellm's Converse transformation explicitly raises on it (`litellm/llms/bedrock/chat/converse_transformation.py`, `map_tool_choice_values`: *"Bedrock doesn't support tool_choice=none"*) unless `drop_params=True`, in which case the param is silently **dropped** — landing back on default-auto behavior anyway. So on Bedrock, `"none"` buys nothing, and on quirky gateways it adds a new 400 risk that would needlessly demote us to the flattened fallback.

`"auto"` is the one value every endpoint accepts verbatim, and the no-tool-call intent is enforced by two guards instead:
1. The compaction prompt opens with an explicit "do not use or call any tools" instruction.
2. If the model calls a tool anyway, we detect it (`fallback_reason: "model responded with a tool call instead of a summary"`) and retry once via the flattened no-tools fallback.

Verified live against Bedrock: the primary path completes with `fallback_used: false`.

---

holmes/core/truncation/compaction.py
```diff
@@ -127,47 +251,86 @@ def compact_conversation_history(
...
+            fallback_reason = "no message in summarization response"
+        elif getattr(response_message, "tool_calls", None):
+            fallback_reason = "model responded with a tool call instead of a summary"
+        elif not _extract_text_content(response_message).strip():
+            fallback_reason = "summarization response contains no text"
+    except Exception as e:
+        fallback_reason = f"summarization request failed: {e}"
+
+    if fallback_reason:
+        # Compatibility fallback: some gateways mis-translate tool blocks / tools
```
Avi-Robusta, line 283: I would protect fallback compaction in a try catch too to prevent unexpected behavior

moshemorad: Done in 8efc1d05 — the fallback request is now wrapped in try/except. On failure it records both errors in `fallback_reason` (visible in the compaction event metadata) and degrades through the existing empty-summary path, returning the original history unchanged instead of aborting the turn. Added a unit test for the double-failure case (`test_compaction_returns_original_history_when_fallback_also_fails`).

---

holmes/core/truncation/compaction.py
```diff
@@ -86,22 +101,131 @@ def _strip_images_for_compaction(messages: list[dict]) -> list[dict]:
...
+    if isinstance(content, list):
+        return "\n".join(
+            block.get("text", "")
+            for block in content
+            if isinstance(block, dict) and block.get("type") == "text"
+        )
+    return ""
+
+
 def compact_conversation_history(
```
Avi-Robusta, line 196: I ran a test with a small context window which would cause compaction at like 44k tokens and would compact to like 43k tokens at every call which wastes more tokens here then it would save, maybe have a minimum token requirement for compaction i.e. more then 50k/60k tokens, could be in separate ticket

moshemorad: Agreed this is a real pathology on small context windows: the preserved parts (system prompt + last user prompt + the summary itself) dominate the compacted history, so compression approaches zero and compaction re-fires every turn, each run costing a full-history summarization call. The existing guard only rejects results that don't reduce tokens at all — 44k→43k passes it, and the spend has already happened by then.

Agree it belongs in a separate ticket: the trigger/threshold logic (`compact_if_necessary`) is unchanged by this PR — this behavior exists on master too, and this PR only fixes the request/output *shape* of the summarization call. For the follow-up, two complementary guards worth considering:

1. **Absolute floor** — skip compaction below N total tokens (your 50–60k suggestion), since small histories rarely compress enough to pay for the call.
2. **Minimum expected-savings check** — before compacting, estimate the preserved floor (system + last user + typical summary size); if projected savings are marginal, skip and let the hard-limit error message handle it. That also protects large-window setups from repeated near-no-op compactions, which an absolute floor alone wouldn't.

---

holmes/core/truncation/input_context_window_limiter.py
```diff
@@ -124,13 +120,22 @@ def compact_if_necessary(
...
                 "max_context_size": max_context_size,
                 "threshold_pct": get_context_window_compaction_threshold_pct(),
+                # Whether the cache-friendly primary summarization request was
+                # abandoned for the flattened/no-tools fallback (which cannot
+                # reuse the agentic prompt cache). Key for diagnosing zero
+                # cached_tokens on the compaction call from event data alone.
+                "fallback_used": compaction_result.fallback_used,
             }
+            if compaction_result.fallback_reason:
+                compaction_stats["fallback_reason"] = compaction_result.fallback_reason
```
Avi-Robusta, line 130: the fallback_reason could contain URLs, request IDs, or auth hints that end up user-visible
now sure how this is used but just making you aware

holmes/core/truncation/compaction.py
moshemorad: (replying to the fallback_reason visibility note) Fair flag. Two mitigating facts:

1. **This isn't a new exposure class** — Holmes already sends raw exception text to the client on every error path: `stream.py:116` (`create_sse_error_message(description=str(e))`), `tool_calling_llm.py:813` (`Tool call failed: {e}`), and `tool_calling_llm.py:1267` (`{type(e).__name__}: {e}`). `fallback_reason` follows the same convention, and its audience is the same operator who configured the LLM endpoint (the metadata rides the `conversation_history_compacted` event).
2. The full detail is also in server logs (`logging.warning` at the fallback site), so trimming the event copy wouldn't lose diagnosability.

If we want to harden this, the right scope is all of the error paths above at once (e.g. a shared sanitizer that keeps exception type + provider error code and strips URLs/request IDs) — happy to do that as a separate ticket alongside the compaction-floor one.

---

holmes/core/truncation/input_context_window_limiter.py
moshemorad: Fair flag — answered in detail [in the thread above](https://github.com/HolmesGPT/holmesgpt/pull/2214#discussion_r3713267792) (posted there by mistake). Short version: it's not a new exposure class — Holmes already streams raw exception text to the client on every error path (`stream.py:116`, `tool_calling_llm.py:813`, `tool_calling_llm.py:1267`), and the audience of this event metadata is the operator who configured the endpoint. If we want to sanitize, the right scope is all those paths at once via a shared sanitizer — separate ticket.

---

holmes/core/truncation/compaction.py
```diff
@@ -127,47 +251,86 @@ def compact_conversation_history(
...
+    if fallback_reason:
+        # Compatibility fallback: some gateways mis-translate tool blocks / tools
+        # (ROB-424 was reported on Kong AI Gateway and vanilla LiteLLM proxies).
+        # Flattening tool messages to text and sending no tools is accepted by
+        # every OpenAI-compatible endpoint.
+        logging.warning(
+            f"Compaction: primary summarization attempt unusable ({fallback_reason}); "
+            "retrying with tool messages flattened to text and no tools attached"
+        )
+        flattened_history, _ = strip_system_prompt(conversation_history)
```
Avi-Robusta, line 291: just pointing out here we strip system prompt but in the new compaction above we do not, is this on purpose? why do we only do it here?

moshemorad: On purpose, yes — the two paths optimize for different things:

- **Primary keeps the system prompt** because its goal is a request whose prefix is byte-identical to the previous agentic call (system + tools + native history). That's what guarantees `toolConfig` is defined on Converse-translating gateways *and* lets this request — the largest one Holmes makes — reuse the agentic prompt-cache prefix. Dropping the system message there would break the cache prefix at position 0.
- **The fallback strips it** because by the time it runs, the native shape has already failed and the history is being rewritten anyway (tool blocks flattened) — so there's no cache prefix left to preserve. What's left to optimize is compatibility and summary quality: the agentic system prompt is a large investigator-persona/tool-usage instruction set that competes with the compaction instructions on a request that deliberately attaches no tools (and it may be exactly what induced the stray tool call that triggered the fallback). Stripping it also saves tokens on the retry.

Either way nothing is lost: the system prompt is captured up front (`system_prompt_message`) and re-attached verbatim as message 0 of the compacted history — the summarizer never needs to see or summarize it.

---

Avi-Robusta (changes requested):
looks good, left some comments

holmes/core/truncation/compaction.py
Avi-Robusta: 15k ish tokens are wasted each compaction if you are adding it for no benifit and is the system prompt added afterwards wouldnt this be a moot point?

moshemorad: It's the opposite of a waste — keeping it is what makes the request cheap. The re-attachment afterward is about the *output* shape; keeping it in the request is about the *input* price:

- With the system prompt kept, the summarization request's prefix (system + tools + history) is byte-identical to the previous agentic call, so nearly the whole input is billed as **prompt-cache reads** at ~10% of the normal input price. The "15k wasted" system-prompt tokens actually cost ~1.5k full-price-equivalent.
- Strip the system message and the cache prefix breaks at **position 0** — providers match cache prefixes from the first block, so *everything after it* (the entire history, easily 100k+ tokens at compaction time) is re-billed as fresh input at full price. You'd save ~1.5k-equivalent and pay ~90k+-equivalent extra.

We measured this live on Bedrock during the PR work: the prefix-preserving request ran at roughly half the cost of a prefix-broken one on the same history (~40k cached-read tokens vs. zero). Same reason the instructions are *appended* as the final user message rather than prepended or merged into the system prompt.

---

holmes/core/truncation/compaction.py
Avi-Robusta: not sure if risking running tools on all platforms is the correct tradeoff

moshemorad: To be precise about the risk: **no tool ever runs from the compaction call, on any platform.** The compaction path never executes tool calls — if the response contains one, it's discarded, treated as an unusable summary (`fallback_reason: "model responded with a tool call instead of a summary"`), and we retry via the flattened no-tools fallback. The worst case of `auto` is one wasted summarization request, not tool execution.

And `"none"` specifically breaks the customer topology this PR fixes: Holmes usually reaches Bedrock **through an OpenAI-compatible proxy** (LiteLLM proxy / Kong). Client-side litellm sees an OpenAI route and passes `tool_choice="none"` through verbatim — our `drop_params=True` can't help — and then the *proxy's* translation layer hits the same "Bedrock doesn't support tool_choice=none" and 400s (unless that operator happened to enable drop_params server-side). We'd forfeit the primary path exactly where ROB-424 lives.

So the trade-off is: `auto` = rare wasted request, detected and recovered; `none` = deterministic 400 on proxy→Bedrock setups. If we ever see the stray-tool-call fallback firing frequently in the `fallback_reason` telemetry this PR adds, we can revisit (e.g. try `none` first and fall back to `auto` on 400) — but that's an extra round-trip for a problem we haven't observed in practice.

---

moshemorad:
/eval
tags: compaction

Avi-Robusta (approved):
I would add a pytest that verifies that the compaction second flow is called if the first one throws, not a blocker so Im approving

moshemorad:
Thanks for the approval! Re the pytest note — that coverage already exists in `tests/core/truncation/test_compaction.py`:

- `test_compaction_falls_back_when_primary_request_fails` — primary raises → asserts the second (flattened, no-tools) flow is called and its summary is used
- `test_compaction_falls_back_when_model_calls_a_tool` — same for the tool-call trigger
- `test_compaction_returns_original_history_when_fallback_also_fails` — both flows fail → original history returned unchanged

---

PR #2237, opened 2026-06-26
feat: enable HTTPS support for HolmesGPT API with in-app TLS

theTibi:
The API server only listened on plain HTTP. Add optional in-app TLS so it
can terminate HTTPS itself in any environment (Docker, bare process, or
Kubernetes) without an ingress or reverse proxy.

TLS is driven purely by config: when HOLMES_SSL_CERTFILE and
HOLMES_SSL_KEYFILE are both set the server serves HTTPS, otherwise it
serves HTTP exactly as before. Misconfiguration fails fast — one of the
pair missing, a referenced file missing, or mTLS/password settings given
without a cert+key — rather than silently downgrading to plaintext.

- env_vars: add HOLMES_SSL_CERTFILE, HOLMES_SSL_KEYFILE,
  HOLMES_SSL_KEYFILE_PASSWORD, and HOLMES_SSL_CA_CERTS (the last enables
  mTLS / client-certificate verification).
- server: build_ssl_kwargs() resolves and validates the uvicorn ssl_*
  args up front (before the slow pre-start sync) and logs the scheme.
- helm: opt-in `tls` block, default off and fully backward compatible.
  When enabled, mount a user-provided TLS secret, inject the env, set the
  Service appProtocol to https, and serve HTTPS from the pod (no ingress).
  Probes use the HTTPS scheme; under mTLS they fall back to tcpSocket
  because the kubelet cannot present a client cert. The secret is
  projected with explicit items so the key password is never written to
  disk, and rendering fails if tls.enabled is set without secretName.
- tests: unit-test build_ssl_kwargs() across the HTTP, HTTPS, mTLS, and
  fail-fast cases.
- docs: document HOLMES_SSL_* and the Kubernetes / Docker HTTPS setup.

## Summary by CodeRabbit

* **New Features**
  * Added support for serving the API over HTTPS using `HOLMES_SSL_*`, including optional mutual TLS and private key passphrase support.
  * Helm deployments can now enable HTTPS via configuration, updating the Service protocol and health checks automatically.

* **Documentation**
  * Expanded Docker and Kubernetes installation guides with TLS enablement steps, secret setup, and HTTPS `/healthz` verification.
  * Updated environment variable reference with required/mutually required TLS settings and noted startup-only certificate loading (restart needed after rotation).

* **Tests**
  * Added unit tests validating TLS/mTLS argument construction and fast-failure on misconfiguration.


moshemorad:
@claude review

moshemorad:
Hi @theTibi Can you please go over the code rabbit comments?
Thanks!

theTibi:
@moshemorad sorry for delay. it is done now. If coderabbit will have other comments I will handle those as well.

moshemorad (approved):

PR #1789, opened 2026-03-15
Handle null analysis result in chat endpoint response

aantn:
## Summary
Fixed a potential issue where the chat endpoint could return a null `analysis` field in the response when the LLM call result is null or undefined.

## Changes
- Modified the `ChatResponse` instantiation in the `/chat` endpoint to provide an empty string as a fallback when `llm_call.result` is null or falsy
- Changed `analysis=llm_call.result` to `analysis=llm_call.result or ""`

## Details
This change ensures that the `analysis` field in the `ChatResponse` will always contain a string value (either the actual LLM result or an empty string), preventing potential null/undefined issues in downstream consumers of the API response. This is a defensive programming practice that improves API reliability and consistency.

https://claude.ai/code/session_01AAJrVMD3b6jKUqexNT4DHM

## Release Notes

* **Bug Fixes**
  * Fixed an issue in the chat API where the analysis field could return null values. The analysis field now consistently returns a string response, improving API reliability and preventing potential errors.

Issue #1676, opened 2026-03-05
/api/chat returns 500 when model's final response has empty content (ChatResponse.analysis is None)


aaronthebaron:
### What happened?

When using the /api/chat endpoint with conversation history (follow-up questions), the LLM occasionally returns a final response with content: null and no tool calls. This causes a Pydantic validation error in server.py line 518 because ChatResponse.analysis is typed as str (required) but receives None.

Error below:

  Error in /api/chat: 1 validation error for ChatResponse  analysis
    Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]

  Traceback (most recent call last):
    File "/app/server.py", line 518, in chat
      return ChatResponse(
  pydantic_core._pydantic_core.ValidationError: 1 validation error for
  ChatResponse
  analysis
    Input should be a valid string [type=string_type, input_value=None,
  input_type=NoneType]

### What did you expect to happen?

Pydantic validation doesn't blow up.

### How can we reproduce it (as minimally and precisely as possible)?

1.  Send an initial question to /api/chat (succeeds)
2. Send a follow-up question in the same conversation using the returned conversation_history
3. If the model's final turn produces content: null (observed with  claude-sonnet-4-6 via litellm), the server returns a 500

The model does complete its investigation successfully (tool calls execute, results are returned), but its final summarization turn occasionally emits an empty response (2 output tokens, content: null, tool_calls: null).

aantn:
Hi @aaronthebaron are you able to reproduce on master? 

I don't understand how this part can happen:

> the LLM occasionally returns a final response with content: null and no tool calls

aantn:
There's a fix here you can try: https://github.com/HolmesGPT/holmesgpt/pull/1789/changes
But I'd love to get more information if possible to understand more. The real problem is that LLM is returning an empty response.

aaronthebaron:
Thank you @aantn for your response. I am as confused as you are, but will try master and your fix when I get a chance this week.

Here's what I noticed in my testing, for simple initial questions like, "What is the state of the ArgoCD application [app]" I would get a response each time with no issues. Only on continuing the conversation or a conversation which required more tokens like "ArgoCD application is having trouble, lool at the application and tell me how to fix it." would this error occur.

I have a middleware container slack bot that is facilitating this convo from slack so it's possible there's an issue there but repeated testing suggested the fix you linked and that the issue was in holmesgpt.

Let me do some more testing with your fix and against master and see what I can see.

[...]

PR #2241, opened 2026-06-28
fix: preserve last non-empty answer when closing turn is empty (#2185)

### LINKED ISSUE #2185: Bug: Final answer is lost when the model's closing turn is empty (analysis comes back empty)
### What happened?

The agentic loop in `ToolCallingLLM.call_stream` returns `terminal_data["content"]` — the content of the LLM's **final** turn — as the answer (`LLMResult.result`, and therefore `ChatResponse.analysis` for `/api/chat`).

Some providers (observed with `claude-sonnet-4-6` via litellm) emit their answer in a turn that **also** carries a trailing tool call (e.g. `TodoWrite`), and then close the loop with an empty turn (`content=None` / `""`, no tool calls). The loop terminates on that empty turn, so the empty content is returned and **the real answer the model already produced is dropped**.

Concretely, in the logs the model clearly synthesises an answer:

```
AI: ## Summary
... <full answer> ...
```

…but `/api/chat` returns `analysis = ""`. The investigation did run (tool calls executed, results returned); only the final summarisation turn came back empty.

The answer the model produced should be returned, not silently discarded. An empty closing turn after a substantive answer should not erase that answer.

1. Ask `/api/chat` a question that triggers a multi-step investigation (tool calls), with `claude-sonnet` via litellm.
2. Occasionally the model emits its final summary in a turn that also requests one last tool call (e.g. `TodoWrite`), then produces an empty closing turn (`content=null`, no tool calls).
3. `call_stream` yields `ANSWER_END` with `content = response_message.content` from that empty turn → `result` / `analysis` is empty, even though the summary was generated one turn earlier.

Related: this is the same family as #1676. #1789 stopped the empty content from **crashing** `/api/chat` (analysis is now coerced to `""` instead of raising a 500), but the **answer is still lost** — clients receive an empty response and may pointlessly re-run the whole investigation (extra latency + cost).

Proposed fix in #2184: track the last non-empty assistant text across iterations and fall back to it when the terminating turn has no text.

1HazyOne707:
## Problem

`call_stream` returns `terminal_data["content"]` — the content of the LLM's **final** turn — as the answer. Some providers (observed with `claude-sonnet-4-6` via litellm) emit the real answer in a turn that also carries a trailing tool call, then close the loop with an empty turn (`content=None`, no tool calls). The loop terminates on that empty turn, so the empty content is returned and **the real answer is silently dropped**.

This is the same family as #1676. #1789 stopped the empty content from crashing `/api/chat` (analysis is now `""` instead of a 500), but the answer is still lost.

## Fix

Track `_last_non_empty_content` across `AI_MESSAGE` events in `call_stream`. When `ANSWER_END` fires with empty content, fall back to the last non-empty assistant text. When no prior non-empty content exists, behaviour is unchanged.

## Tests

Three new tests in `TestEmptyClosingTurnFallback`:
- Empty closing turn falls back to last non-empty `AI_MESSAGE` content
- Non-empty closing turn is used directly (no fallback)
- Truly empty response (no prior content) returns `None` as before

Closes #2185

## Summary by CodeRabbit

* **Bug Fixes**
  * Streamed responses now preserve the last meaningful assistant reply when a final closing turn is empty.
  * If the final turn includes text, that text is still shown as expected.
  * If no prior reply text exists, the end result remains empty.
* **Tests**
  * Added coverage for streamed tool-use conversations to verify the updated end-of-response behavior, including a fallback for empty closing turns.

### DIFF
--- a/holmes/core/tool_calling_llm.py
+++ b/holmes/core/tool_calling_llm.py
@@ -1056,6 +1056,7 @@ class ToolCallingLLM:
 
         self._request_context = request_context
         all_tool_calls: list[dict] = []
+        _last_non_empty_content: Optional[str] = None  # fallback for empty closing turns (#2185)
 
         # Process tool decisions if provided (approval resume)
         if msgs and tool_decisions:
@@ -1259,10 +1260,17 @@ class ToolCallingLLM:
                         metadata["finish_reason"] = fr
                 except (AttributeError, IndexError, TypeError):
                     pass
+                # Fall back to the last non-empty assistant text if the
+                # terminating turn is empty. Some providers (e.g. claude-sonnet-4-6
+                # via litellm) emit the real answer in a turn that also carries a
+                # trailing tool call, then close the loop with an empty turn.
+                # Without this fallback that empty content silently replaces the
+                # answer the model already produced. See issue #2185.
+                final_content = response_message.content or _last_non_empty_content
                 yield StreamMessage(
                     event=StreamEvents.ANSWER_END,
                     data={
-                        "content": response_message.content,
+                        "content": final_content,
                         "messages": messages,
                         "metadata": metadata,
                         "tool_calls": all_tool_calls,
@@ -1276,6 +1284,10 @@ class ToolCallingLLM:
             reasoning = getattr(response_message, "reasoning_content", None)
             message = response_message.content
             if reasoning or message:
+                # Track last non-empty assistant text for fallback on empty
+                # closing turns (see issue #2185).
+                if message and message.strip():
+                    _last_non_empty_content = message
                 yield StreamMessage(
                     event=StreamEvents.AI_MESSAGE,
                     data={
--- a/tests/test_tool_calling_llm.py
+++ b/tests/test_tool_calling_llm.py
@@ -1702,3 +1702,81 @@ class TestFrontendNoopToolFlow:
         tool_names = [t["function"]["name"] for t in tools_sent]
         assert "kubectl_get" in tool_names, "Backend tool should be included"
         assert "navigate_to_page" in tool_names, "Noop tool should be included"
+
+
+# ---------------------------------------------------------------------------
+# Issue #2185: empty closing turn should not erase the real answer
+# ---------------------------------------------------------------------------
+
+class TestEmptyClosingTurnFallback:
+    """When the model emits its answer in a turn that also carries a trailing
+    tool call, then closes the loop with an empty turn, the real answer must
+    not be lost (issue #2185)."""
+
+    @patch("holmes.core.tool_calling_llm.compact_if_necessary")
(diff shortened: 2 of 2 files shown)


1HazyOne707:
Refactored to focus solely on the empty-closing-turn fix per CodeRabbit review. The two-phase HealthCheck execution logic has been split into a separate PR: #2243. This PR now only touches `holmes/core/tool_calling_llm.py` and `tests/test_tool_calling_llm.py` to fix the issue where empty closing turns were silently dropping the answer.

mdecalf:
Just flagging that #2184 already fixes this exact issue (opened 2026-06-11, before this one) — same root cause, same approach (tracking the last non-empty assistant content across the tool loop and falling back to it when the closing turn is empty). It's tested (unit tests for the None/blank/no-prior-content cases) and ready to merge.

Might be worth coordinating to avoid two PRs fixing the same bug — happy to have your tests/scenarios folded into #2184 if there's anything additional there, or vice versa, whichever the maintainers prefer.

1HazyOne707:
Thanks for the detailed review and for putting this fix together. I compared the changes in #2184 and #2241 and wanted to share a few observations.
The simplified implementation in #2241 looks cleaner and easier to maintain, but I noticed a subtle behavior difference from the original fix.
The current fallback:

```python
final_content = response_message.content or _last_non_empty_content
```

handles `None` and `""`, but it does not handle whitespace-only responses.
For example:

```python
response_message.content = "   "
```

is truthy in Python, so the fallback returns the whitespace string instead of recovering the previous substantive assistant response.
The original #2184 implementation handled this by checking whether the content was actually substantive:

```python
if final_content and final_content.strip():
```

which preserves the intended behavior that only meaningful assistant text should become the final answer.
I think the implementation direction in #2241 is good, and the smaller patch is easier to follow. My suggestion would be either:

```python
final_content = (
    response_message.content
    if response_message.content and response_message.content.strip()
    else _last_non_empty_content
)
```

or adding a regression test covering a blank/whitespace final turn.
The additional cases I think are worth protecting:

1. Final turn with `content=None` → recover previous answer ✅
2. Final turn with whitespace-only content → recover previous answer ⚠️
3. No prior substantive assistant text → remain empty ✅

The reason I think the whitespace case is important is that provider behavior can vary (especially through adapters like LiteLLM), and the invariant we probably want is:
ANSWER_END should contain the last substantive assistant message, not simply the last non-null content field.
Overall, this is a good fix. I think adding the whitespace regression guard would make the behavior more robust and prevent a similar class of answer-loss bugs in the future.

1HazyOne707:
Closing in favor of #2184, which was opened first and handles a couple of cases mine doesn't (whitespace-only content, and the trace-logging path). Thanks for the review either way.

PR #2246, opened 2026-06-29
Bound conversation/tool-call claiming to worker capacity (ROB-388)

naomi-robusta:
## Summary

This PR deprecates the intermediate `queued` conversation lifecycle status by having the claim RPC land conversations directly in `running`. The worker now claims only as many conversations as it has free pool slots, leaving surplus conversations `pending` in the database for the next poll or another Holmes instance to claim. This enables better cross-instance load balancing and simplifies the worker's internal state management.

## Key Changes

**ConversationWorker (`holmes/core/conversations_worker/worker.py`)**:
- Removed `_queued_tasks` deque and `_queued_lock` — conversations no longer queue locally
- Replaced `claim_conversations()` with `claim_n_pending_conversations(limit)` that respects free pool capacity
- Added `_free_claim_slots()` to calculate available executor capacity
- Refactored `_dispatch_queued()` into simpler `_dispatch(task)` that submits directly without DB status transition (claim RPC already set `running`)
- Updated `_try_claim_and_dispatch()` to claim only as many conversations as there are free slots
- Changed `_process_conversation_safe()` to signal the claim loop via `_notify_event` when a slot frees up, instead of calling `_dispatch_queued()`
- Updated lifecycle documentation: `pending → running (claimed + processing) → completed/failed`

**SupabaseDal (`holmes/core/supabase_dal.py`)**:
- Added `claim_n_pending_conversations(holmes_id, limit)` RPC that claims up to `limit` pending conversations and lands them directly in `running` status
- Added `claim_n_pending_tool_calls(holmes_id, limit)` RPC for the tool-call worker (same pattern)
- Both methods skip the RPC entirely if `limit <= 0` (no free capacity)

**ToolCallWorker (`holmes/core/conversations_worker/tool_call_worker.py`)**:
- Applied the same bounded-claiming pattern: track `_active_count` and only claim as many tool calls as there are free pool slots
- Updated `_try_claim_and_dispatch()` to use `claim_n_pending_tool_calls(limit)` instead of `claim_tool_calls()`
- Modified `_execute_safe()` to decrement `_active_count` in a `finally` block and signal the claim loop when done
- Updated docstring to reflect the new RPC name

**ConversationStatus enum (`holmes/core/conversations_worker/models.py`)**:
- Marked `QUEUED` as deprecated in docstring (no longer produced by current worker)
- Kept `QUEUED` in `updatable_values()` for backwards compatibility with in-flight rows and mixed-version rollouts

**Tests**:
- Updated `test_worker_lifecycle.py`: removed `_queued_tasks` and `_queued_lock` setup; rewrote claim/dispatch tests to verify the new bounded-claiming behavior
- Renamed `test_try_claim_and_dispatch_claims_all_and_queues()` → `test_try_claim_and_dispatch_claims_only_free_slots()`
- Renamed `test_dispatch_queued_*` tests → `test_dispatch_*` (simpler, no status transition)
- Added `test_try_claim_and_dispatch_passes_remaining_capacity_as_limit()` to verify capacity calculation
- Updated `test_process_conversation_safe_*` to verify `_notify_event` signaling instead of `_dispatch_queued()` calls
- Updated integration test `test_queued_state_observed()` → `test_pending_backlog_observed()` to verify surplus conversations stay `pending` while pool is at capacity
- Added DAL contract tests for `claim_n_pending_conversations()` and `claim_n_pending_tool_calls()`
- Added tool-call worker tests for bounded claiming and slot management

## Implementation Details

- **Capacity-aware claiming**: The worker calculates free slots as `MAX_CONCURRENT - len(active_conversation_ids)` and passes this to the claim RPC, ensuring it never holds more than `MAX_CONCURRENT` running conversations
- **No

https://claude.ai/code/session_01M5RaU2S5paAqbNW6GUF3K4

## Summary by CodeRabbit

* **New Features**
  * Worker pools now claim backlog in capacity-bounded batches and immediately dispatch claimed work for both conversations and remote tool calls.
* **Bug Fixes**
  * Improved capacity/slot accounting to prevent leaks across execution, dispatch, and shutdown race conditions.
* **Documentation**
  * Clarified the lifecycle around deprecated `queued/QUEUED`; the worker now transitions directly to `running` and does not produce `queued`.
* **Tests**
  * Updated and added stress and integration coverage to validate `pending → running` behavior, ensure `queued` never appears, and verify concurrency/burst correctness.


holmes/core/conversations_worker/worker.py
```diff
@@ -460,67 +462,30 @@ def _try_claim_and_dispatch(self) -> None:
...
-                        task.conversation_id,
-                        exc_info=True,
-                    )
-                    with self._queued_lock:
-                        self._queued_tasks.appendleft(task)
-                    break
-
-                with self._active_lock:
-                    self._active_conversation_ids.add(task.conversation_id)
+            if not self._running or self._executor is None:
```
moshemorad, line 474: Thinking here loudly, if we claimed and than the process is exists we might want to try update the status so the new instance will grab them, no?

It isn't a new gap introduced so maybe worth to open linear ticket for it?

moshemorad (approved):

PR #2253, opened 2026-07-01
feat(helm): add extraEnvVarsSecrets + envRef: sugar for modelList

mainred:
## Summary

- Wiring one API key into `modelList` currently takes ~10 lines: an `additionalEnvVars` entry per key (with nested `valueFrom.secretKeyRef`) **plus** a matching `api_key: "{{ env.KEY }}"` in the model config. Adding a second provider doubles the boilerplate.
- This PR adds two chart-level features that let users express the same setup in about half the lines and never repeat a key name — no Python code changes.
- New top-level **`extraEnvVarsSecrets`** value: a list of Kubernetes Secret names auto-mounted onto the Holmes pod via `envFrom.secretRef`, exposing every key as an env var. Splitting keys across multiple Secrets (e.g. one per provider) is a one-line addition.
- New **`envRef:VAR`** prefix accepted on any string value inside `modelList`: rewritten at chart-render time to `"{{ env.VAR }}"`, which the existing runtime `get_env_replacement` in `holmes/utils/env.py` resolves as before.

### Before → After

**Before** (10+ lines, doubles per key):
```yaml
additionalEnvVars:
  - name: OPENAI_API_KEY
    valueFrom:
      secretKeyRef:
        name: holmes-secrets
        key: openai-api-key
modelList:
  gpt-4.1:
    model: openai/gpt-4.1
```

**After** (5 lines, adds 1 line per additional key):
```yaml
extraEnvVarsSecrets:
  - holmes-secrets
modelList:
  gpt-4.1:
    api_key: envRef:OPENAI_API_KEY
    model: openai/gpt-4.1
```

**Multiple Secrets** (split keys by provider / rotation cadence):
```yaml
extraEnvVarsSecrets:
  - holmes-llm-keys      # OPENAI_API_KEY, ANTHROPIC_API_KEY
  - holmes-aws-keys      # AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY
modelList:
  gpt-4.1:
    api_key: envRef:OPENAI_API_KEY
    model: openai/gpt-4.1
  bedrock-sonnet:
    aws_access_key_id: envRef:AWS_ACCESS_KEY_ID
    aws_secret_access_key: envRef:AWS_SECRET_ACCESS_KEY
    aws_region_name: us-east-1
    model: bedrock/us.anthropic.claude-sonnet-4-5-20250929-v1:0
```

### Files changed
- `helm/holmes/values.yaml` — new `extraEnvVarsSecrets: []` field with usage comment; updated `additional_env_froms` comment to reference the new field.
- `helm/holmes/templates/holmes.yaml` — extended `envFrom` block: iterates `extraEnvVarsSecrets` emitting one `- secretRef: {name}` per entry, then appends any `additional_env_froms`.
- `helm/holmes/templates/toolset-config.yaml` — one-line `regexReplaceAll` that rewrites `envRef:VAR` → `"{{ env.VAR }}"` in the rendered `model_list.yaml`.
- `helm/holmes/templates/validate.yaml` — chart-render-time guard that fails `helm install` if any `envRef:VAR` name isn't a POSIX env-var identifier (Kubernetes silently drops non-conforming keys from `envFrom.secretRef` mounts).
- `docs/reference/helm-configuration.md` — new "Simplified API Key Configuration" section and a second variant of the multi-provider example using the sugar.

### Design choices

- **`extraEnvVarsSecrets` is a list.** Users often want to split API keys across multiple Secrets (per-provider rotation, per-team ownership, existing IAC pattern). One Secret per entry, all merged into the pod's `envFrom` — same env-var namespace, so same-name keys shadow each other in list order.
- **`envRef:` prefix — not `secretRef:` and not `env:`.** `secretRef` is Kubernetes-native terminology for `envFrom.secretRef`; reusing it inside a value string would suggest the sugar is talking about Kubernetes Secret keys directly, when in fact it's just referencing an env var by name. A bare `env:` prefix collides visually with the pod's own `env:` list in `additionalEnvVars` and with the runtime templating `{{ env.VAR }}` — reading a values.yaml, it isn't obvious that `api_key: env:OPENAI_API_KEY` is a chart-render-time rewrite. `envRef:` mirrors the existing Kubernetes `secretRef` / `configMapRef` / `fieldRef` convention: the `Ref` suffix reads unambiguously as "a reference to an env var" and stays visually distinct from both the pod's `env:` block and the runtime `{{ env.X }}` target.
- **Hard rename, no shim.** The prior `apiKeysSecret` / `secretRef:VAR` / `env:VAR` designs were on an unmerged branch — no shipped users, so no back-compat plumbing. Anyone upgrading a checkout of this branch will render an empty `envFrom` block and a literal `secretRef:...` / `env:...` in their model config, which fails loud at auth time.

## Test plan

Verified locally with `helm lint` (clean) and `helm template` against seven scenarios:

- [x] **Multi-secret list** — `extraEnvVarsSecrets: [holmes-llm-keys, holmes-aws-keys, holmes-slack-token]` + 4 `envRef:` values across 3 models → three `- secretRef:` blocks emitted in order; all 4 `envRef:` values rewritten to `"{{ env.VAR }}"`; non-`envRef:` value (`aws_region_name: us-east-1`) left untouched.
- [x] **Empty list omits envFrom** — `extraEnvVarsSecrets: []` and no `additional_env_froms` → `envFrom` block completely absent from the Deployment (0 occurrences), matching pre-change output byte-for-byte.
- [x] **`additional_env_froms` alone** — mixed `secretRef` + `configMapRef` entries with empty `extraEnvVarsSecrets` → `envFrom` block appears with both entries; composition path unaffected.
- [x] **Validator rejects hyphenated names** — `envRef:AZURE-API-KEY` → `helm template` aborts with `modelList.gpt-4.1.api_key: invalid envRef: env var name "AZURE-API-KEY". Names must match [A-Za-z_][A-Za-z0-9_]* ...` (Kubernetes silently drops non-conforming keys from `envFrom.secretRef` mounts, so failing at render time is correct).
- [x] **Regex boundary holds** — `envRef:OPENAI_API_KEY` in value position → rewritten; `description: "use envRef:SOMETHING_ELSE syntax as noted in docs"` → left untouched. The `(?m): envRef:...$` anchor is still tight against false positives.
- [x] **Legacy fields silently ignored** — someone still on `apiKeysSecret: foo` + `secretRef:BAR` / `env:BAR` → 0 `envFrom` blocks emitted, literal prefix passes through to the rendered `model_list.yaml`. The auth failure at runtime prompts migration — intentional consequence of the hard rename.
- [x] **`helm lint helm/holmes`** — clean.

## History: earlier iterations and CodeRabbit review fixes

**Earlier designs (all on this branch, all superseded):**

1. `apiKeysSecret: <string>` (single Secret) + `secretRef:VAR` prefix — renamed because `apiKeysSecret` didn't support multiple Secrets and `secretRef:` collided with Kubernetes' native `envFrom.secretRef` terminology.
2. `extraEnvVarsSecrets: [<list>]` + `env:VAR` prefix — renamed because a bare `env:` collided visually with both the pod's `env:` block and the runtime `{{ env.VAR }}` target.
3. **Current**: `extraEnvVarsSecrets: [<list>]` + `envRef:VAR` — `envRef` mirrors the Kubernetes `*Ref` convention and stays unambiguously about env-var references.

**CodeRabbit review fixes (commit `2bbf72c0d`) — still applicable:**

1. **Docs**: Multi-provider example uses the latest Claude 4.5 model IDs (`anthropic/claude-sonnet-4-5-20250929` and `bedrock/us.anthropic.claude-sonnet-4-5-20250929-v1:0`).
2. **Template**: The value-position anchor `(?m): envRef:([A-Za-z_][A-Za-z0-9_]*)$` (carried over from the tightened `secretRef:` / `env:` versions) — Helm's `toYaml` always emits scalar values as unquoted plain scalars in `key: value` form, so `(?m)$` covers every real-world usage while preventing false positives inside `notes:` or `description:` strings.

## Summary by CodeRabbit

* **New Features**
  * Added `extraEnvVarsSecrets` to mount one or more Kubernetes Secrets into the Holmes pod via `envFrom`.
  * Added `env:` shorthand support in `modelList` to reference mounted env vars at chart-render time.
* **Documentation**
  * Updated the Helm configuration reference with setup instructions, env-var naming constraints, compatibility guidance, and updated examples.
* **Bug Fixes**
  * Improved `envFrom` handling so `extraEnvVarsSecrets` can be combined with existing additional env sources.
* **Tests / Validation**
  * Added Helm-time validation that `env:` keys use a safe env-var name format (invalid values fail rendering).


helm/holmes/templates/toolset-config.yaml
```diff
@@ -220,6 +220,15 @@ data:
...
     mcp_servers: {{ toYaml $mcpServers | nindent 6 }}
   model_list.yaml: |-
-    {{ .Values.modelList | toYaml | nindent 4 }}
+    {{- /*
+      Rewrite any string value in modelList that starts with `secretRef:VAR`
+      to the runtime template `"{{ env.VAR }}"`. The corresponding env var
+      must be present on the Holmes pod — set `apiKeysSecret` (or add
+      `additional_env_froms` / `additionalEnvVars`) so it is.
+    */ -}}
+    {{- $rendered := .Values.modelList | toYaml -}}
```
mainred, line 242: An alternative is we make the switch in the python code when rendering the modelList

mainred:
@moshemorad  could you please take another look?

moshemorad (approved):

Issue #2188, opened 2026-06-12
Operator bombs out when using vertex models complaining about INVALID_ARGUMENT


butangero:
### What happened?

[fix-vertex-checks.py.diff.txt](https://github.com/user-attachments/files/28891835/fix-vertex-checks.py.diff.txt)

We use vertes/gemini for our models, and until turning on the operator things worked fine. Once the operator was turned on, all healthchecks would bomb out with an error along the lines of:

`2026-06-12 16:45:03.434 ERROR LLM BadRequestError on model=vertex_ai/gemini-3.5-flash (streaming iteration 1): litellm.BadRequestError: Vertex_aiException BadRequestError - {
"error": {
"code": 400,
"message": "Request contains an invalid argument.",
"status": "INVALID_ARGUMENT"`

After looking into it, the problem seems to be in checks.py.  The CHECK_RESPONSE_FORMAT is defined as a raw OpenAI-specific dict:

CHECK_RESPONSE_FORMAT = {
    "type": "json_schema",
    "json_schema": {
        "name": "check_response",
        "strict": True,   # ← OpenAI-only field
        "schema": { ... }
    },
}

This dict is passed to litellm.completion(response_format=...) for every operator health check. litellm only auto-translates response_format to provider-specific formats (like Vertex AI's generationConfig.responseSchema) when it receives a Pydantic model. When given a raw dict, it passes it through as-is — Vertex AI sees the OpenAI format and returns INVALID_ARGUMENT.

This code path is only triggered by the operator (via /api/checks/execute). The regular ask CLI never uses response_format. Claude on Vertex works because Claude's request path uses a different schema format.

Fix: Replace the raw dict with the CheckResponse Pydantic model that already existed — litellm converts it correctly for each provider automatically.

I've attached a diff file with the changes that should fix the problem

### What did you expect to happen?

I expect vertex to work as a first-class citizen.

### How can we reproduce it (as minimally and precisely as possible)?

Use a gemini model with vertex and turn on the operator.

### Anything else we need to know?

Nope - you're doing great!

butangero:
This should've been addressed when this [PR](https://github.com/HolmesGPT/holmesgpt/pull/2288) was merged

PR #2267, opened 2026-07-06
Fix issue when using operator with vertex

butangero:
[Issue Addressed](https://github.com/HolmesGPT/holmesgpt/issues/2188)

Fix(checks): use Pydantic model as response_format to fix Vertex AI /Gemini compatibility

Passing a raw OpenAI-format dict with type: json_schema / strict: True as the response_format causes Vertex AI and Gemini to return INVALID_ARGUMENT, breaking the holmesgpt-operator when using those providers.

Replace CHECK_RESPONSE_FORMAT (a hardcoded dict) with the CheckResponse Pydantic model class directly. litellm translates a Pydantic model into the correct structured-output schema for each provider — OpenAI json_schema, Vertex AI responseSchema, etc. — whereas the raw dict was OpenAI-specific.

Tests:
- Add test_checks_response_format_is_pydantic_model to assert ai.call() receives CheckResponse (the Pydantic class) rather than a raw dict, catching any future regression on non-OpenAI providers.
- Fix test_log_buffering_filter flake under pytest-xdist: in some workers the root logger has no handlers, so the log filter was installed on nothing and the buffer stayed empty. Add a temporary NullHandler when needed so the filter has somewhere to sit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of AI-powered check execution by standardizing the structured response format used across providers.
  * Reduced the likelihood of check-run failures caused by mismatched response-format expectations.
  * Strengthened logging behavior during interactive usage to prevent missing-handler issues.
* **Tests**
  * Added coverage to ensure the CLI passes the expected structured response format during monitor-mode runs.
  * Updated interactive logging test setup to be resilient in parallelized environments.

### DIFF
--- a/holmes/checks/checks.py
+++ b/holmes/checks/checks.py
@@ -5,7 +5,7 @@ import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Type
 
 import requests  # type:ignore
 import yaml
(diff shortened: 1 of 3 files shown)


butangero:
this needs a little more work.

PR #2272, opened 2026-07-08
Additional configurations for Helm chart: `command` and probes

dabico:
Hi folks! I'm testing Holmes out with the [AI Assistant](https://github.com/headlamp-k8s/plugins/tree/b84332fea79153187dd218ebce874f39a6929a60/ai-assistant) plugin for Headlamp. The setup needed for this plugin is quite specific and involves the use of features like the experimental server:

> ### 3. Render and Patch the Helm Template
>
> Holmes requires enabling the AG-UI server for the AI Assistant to communicate with it. Render the template and update the container command:
>
> ```bash
> helm template holmesgpt robusta/holmes -f values.yaml > rendered.yaml
> ```
>
> In `rendered.yaml`, find the container command and change it to:
>
> ```yaml
> command: ['python3', '-u', '/app/experimental/ag-ui/server-agui.py']
> ```

It's currently not possible to tweak the container `command` in the chart. Furthermore, the experimental server doesn't provide the same endpoints that the liveness/readyness probes rely on, which leads to restart loops.

Right now, one would either have to do as the guide suggests or use techniques like applying [Kustomize patches via Helm hooks](https://austindewey.com/2020/07/27/patch-any-helm-chart-template-using-a-kustomize-post-renderer/). I'm not particularly fond of either approach. To that end, I've just introduced some values that allow you to override the `command` and disable the probes for the `Deployment`.

This way, consumers can:

- Customize whether they want to use the vanilla or experimental version.
- Tweak or even remove the probes based on the version they are using.

## Summary by CodeRabbit

* **New Features**
  * Added configurable container startup commands and arguments through Helm chart values.
  * Added configurable liveness and readiness probes, including enablement, timing, and failure thresholds.
  * Probes remain enabled by default and can be disabled through chart configuration.

* **Improvements**
  * Health check settings can now be tuned for different deployment environments instead of relying solely on fixed defaults.

### DIFF
--- a/helm/holmes/templates/holmes.yaml
+++ b/helm/holmes/templates/holmes.yaml
@@ -103,9 +103,9 @@ spec:
           {{- toYaml . | nindent 10 }}
           {{- end }}
         command: ["python3", "-u", "server.py"]
-        {{- if .Values.args }}
+        {{- with .Values.args }}
         args:
-        {{- toYaml .Values.args | nindent 8 }}
+          {{- toYaml . | nindent 10 }}
         {{- end }}
         ports:
         - containerPort: 5050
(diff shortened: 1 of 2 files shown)


dabico:
@coderabbitai full review

PR #2293, opened 2026-07-15
feat: support pre-acquired AZURE_AD_TOKEN for Entra ID auth

wanghaoran1988:
## Summary

- When `AZURE_AD_TOKEN` environment variable is set, `get_azure_ad_token()` returns its value directly instead of acquiring a token via `DefaultAzureCredential`
- `DefaultAzureCredential` remains the fallback when `AZURE_AD_TOKEN` is not set
- No changes to the existing `AZURE_AD_TOKEN_AUTH` flow — this simply adds a short-circuit for pre-acquired tokens

## Motivation

ARO-RP runs HolmesGPT investigation pods on a Hive AKS cluster that has no Azure identity of its own. The RP acquires a short-lived Entra ID token via its own MSI and injects it into the pod as `AZURE_AD_TOKEN`. Without this change, the pod falls back to `DefaultAzureCredential` which fails because the pod has no credentials.

The token is valid for ~1 hour, and investigation pods time out at 10 minutes, so the pre-acquired token is always valid for the lifetime of the pod.

## Test plan

- [ ] Verified `AZURE_AD_TOKEN` env var is used when set
- [ ] Verified `DefaultAzureCredential` fallback works when `AZURE_AD_TOKEN` is not set
- [ ] Tested end-to-end with ARO-RP investigation pods on Hive AKS cluster

## Summary by CodeRabbit

* **New Features**
  * Added support for using a pre-acquired Azure AD token via the `AZURE_AD_TOKEN` environment variable.
  * When `AZURE_AD_TOKEN` is set and non-empty, it’s used immediately; otherwise, the app continues existing token acquisition and cache refresh.

* **Documentation**
  * Updated Azure token authentication documentation to include the `AZURE_AD_TOKEN` option.

* **Tests**
  * Added tests covering environment override behavior (including empty-string fallback) and cache reset between runs.

### DIFF
--- a/holmes/core/azure_token.py
+++ b/holmes/core/azure_token.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import threading
 import time
 from typing import Optional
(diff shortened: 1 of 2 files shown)


moshemorad (dismissed):

moshemorad:
Hi @wanghaoran1988 Looks like the DCO is falling, can you make sure to sign the commit.
If you will click on the failed DCO check in the PR you will get instruction how to fix it.

moshemorad (approved):

PR #2301, opened 2026-07-20
ROB-635 ROB-674 Reduce system prompt and tool description sizes (~50-60% fewer prompt tokens)

naomi-robusta:
## What

Condenses the Holmes prompts for Claude 4.x-class models while keeping every behavioral contract. Newer models over-trigger on repetition, ALL-CAPS emphasis and consequence language ("VIOLATION CONSEQUENCES", "FAILURE = INCOMPLETE INVESTIGATION"), so this cuts the scaffolding and keeps the domain facts, output contracts and non-obvious constraints.

**Sizes (cl100k tokens, representative server config with core_investigation enabled):**
- System prompt: **~15.3K → ~6.1K (−60%)**; full request payload incl. tool schemas: **−45%**
- `generic_ask.jinja2` TodoWrite/phases section: 2,536 → 331 tokens
- `investigator_instructions.jinja2`: 13.9KB → 1.7KB (was a verbatim coding-assistant prompt with dark-mode/npm examples and a "4 lines max" rule conflicting with the style guide)
- `conversation_history_compaction.jinja2`: ~1,200 → ~390 tokens
- Disabled-toolsets listing: one line per toolset; docs-URL pattern stated once, only exceptions listed
- bash prefix guidance ~40% smaller; prometheus/robusta llm-instructions deduped (rules stated 2-3× → once); `fetch_pod_logs` + kubernetes.yaml tool schemas tightened; per-message TodoWrite reminder ~120 → ~45 tokens

## Verification (live evals via OpenRouter, k3s + Elasticsearch Cloud)

Baselined every suite on the old prompts first, then re-ran per change group. Models chosen by production traffic distribution (opus-4.6 #1, opus-4.7 #2, haiku-4.5 #3, sonnet-4.6).

| Suite | opus-4.6 | opus-4.7 | haiku-4.5 | sonnet-4.6 |
|---|---|---|---|---|
| multicluster/transparency ES (13) | 12/13 baseline → 12/13 (same pre-existing 235 failure) | 13/13 | 9/13 baseline → 7–9/13 (within its 7–11 variance band, n=7 runs) | 10/13 |
| k8s regression (13) | 12/13 baseline → 12–13/13 (`24_misconfigured_pvc` baseline failure now passes) | — | — | — |
| ES quirks/skills (6) | 6/6 → 6/6 | 6/7 | n/a (0/6 at baseline, pre-existing) | 6/7 |
| bash prefix rules (9) | 9/9 | 8/9 (205 flaky ~1/3 runs on all configs) | — | — |
| compaction (7) | 7/7 baseline → 7/7 | — | — | — |

A more aggressive rewrite (extra ~480 tokens saved) was built and **rejected by evals**: haiku-4.5 averaged 6.75/13 with it vs 8.6/13 (n=9 runs), while opus was unaffected — weaker models need the extra structure, so this PR keeps the version that's safe across the whole prod model mix.

**Not eval-verified here** (no creds/infra in the dev sandbox): datadog/newrelic/coralogix instructions (untouched), OpenSearch PPL query-assist docs (untouched — at 56KB it's the biggest remaining candidate, needs a PPL eval first), prometheus eval fixtures can't run in the sandbox so the prometheus edits are strictly dedup-only. The `/eval` comment on this PR covers these areas in CI.

**Known-flaky tests observed during this work** (fail ~1/3 runs on baseline AND reduced prompts): `254_elasticsearch_dr_test_log_check`, `205_bash_deployment_logs_all_pods`; `235_elasticsearch_cluster_mismatch` fails consistently on every config/model except opus-4.7. Worth hardening separately.

## Test changes
- `test_toolsets_instructions.py`, `test_fetch_logs.py`: string pins updated to the new wording (same semantics).

https://claude.ai/code/session_01XgDuG2hb6jmSqYpD7iGUkd

---
_Generated by [Claude Code](https://claude.ai/code/session_01XgDuG2hb6jmSqYpD7iGUkd)_

## Summary by CodeRabbit

* **Improvements**
  * Strengthened TodoWrite multi-step workflows, task state handling, and evidence-backed completion with clearer blocker follow-ups.
  * Refined user-facing prompt and tool instructions for log retrieval, including consistent summaries, pod/namespace identification, and timeframe behavior.
  * Updated guidance for Bash command approval/allow-lists, Kubernetes/Prometheus/Robusta investigations, permissions-focused troubleshooting, and conversation-history compaction.
  * Improved messaging when toolsets are disabled or unavailable, including clearer next actions.
* **Tests**
  * Updated prompt wording assertions and added a new TodoWrite multi-step audit regression fixture.


naomi-robusta:
/eval
markers: regression or elasticsearch or logs or prometheus or compaction or multi-cluster or transparency or datadog
iterations: 3

Covers every area this PR touches: the regression suite plus the prompt sections verified locally (elasticsearch/multicluster/transparency/logs/compaction) and the areas that could not be verified in the dev sandbox (prometheus — infra can't run there; datadog — no creds there). 3 iterations for per-test signal instead of single-run flake noise.

---

naomi-robusta:
## Local high-iteration confidence runs (ITERATIONS=3, OpenRouter, k3s + ES Cloud)

Run on this branch's final state, models chosen by production traffic share:

| Suite | opus-4.6 | haiku-4.5 |
|---|---|---|
| bash prefix rules (9 tests ×3) | **27/27** | — |
| ES quirks/skills (7 tests ×3) | **19/21** (276, 279 one flake each — both pass on rerun) | — |
| multicluster/transparency (13 tests ×3) | **35/36** excluding `235` (pre-existing: fails on master's prompt too, all models except opus-4.7); the 1 miss is known-flaky `254` | 24/39 — per-iteration 9/13, 9/13, 6/13, same band and same recurring failure set (235/254/271) as the old-prompt baseline (mean 8.6/13 over 5 runs) |
| `191_elasticsearch_query_profile` (the ❌ in the label-triggered CI run above) | **3/3** — does not reproduce on opus-4.6 | — |

Notes for reviewers:
- The 6 setup failures in the label-triggered run are largely explained by that CI run and my local runs sharing the ES Cloud instance concurrently (same indices, colliding `before_test`) plus the infra-heavy 184/186/189 explosion tests; the `/eval` 3-iteration run above was triggered afterward and is the authoritative CI signal.
- Net: no reproducible regression vs the old prompt on any production model tested; two baseline failures (`24_misconfigured_pvc`, and `235` on opus-4.7) now pass.

---

naomi-robusta:
**Note on the first `/eval` run ([29736744262](https://github.com/HolmesGPT/holmesgpt/actions/runs/29736744262), "101 regressions"):** that run is not a valid signal — three compounding problems, none prompt-related:

1. **Oversized request (my mistake):** ~155 tests × 3 iterations on the single-node KIND cluster left the node severely overcommitted — fixture pods never reached readiness, producing 120 setup failures and degrading the tests that did run (e.g. `61_exact_match_counting`'s pods were created but never ready; it passes 3/3 in CI's own label-triggered run and locally).
2. **Shared-ES collision:** the elasticsearch tests ran concurrently with local 3-iteration confidence runs against the same ES Cloud instance (same indices, colliding setup/cleanup). Every ES test in this CI run failed while the same tests scored 19/21 locally in the same hour.
3. **Results comment exceeded GitHub's 64KB limit**, so the workflow's report step 422'd (worth a follow-up fix in `eval-regression.yaml`: truncate/attach the table when >65K).

Retrying at the label-run's scale below. The trustworthy signals so far remain: the label-triggered run (40/47, all failures explained) + the local 3-iteration table above.

---

naomi-robusta:
/eval
markers: regression or elasticsearch or prometheus or transparency
iterations: 2

Right-sized rerun of the earlier oversized request: ~65 unique tests × 2 iterations (close to the label-run scale that completed cleanly), keeping prometheus coverage (the area not verifiable in the dev sandbox) and dropping the infra-heavy logs/datadog markers that overloaded KIND. No local eval runs are hitting the shared ES instance now.

---

naomi-robusta:
/eval
branch: master
tags: regression or elasticsearch or prometheus or transparency
iterations: 2

Master-branch baseline at identical scope to the previous run, to separate real prompt regressions from environment/flake noise before deeper analysis. (Note: the previous run's results comment failed to post — GitHub's 64KB comment limit; results were read from the workflow logs: 145/172 passed, failures concentrated in transparency-tagged tests, under investigation.)

---

naomi-robusta:
## Branch vs master: identical-scope CI comparison (87 tests × 2, opus-4.6)

Both results comments failed to post (64KB comment limit — see note below); results extracted from workflow logs:

| | branch `claude/reduce-prompt-size-wcxmas` ([run](https://github.com/HolmesGPT/holmesgpt/actions/runs/29745876606)) | `master` baseline ([run](https://github.com/HolmesGPT/holmesgpt/actions/runs/29749012923)) |
|---|---|---|
| **Passed** | 145/172 | 147/172 |
| **Failed both iterations** | 19, 28, 65, 176, 235, 249, 184 | **19, 28, 65, 176, 235, 249** (identical set) |
| **Failed one iteration (flakes)** | 109, 185, 254 | 24, 270, 276 |
| **Setup failures** | 12 | 10 |

**Conclusion: no reproducible regression.** Every consistently-failing test fails identically on master:

- `19_detect_missing_app_details` — **broken fixture**: expects Holmes to surface `db-certs-authenticator`, but no before_test/manifest/mock creates it (the name exists only in unrelated compaction-test JSONs). Cannot pass under RUN_LIVE on any branch. Worth a separate fix.
- `65_health_check_followup` — expects "seems healthy", but opus-4.6 (on either prompt) probes the fixture's busybox fake-app deeply enough to discover nothing listens on its declared ports, and truthfully reports it as fake. Test predates this model's thoroughness.
- `28_permissions_error`, `249_overconfidence_postgres`, `176_network_policy`, `235_elasticsearch_cluster_mismatch` — fail 2/2 on both branches (235 already documented above as pre-existing).
- Also verified locally: 19 and 65 fail 0/3 on **both** the branch and a master worktree in the same k3s + opus-4.6 environment (prompt as the only variable).
- The branch-side flakes (109/185/254) and master-side flakes (24/270/276) are disjoint single-iteration misses — noise. Notably master flaked on `24_misconfigured_pvc`, which this branch passes consistently.

Separate follow-up worth filing: the eval workflow's results comment exceeds GitHub's 65,536-char limit for any run ≳90 rows, so the report step 422s and no results get posted (this also hides results for iterations≥2 runs). Suggest truncating the per-test table or attaching it as an artifact/gist in `eval-regression.yaml`.

---

naomi-robusta:
/eval
tags: hard or datadog
iterations: 2

Gap-closing run 1 of 2 (~52 tests): full hard tier + datadog suite (includes 91f_datadog_logs_historical_pod — the datetime-hallucination sentinel from the earlier prompt-ablation study). Medium tier follows as a separate run to stay under the KIND capacity that broke the first oversized request.

---

naomi-robusta:
/eval
branch: master
id: 108_logs_nearby_lines or 181_connectivity_check_http or 182_connectivity_check_http_url or 272_skill_suggestion_kubernetes or 206_bash_deployment_logs_all_pods_all_toolsets or 213_grafana_render_spike_detection or 237_grafana_large_dashboard_spike
iterations: 3

---

naomi-robusta:
/eval
branch: master
tags: hard
id: 108_logs_nearby_lines or 181_connectivity_check_http or 182_connectivity_check_http_url or 272_skill_suggestion_kubernetes or 206_bash_deployment_logs_all_pods_all_toolsets or 213_grafana_render_spike_detection or 237_grafana_large_dashboard_spike
iterations: 3

---

naomi-robusta:
## Disposition of the 11 `hard or datadog` failures — none are regressions from this PR

Compared the [branch run](https://github.com/HolmesGPT/holmesgpt/actions/runs/29751682286) (2 iters) against a [master baseline](https://github.com/HolmesGPT/holmesgpt/actions/runs/29759335056) of the same failing tests (3 iters, same model `opus-4.6`):

| Test | This PR | master | Verdict |
|---|---|---|---|
| `108_logs_nearby_lines` | 0/2 | **0/3** | Fails identically on master. Both branches find the staging-endpoint misconfig but miss the config-reload event preceding the first error. The "pass" in yesterday's ci-benchmark was a lucky single iteration. |
| `181_connectivity_check_http` | 0/2 | **0/3** | Environmental: the eval sandbox denies Holmes's `curl` commands (denied on master runs too), so exact HTTP status codes are unobtainable and the judge scores 0 on "must directly verify". |
| `182_connectivity_check_http_url` | 0/2 | **0/3** | Same as 181, plus the `internet` toolset isn't enabled in this eval's config — full verification is impossible in CI. |
| `272_skill_suggestion_kubernetes` | 0/2 | 1/3 | Known-flaky: its test_case.yaml documents the SuggestSkills trigger fires only ~25% of the time on opus-4.6. |
| `206_bash_deployment_logs_all_pods_all_toolsets` | 1/2 | 3/3 | Variance. The failing iter sampled 5 of 100 pods and missed pod-79's unique error — but the "3 representative pods is enough" guidance that drives this shortcut exists verbatim on master. Worth keeping an eye on. |
| `213_grafana_render_spike_detection` | 1/2 | 2/3 | Flaky on both branches — noise. |
| `237_grafana_large_dashboard_spike` | 1/2 | 3/3 | Small-n variance (1 failure across both runs of this test on the branch). |

Also checked the removed prompt text against each failure mode: the five-whys / root-cause-depth guidance survived the reduction, no curl/internet-toolset instructions were removed, and no prompt on either branch ever encoded the "read log lines before the first error" skill that 108 tests.

Notes for future `/eval` users on this PR: the `evals-tag-elasticsearch` label gets ANDed into `/eval` runs that don't specify `tags:` explicitly — that's what killed the [first baseline attempt](https://github.com/HolmesGPT/holmesgpt/pull/2301#issuecomment-5024079466) (collected 1 test). Always pass `tags:` explicitly when combining with `id:`.

---

holmes/core/prompt.py
```diff
@@ -136,11 +136,10 @@ def append_all_files_to_user_prompt(
 
 def get_tasks_management_system_reminder() -> str:
```
Sheeproid, line 137: @naomi-robusta this relates to the TodoWrite tool. Right now, we've disabled it multiple places, but it might not be disabled by default.

There are to possiblities:
1. The test did run with 'Fast Mode' on. In this case, changing this system reminder had no effect and this change was essentially never tested.
2. The test didn't run with 'Fast Mode'. In this case, the entire testing process might be wrong and we need to re-run it with Fast Mode on.

holmes/plugins/prompts/_default_log_prompt.jinja2
```diff
@@ -1,13 +1,4 @@
...
-* If you are not given the pod's exact name, or only have an application name or a deployment name, look for related pods using kubectl commands. Ask the user if you can't infer the pod logs.
-* Do fetch application logs yourself and DO not ask users to do so
-* If you have an issue id or finding id, use `fetch_finding_by_id` as it contains time information about the issue (`starts_at`, `updated_at` and `ends_at`).
-** Then, use `start_time=-300` (5 minutes before `end_time`) and `end_time=<issue start_at time>`  when calling `fetch_pod_logs`.
-** If there are too many logs, or not enough, narrow or widen the timestamps
-* If the user did not explicitly ask about a given timeframe, ignore the `start_time` and `end_time` so it will use the default.
-* IMPORTANT: ALWAYS inform the user about the actual time period fetched (e.g., "Looking at logs from the last <X> days")
-* IMPORTANT: If a limit was applied, ALWAYS tell the user how many logs were shown vs total (e.g., "Showing latest <Y> of <Z> logs")
-* IMPORTANT: If any filters were applied, ALWAYS mention them explicitly
+* Use the tool `fetch_pod_logs` to access an application's logs yourself — do not ask users to do so.
```
Sheeproid, line 1: do we really need the logs prompt? I saw that haiku was not performing well in some cases, but was the log prompt the main diff?

Sheeproid: same questions for all other log prompt related files

holmes/plugins/prompts/_toolsets_instructions.jinja2
```diff
@@ -26,45 +24,13 @@
...
-      {%- if toolset.error %}
-    *  error: {{ toolset.error }}
-      {%- endif -%}
-    {%- else %}
-    *  status: {{ toolset.status.value }}
-    {%- endif %}
-    {%- if toolset.docs_url %}
-    *  setup instructions: {{ toolset.docs_url }}
-    {%- endif -%}
+{%- set default_docs_url = "https://holmesgpt.dev/data-sources/builtin-toolsets/" ~ toolset.name.split("/")[0] ~ "/" %}
```
Sheeproid, line 29: did you run a test where tool is disabled and holmes needs to tell the user to enable it, so this case covered?

holmes/plugins/prompts/_toolsets_instructions.jinja2
```diff
@@ -26,45 +24,13 @@
...
-   - For custom toolsets: https://holmesgpt.dev/data-sources/custom-toolsets/
-   - For remote MCP servers: https://holmesgpt.dev/data-sources/remote-mcp-servers/
-
-When providing configuration guidance, always prefer the specific toolset documentation URL when available.
-When the user needs access to a REST API that has no dedicated toolset, proactively suggest the HTTP connector as the simplest option.
+If you need a system you can't reach with your current tools: for a matching FAILED toolset, tell the user and copy its error into your response; for a matching disabled toolset (including when users ask how to set up or grant access to a system), ask the user to configure it and share its setup URL. If no toolset matches, say you're missing an integration for that system and that one can be added via an [HTTP connector](https://holmesgpt.dev/data-sources/api-toolsets/) (quickest for REST APIs — supports auth, endpoint whitelisting and response filtering), an [MCP server](https://holmesgpt.dev/data-sources/remote-mcp-servers/), or a [custom toolset](https://holmesgpt.dev/data-sources/custom-toolsets/).
```
Sheeproid, line 36: same question as the other one in this file

holmes/plugins/prompts/conversation_history_compaction.jinja2
```diff
@@ -1,89 +1,16 @@
-This conversation history is too long and no longer fits your context window. 
-Your task is to create a detailed summary of the conversation so far. This will help continuing the investigation after summarisation is done.
```
Sheeproid, line 2: have you run compaction evals?

holmes/plugins/prompts/generic_ask.jinja2
```diff
@@ -1,305 +1,65 @@
 {% if intro_enabled %}
-You are HolmesGPT version {{ holmes_version }}, a tool-calling AI assist provided with common devops and IT tools that you can use to troubleshoot problems or answer questions.
-Ask for multiple tool calls at the same time as it saves time for the user.
+You are HolmesGPT version {{ holmes_version }}, an AI agent for troubleshooting and answering questions about DevOps, IT and observability systems, using tool calls to gather data.
+Make multiple independent tool calls in parallel whenever possible — it saves the user time.
```
Sheeproid, line 3: Have you checked this iteration still works as expected?
In general, I"m missing a cost comparison between the evals, which is what matters in addition to the correctness of them. Did you make one?

holmes/plugins/prompts/generic_ask.jinja2
```diff
@@ -1,305 +1,65 @@
...
-  4. If verification shows the data is from a different cluster (commonly `{{ cluster_name }}` itself, because that's the cluster this Holmes is on) → you do NOT have data for the cluster the user asked about. You MUST: (a) state this plainly and prominently at the top of your answer, (b) NOT present those mismatched findings as the root cause of the user's reported incident, (c) say which cluster(s) you DID find data for and label any summary of that data accordingly, (d) suggest the user point you at the right data source / Holmes agent for the cluster they actually care about.
-- Never silently relabel data from one cluster as if it were from another. Every finding must carry its real cluster of origin.
+**Multi-cluster awareness:** your local Kubernetes context (kubectl, in-cluster APIs, cluster-scoped toolsets) only sees the cluster `{{ cluster_name }}`. External observability toolsets (Elasticsearch/OpenSearch, Datadog, Loki, Grafana, Prometheus, NewRelic, Coralogix, etc.) may contain data from other clusters too — inspect what the data actually contains rather than assuming.
```
Sheeproid, line 9: Looks like this part of the prompt is what conflicts with our new multi-agent mcp we might want to remoev/fix this either now or in a different issue.

Also, did you run any evals that checks for this problem of internal vs external clusters and usage of tools like elasticsearch?

holmes/plugins/prompts/generic_ask.jinja2
```diff
@@ -1,305 +1,65 @@
...
 
 {% if skills_enabled -%}
-# Skill Usage:
-If a skill in the catalog clearly matches the issue being investigated, fetch it using the `fetch_skill` tool before diving into other tools.
-Only fetch skills that are relevant to the specific issue — do not fetch skills speculatively or "just in case".
-If no skill matches, skip this step and investigate directly with available tools.
-
-After fetching a skill, read the content returned in the tool's data field and follow its steps.
-Skill content takes priority over general investigation steps.
+# Skill Usage
```
Sheeproid, line 24: Have you run a test that uses skills to verify holmes still knows how to use skills?

holmes/plugins/prompts/generic_ask.jinja2
```diff
@@ -1,305 +1,65 @@
...
+If a skill in the catalog clearly matches the issue being investigated, fetch it with `fetch_skill` before diving into other tools — but only when it clearly matches, never speculatively.
+Read the fetched skill's content and follow its steps; skill content takes priority over general investigation steps.
 {%- endif %}
 
 {% if todowrite_enabled %}
-CLARIFICATION REQUIREMENT: Before starting ANY investigation, if the user's question is ambiguous or lacks critical details, you MUST ask for clarification first. Do NOT create TodoWrite tasks for unclear questions.
```
Sheeproid, line 37: have you ran a test todowrite_enabled to see it has an effect? (also relates to my comments about fast mode)

holmes/plugins/prompts/generic_ask.jinja2
```diff
@@ -1,305 +1,65 @@
...
-* Treat error messages as exact diagnostic evidence. `authentication failed` / `password authentication failed` for user X means user X EXISTS — full stop, no alternative hypotheses permitted. `role does not exist` / `user not found` means the user is absent. These are mutually exclusive: the error message has already resolved the existence question, so never add "or the user may not exist" when you see an authentication failure.
-* Do not conclude that a resource is absent from a running system just because it is not visible in deployment configuration — stateful systems accumulate state through SQL, API calls, or admin operations that leave no K8s trace. If you cannot read a value (e.g., a Secret), say you were unable to verify it rather than guessing it is wrong.
-* ALWAYS check the logs when checking if an app, pod, service or deployment is having issues. Something "running" and reporting healthy does not mean it is without issues.
+# Investigation guidelines
+
+* Gather data with as many tool calls as needed before responding; keep going until you reach the deepest root cause you can find.
```
Sheeproid, line 44: I doubt that haiku really needs all of these instructions. I noticed that Claude said they are. but I think they don't. Perhaps we can have a followup ticket to reduce the system prompt even more.

holmes/plugins/prompts/generic_ask.jinja2
```diff
@@ -309,41 +69,25 @@ If the answer to any of those questions is 'yes' - The investigation is INCOMPLE
...
-
-{% if general_instructions_enabled %}
-# Special cases and how to reply
-
-* Make sure you differentiate between "I investigated and found error X caused this problem" and "I tried to investigate but while investigating I got some errors that prevented me from completing the investigation."
-* As a special case of that, If a tool generates a permission error when attempting to run it, follow the Handling Permission Errors section for detailed guidance.
-* That is different than - for example - fetching a pod's logs and seeing that the pod itself has permission errors. in that case, you explain say that permission errors are the cause of the problem and give details
-* Issues are a subset of findings. When asked about an issue or a finding and you have an id, use the tool `fetch_finding_by_id`.
-* For any question, try to make the answer specific to the user's environment.
+If a tool hits a permissions error (e.g. `Error from server (Forbidden):`): identify the missing resource, API group and verbs from the error; check which user/service account you run as and its permissions; report this to the user with a link to https://holmesgpt.dev/data-sources/permissions/
```
Sheeproid, line 72: Have we tested this scenario to see it is not harmed and Holmes links to https://holmesgpt.dev/data-sources/permissions/ ?

holmes/plugins/toolsets/bash/bash_instructions.jinja2
```diff
@@ -1,71 +1,19 @@
...
-- Good: `kubectl scale deployment nginx -n prod --replicas=3` (prefix `kubectl scale deployment nginx -n prod` matches)
-- Bad: `kubectl scale --replicas=3 deployment nginx -n prod` (prefix won't match)
-
-General principle: identifiers first (what you're targeting), parameter values last (how you're affecting it).
-
-## Prefix Guidelines
-
-For each command segment (separated by `|`, `&&`, `||`, `;`, `&`), provide a `suggested_prefixes` entry based on the operation type:
-
-### Read-only operations
```
Sheeproid, line 41: have you ran all the bash evals? I added these instructions because Holmes chose the wrong prefixes a lot of the times

holmes/plugins/toolsets/investigator/investigator_instructions.jinja2
```diff
@@ -1,244 +1,14 @@
-# Task Management
```
Sheeproid, line 1: same question I had in other places as to is this flow is tested or not

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -142,12 +142,12 @@ def __init__(self, toolset: "KubernetesLogsToolset"):
                 description="Kubernetes namespace", type="string", required=True
             ),
             "start_time": ToolParameter(
```
Sheeproid, line 144: BTW is the loggin_api file is still used? I thought we replaced it

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -142,12 +142,12 @@ def __init__(self, toolset: "KubernetesLogsToolset"):
...
-                description=f"Start time for logs. Can be an RFC3339 formatted datetime (e.g. '2023-03-01T10:30:00Z') for absolute time or a negative string number (e.g. -3600) for relative seconds before end_time. Default: -{DEFAULT_TIME_SPAN_SECONDS} (last {DEFAULT_TIME_SPAN_SECONDS // SECONDS_PER_DAY} days)",
+                description=f"RFC3339 datetime (e.g. '2023-03-01T10:30:00Z') or negative number as string for seconds before end_time (e.g. '-3600'). Default: -{DEFAULT_TIME_SPAN_SECONDS} (last {DEFAULT_TIME_SPAN_SECONDS // SECONDS_PER_DAY} days)",
```
Sheeproid, line 145: have you tested this date scenarios?

holmes/plugins/toolsets/logging_utils/logging_api.py
```diff
@@ -142,12 +142,12 @@ def __init__(self, toolset: "KubernetesLogsToolset"):
...
                 type="string",
                 required=False,
             ),
             "end_time": ToolParameter(
-                description="End time for logs. Must be an RFC3339 formatted datetime (e.g. '2023-03-01T12:30:00Z'). If not specified, defaults to current time.",
+                description="RFC3339 datetime (e.g. '2023-03-01T12:30:00Z'). Default: now.",
```
Sheeproid, line 150: same question

holmes/plugins/toolsets/prometheus/prometheus_instructions.jinja2
```diff
@@ -9,81 +9,31 @@ You are using Coralogix Prometheus.
 * Example: `container_cpu_utilization{namespace="test-173"}` can fail if the label is named differently (e.g., `k8s_namespace_name`) or the metric is named differently. Do not assume names—only use labels you have seen returned or been told exist.
 {%- endif %}
```
Sheeproid, line 11: I noticed Claude saying it was not able to run Prometheus tests. are you trusting this change then?

holmes/plugins/toolsets/robusta/robusta_instructions.jinja2
```diff
@@ -1,57 +1,22 @@
 # Configuration and alerts history
-* Use fetch_configuration_changes_metadata to get historical configuration changes in a cluster or for a specific workload.
```
Sheeproid, line 2: were relevant tests ran?

Sheeproid (changes requested):
Hi, there are a lot questions here, please discuss with me.
In general:
1.  I might be missing context about which tests you ran to make sure different parts are correct
2. Did you ran a cost-comparison analysis of the eval, showing which one costs more? If not, then please do as we must prove this change actually reduces the cost and not just the prompt size. For example, if we removed parts how to run tools in the system prompt and now Holmes runs more iterations, then we still lose money and tokens. Note - we need to compare both correctness, cost, AND time

naomi-robusta:
@Sheeproid Thanks for the thorough review — answering everything in one place. TL;DR: the cost analysis you asked for exists and shows **~18% lower cost per eval with no iteration inflation**; most flows you asked about have direct eval coverage; and you caught one real gap (TodoWrite) which I've now verified separately and propose to close with a follow-up eval fixture.

## 1. Cost comparison (your main ask)

Same 7 tests, same model (`opus-4.6`), [branch run](https://github.com/HolmesGPT/holmesgpt/actions/runs/29751682286) (2 iters) vs [master run](https://github.com/HolmesGPT/holmesgpt/actions/runs/29759335056) (3 iters), averaged per test:

| Test | Branch $/run | Master $/run | Δ cost | Branch turns | Master turns |
|---|---|---|---|---|---|
| 108_logs_nearby_lines | $0.240 | $0.287 | **↓16%** | 6.0 | 5.3 |
| 181_connectivity_check_http | $0.120 | $0.144 | **↓17%** | 3.0 | 3.0 |
| 182_connectivity_check_http_url | $0.098 | $0.136 | **↓28%** | 2.0 | 2.3 |
| 206_bash_deployment_logs… | $0.203 | $0.252 | **↓19%** | 3.5 | 4.0 |
| 213_grafana_render_spike | $0.238 | $0.281 | **↓15%** | 4.5 | 4.7 |
| 237_grafana_large_dashboard | $0.222 | $0.272 | **↓18%** | 4.5 | 5.0 |
| 272_skill_suggestion | $0.166 | $0.193 | **↓14%** | 3.0 | 4.3 |
| **Mean** | **$0.184** | **$0.224** | **↓18%** | **3.8** | **4.1** |

Directly to your worry ("fewer instructions → more iterations → still lose money"): **turns went slightly down** (3.8 vs 4.1 avg) and tool calls are flat (4.7 vs 5.1 avg). Every single test got cheaper. The benchmark comparison in the branch run shows the same on 179_grafana (cost ↓15–19%, same turns/tool-calls).

## 2. Per-comment answers (what was tested)

- **`_toolsets_instructions` — disabled-toolset flow (both comments):** directly covered — `266_toolset_disabled_vs_no_data` 2/2 pass and `111_disabled_datadog_traces` 2/2 pass on the branch (Holmes correctly told the user to enable the toolset / distinguished disabled from no-data).
- **Log prompts (`_default_log_prompt`, `_fetch_logs`, 108):** the whole datadog logs family (91a–91i, 12/12 iters passed), `205/206`, `21`, `25` all ran on the branch. `108` fails 0/2 on the branch but also **0/3 on master** — pre-existing, not from this diff. On "do we need the log prompt at all" — plausible for the follow-up reduction; this PR only condenses it.
- **Compaction evals:** yes — `233_compaction_prometheus_data` 2/2 and `238_image_compaction_overflow` 2/2 (6 compactions exercised per run).
- **`generic_ask` iteration loop:** exercised by the full battery (126 branch runs); cost/turn data above shows no behavioral drift.
- **Permissions section (line 72):** `21_job_fail_curl_no_svc_account` passed 2/2 (permission error inside pod correctly treated as root cause, not investigation failure). The docs-link text is retained; there's no eval asserting the URL itself — same coverage level as master.
- **Skills (line 24):** branch run generated 10 skills; both skill replays loaded the captured skill and answered correctly (2/2). `272` (SuggestSkills) failed at its documented ~25% trigger rate on **both** branches (0/2 branch, 1/3 master).
- **Multi-cluster / external elasticsearch (line 9):** `185_elasticsearch_cross_region_search` 2/2 pass. Agree the cluster-verification section may conflict with the multi-agent MCP — I'd handle that in a separate issue rather than growing this diff.
- **Bash prefixes (line 41):** `205`, `206`, `181`, `182` ran, plus every k8s eval exercises prefix approval (the runs' denied-command audit shows validation active). `206` passed 1/2 branch vs 3/3 master (small-n; the failing iter sampled pods, driven by the "3 representative pods" rule that exists verbatim on master).
- **Prometheus (line 11):** correct that my local sandbox couldn't run them (cluster limitations) — but CI did, on the branch: `233` 2/2, `281_prometheus_metric_rename` 2/2, `92_cpu_graph` 6/6, `242` 2/2.
- **Robusta instructions:** `91h/91i` empty-query, `242_cpu_graph_robusta_runner`, `92` conversation flows, `111` — all passed on the branch.
- **`logging_api.py` dates (3 comments):** `91d_datadog_metrics_historical_pod` 6/6 and `91f_datadog_logs_historical_pod` 4/4 exercise `start_time`/`end_time` in both RFC3339 and relative forms. And yes, the file is still live — `PodLoggingTool` is the shared base for `fetch_pod_logs` across kubernetes/logs, datadog, coralogix etc.; the eval traces call it.

## 3. TodoWrite / Fast Mode (`prompt.py:137`, `generic_ask:37`, investigator files) — you found a real gap

You're right, with a nuance: it's neither of your two possibilities exactly. The eval harness **disables TodoWrite by default** (drops the tool + turns off `TODOWRITE_INSTRUCTIONS`/`TODOWRITE_REMINDER` unless a fixture sets `enable_todo` — and **no fixture in the repo sets it**), so the TodoWrite prompt reductions were not behaviorally covered by the eval battery — on this branch or on master, ever. Additionally, the `prompt.py:137` reminder is only rendered in the **CLI `ask`** path; `/api/chat` (`build_chat_messages`) passes `include_todowrite_reminder=False` regardless of Fast Mode, so evals mirror production chat there.

What coverage exists now:
- Rendering unit tests (`tests/core/test_prompt.py`) assert the reminder appears in CLI prompts — passing.
- I just ran a live CLI smoke (TodoWrite enabled, opus-4.6, multi-step investigation question) on **both** branch and master: both invoke TodoWrite; the branch produced a 7-task plan, kept statuses updated through completion, and delivered a prioritized answer. Single-shot smoke, not statistics.

Proposed fix for durable coverage: add an `enable_todo: true` eval fixture (follow-up PR) so this flow is regression-tested going forward. I'd rather do that than re-run the whole battery with todos on, since the harness intentionally models the production chat path.

## 4. Agreed follow-ups

- Further reduction for haiku-class models (line 44) — agree the remaining instructions are probably still over-specified for opus-class; follow-up ticket alongside the other identified targets (`kubernetes.yaml` tool schemas — 96% untouched, largest remaining chunk — and the frontend prompt files).
- Multi-agent MCP vs cluster-verification section — separate issue.
- `enable_todo` eval fixture — follow-up PR.

Happy to discuss any of these live.

---

tests/llm/fixtures/test_ask_holmes/283_todowrite_multistep_audit/test_case.yaml
```diff
@@ -0,0 +1,57 @@
...
+      echo "❌ audit_code missing in app-283-$NS"
+      exit 1
+    fi
+  done
+  echo "✅ Test data ready in app-283-*"
+
+after_test: |
+  for NS in billing search media ledger; do
+    kubectl delete namespace "app-283-$NS" --ignore-not-found
+  done
```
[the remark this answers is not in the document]
naomi-robusta: The reentrancy guideline this cites applies to **cloud service evals**, which share a single external account across parallel PR runs. This is a Kubernetes eval: each PR run gets its own KIND cluster, so `app-283-*` namespaces are never shared between runs (per CLAUDE.md: "Kubernetes evals don't have this problem because each PR gets its own KIND cluster"). Within one run, the harness executes `before_test`/`after_test` once per unique eval — parallel iterations share the setup but teardown only runs after all iterations complete. Keeping the namespace deletion matches every other k8s fixture in this suite and the documented cleanup verification step ("`kubectl get namespace app-NNN` should return NotFound").

---

holmes/plugins/prompts/generic_ask.jinja2
naomi-robusta: Good catch — this was a real gap, and testing it found a real regression. No eval ran with `enable_todo: true` (the harness disables todos by default), so I added one: `283_todowrite_multistep_audit`, a 4-namespace audit with unique per-namespace codes. Result: on this branch as reviewed, gpt-4.1-mini called TodoWrite in **0/4** runs vs **3/4** on master — the condensed wording ("use the TodoWrite tool") had lost master's imperative force. Fixed in 9793f9f by restoring "CRITICAL: … you MUST use the TodoWrite tool" (a few tokens): back to master parity with gpt-4.1-mini (2/3) and 3/3 with gpt-4.1, the CI regression model. Answers were correct in all runs either way — the loss was planning/progress visibility, not correctness. The eval is tagged `regression` so this stays covered.

---

holmes/plugins/prompts/generic_ask.jinja2
naomi-robusta: The link is retained in the condensed prompt (generic_ask.jinja2 line 72), and it's now pinned by a unit test added in 58f072a (`test_system_prompt_pins_permission_error_docs_link`) so any future prompt edit that drops the https://holmesgpt.dev/data-sources/permissions/ reference fails CI.

---

holmes/core/prompt.py
naomi-robusta: The gap was real but slightly different from both options: the eval harness strips TodoWrite unless a fixture sets `enable_todo`, and no fixture ever did — so these edits were never eval-tested on any branch. Closing that found a real regression (gpt-4.1-mini called TodoWrite 0/4 vs 3/4 on master). Fixed in 9793f9f by restoring the imperative wording; new eval `283_todowrite_multistep_audit` (tagged `regression`) keeps it covered. Note this line renders only in CLI `ask` — `/api/chat` passes `include_todowrite_reminder=False`.

---

holmes/plugins/prompts/_default_log_prompt.jinja2
naomi-robusta: Halved it rather than removed it — and the data says the log prompt wasn't haiku's problem: haiku's branch scores (9/13, 9/13, 6/13) sit inside the old-prompt baseline band (mean 8.6/13), with the same recurring failures (235/254/271). Log evals hold on the reduced prompt: datadog logs 91a–91i 12/12 iterations; `108` fails 0/3 on master too. Whether it can be dropped entirely needs its own ablation — tracked in ROB-758.

---

holmes/plugins/prompts/_toolsets_instructions.jinja2
naomi-robusta: Yes: `266_toolset_disabled_vs_no_data` 2/2 and `111_disabled_datadog_traces` 2/2 on this branch — Holmes reports the toolset as unavailable and points at enabling it. The rendered format (FAILED + error + setup link) is also pinned by `test_toolsets_instructions.py`.

---

holmes/plugins/prompts/_toolsets_instructions.jinja2
naomi-robusta: Same coverage as the thread above: `266_toolset_disabled_vs_no_data` and `111_disabled_datadog_traces`, both 2/2 on the branch.

---

holmes/plugins/prompts/conversation_history_compaction.jinja2
naomi-robusta: Yes: `233_compaction_prometheus_data` 2/2 and `238_image_compaction_overflow` 2/2 on the branch — 238 triggers ~6 compactions per run, so the reduced prompt is exercised hard. The hard/datadog CI battery also logged 12 compactions with no related failures.

---

holmes/plugins/prompts/generic_ask.jinja2
naomi-robusta: Yes — same 7 tests, same model (opus-4.6), branch vs master CI runs: mean **$0.184 vs $0.224 per run (↓18%)**, turns **3.8 vs 4.1**, tool calls 4.7 vs 5.1. Every test got cheaper (↓14–28%) and turns went slightly *down*, so the saving isn't being paid back as extra iterations.

---

holmes/plugins/prompts/generic_ask.jinja2
naomi-robusta: Evals: yes — `185_elasticsearch_cross_region_search` 2/2, and the multicluster/transparency suite 35/36 locally at 3 iterations (the miss is known-flaky `254`; `235` fails on master's prompt too on all models except opus-4.7).

The multi-agent MCP conflict predates this PR — the section is condensed but unchanged in meaning. Agreed it should be fixed, but separately, so this PR stays behavior-neutral. I can open the issue with this thread as context.

---

holmes/plugins/prompts/generic_ask.jinja2
naomi-robusta: Yes, both directions: branch runs generated skills (26 in the label run, 10 in the hard run), and skill replays loaded 4/4 and 2/2, all scored correct (replay vs primary: ↓31% cost/↓48% tokens, and ↓61%/↓69%). `272_skill_suggestion` failed at its documented ~25% trigger flake rate on both branches (0/2 branch, 1/3 master) — noise, not a prompt effect.

---

holmes/plugins/prompts/generic_ask.jinja2
naomi-robusta: Agreed — filed ROB-758 ("Prompt reduction round 3"). First item is exactly this: haiku-class trims, validated branch-vs-master on cost + turns. Also in scope: `kubernetes.yaml` tool schemas (96% untouched, the largest remaining chunk) and the frontend prompt files.

---

holmes/plugins/toolsets/bash/bash_instructions.jinja2
naomi-robusta: Yes — the bash prefix-rules suite, 9 tests × 3 iterations = **27/27** on opus-4.6 with the reduced instructions. One honest flag: `206_bash_deployment_logs_all_pods` went 1/2 vs master 3/3 (the failing iteration sampled 5 of 100 pods), but the guidance it followed exists verbatim on master — variance, worth watching rather than blocking.

---

holmes/plugins/toolsets/investigator/investigator_instructions.jinja2
naomi-robusta: This file had the same coverage hole as the TodoWrite section in generic_ask — closed by the same fix: eval `283_todowrite_multistep_audit` exercises it end-to-end (plan created before data gathering, statuses updated, 4 namespaces audited), and this file's strengthened wording is part of the 9793f9f regression fix.

---

holmes/plugins/toolsets/logging_utils/logging_api.py
naomi-robusta: Yes — what got replaced was the old kubectl-logs YAML toolset. This file is the base every pod-logging toolset still subclasses (kubernetes logs, Datadog, Coralogix, …), so they all inherit these trimmed parameter descriptions — which is why cutting here pays off across every logging backend at once.

---

holmes/plugins/toolsets/logging_utils/logging_api.py
naomi-robusta: Yes — the historical/date-window evals ran clean on the trimmed descriptions: `91d_datadog_metrics_historical_pod` 6/6 and `91f_datadog_logs_historical_pod` 4/4. 91f is the datetime-hallucination sentinel from the earlier prompt-ablation study, so it's the strongest signal this guidance still prevents what it was written for.

---

holmes/plugins/toolsets/logging_utils/logging_api.py
naomi-robusta: Same evidence as the thread above: `91d` 6/6, `91f` 4/4 on the branch.

---

holmes/plugins/toolsets/prometheus/prometheus_instructions.jinja2
naomi-robusta: Right — the sandbox can't run Prometheus, so this wasn't trusted blind; verification moved to CI on the branch prompt: `233_compaction_prometheus_data` 2/2, `281_prometheus_metric_rename` 2/2, `92_cpu_graph` 6/6, `242` 2/2.

---

holmes/plugins/toolsets/robusta/robusta_instructions.jinja2
naomi-robusta: Yes — via the transparency/multicluster tags that exercise the Robusta toolset: 35/36 locally at 3 iterations, and the identical-scope branch-vs-master CI comparison (145/172 vs 147/172, same consistent-failure sets on both) shows no robusta-flow difference. The one CI flake (`191_elasticsearch_query_profile`) passes 3/3 locally on this branch.

---

Sheeproid (approved):

holmes/plugins/prompts/_default_log_prompt.jinja2
naomi-robusta: Follow-up with direct coverage: #2328 adds eval `284_log_fetch_transparency` (asserts the log-scope disclosure this prompt mandates, with an anti-hallucination ref code) plus a unit pin so the mandate can't be silently dropped. Interesting finding from building it: compliance is only ~2/3 on both gpt-4.1 and opus-4.6 — with master's wording too — so it's tagged `medium` as an improvement target.

---

holmes/plugins/prompts/generic_ask.jinja2
naomi-robusta: Follow-up: it turns out `28_permissions_error` — the eval that asserts exactly this behavior end-to-end (permission error + the docs link) — has been Setup-Failing on every branch because the harness never sets `TMPDIR`, not because of Holmes. #2328 fixes the fixture; it now passes 3/3 on the current prompt. Combined with the unit pin from 58f072a, this scenario is now covered both behaviorally and structurally.

---

PR #2311, opened 2026-07-21
ROB-677 Implement remote tool call approval workflows

Avi-Robusta:
## Summary

Adds user approval for **cross-cluster remote tool calls**. Previously a tool that required approval (e.g. a bash command) on a remote cluster was rejected outright — the approval requirement never reached the calling user. Now it surfaces to the original caller's normal approval flow (CLI callback, Slack form, or web UI) and, once approved, the tool actually runs on the remote cluster.

The design is **metadata-only**: approval state travels in the existing `metadata` and `tool_response` JSONB columns of `RemoteToolCalls`. **No database schema change or migration** — the earlier storage-schema approach was dropped.

## How it works

1. The caller's LLM invokes a `remote_*` MCP tool (via relay / platform-mcp), which creates a `RemoteToolCalls` row.
2. The target Holmes (`ToolCallWorker._execute`) claims the row and runs the tool. If the tool returns `APPROVAL_REQUIRED` and the caller has not yet approved (`metadata.remote_tool_approved` is falsy), the target returns a `tool_response` with `status: "approval_required"` and posts the row as `COMPLETED` — **no target-side polling and no `pending_approval` row status**.
3. Relay renders that response back to the caller.
4. The caller's `RemoteMCPTool._invoke_async` parses `status == "approval_required"` into a `StructuredToolResult(APPROVAL_REQUIRED)`, so the **existing** LLM approval pause/resume fires unchanged — CLI callback, Slack form, or web UI.
5. On approval, the caller re-invokes the tool, injecting reserved args `__robusta_user_approved` (and, for remote tools, `__robusta_session_approved_prefixes`). Relay pops these and forwards them via `metadata.remote_tool_approved` / `tool_request.session_approved_prefixes`.
6. The target now sees `remote_tool_approved=true`, runs the tool, and returns the real result. **Fail-closed:** if the tool still reports `APPROVAL_REQUIRED` after approval, the target errors rather than looping.

Approval tokens are handled entirely by the caller's own local pause/resume (minted on pause, verified on resume); there is no separate remote-token round-trip.

## Scope & behavior

- **Any tool that returns `APPROVAL_REQUIRED` is supported, not just bash.** Approval is gated generically in `Tool.invoke` via a toolset's `approval_required_tools` fnmatch patterns or a tool's `requires_approval()` override. Bash additionally offers the "don't ask again" prefix shortcut (`suggested_prefixes`), which simply doesn't appear for non-bash tools.
- **Per-(conversation, cluster) prefix scoping.** Session-approved bash prefixes are keyed by target agent (`bash_session_approved_agent`). Approving `curl` for cluster A does not auto-approve it on cluster B; legacy (unkeyed) metadata is treated as local-only.
- **The target cluster is named in the approval prompt** — Slack (`on remote cluster \`X\``) and the web UI both show which cluster the command will run on.
- **Version skew is fail-closed.** Cross-cluster execution requires an identical caller/executor Holmes version (a pre-existing guard in `ToolCallWorker._execute`); a mismatch returns a clear `version mismatch` error and never runs the tool.

## Tests

- `tests/core/conversations_worker/test_tool_call_worker.py` — first call surfaces `APPROVAL_REQUIRED`; approved re-invocation runs with `user_approved=True`; fail-closed loop guard; session-prefix forwarding to the invoke context.
- `tests/toolsets/mcp/test_approval_response_parsing.py` — parses the lowercase `approval_required` status; injects the reserved approval arg only when approved; remote one-liner names the target cluster and hides routing/reserved params.
- `tests/toolsets/mcp/test_remote_bash_tool_approval_flow.py`, `tests/toolsets/bash/test_bash_approval_flow.py`, `tests/test_bash_prefix_cluster_scoping.py`.

## Related PRs

- **robusta-dev/relay#658** — relay caller/forwarding side (reserved-arg popping, metadata/`tool_request` forwarding) + Slack cluster naming.
- **robusta-dev/robusta-frontend** (branch `claude/remote-tool-approvals-2uysrg`) — web UI approval modal shows the target cluster.
- **robusta-dev/robusta-storage** — **no changes** (metadata-only design; the earlier schema/migration PR was dropped).

---
_Generated by [Claude Code](https://claude.ai/code/session_0129KYDgRPTDB5GoWpwgCbHE)_


holmes/plugins/toolsets/mcp/toolset_mcp.py
```diff
@@ -782,6 +829,33 @@ def _parse_tool_parameter(
         )
 
     def get_parameterized_one_liner(self, params: Dict) -> str:
+        is_remote = self.name.startswith("remote_")
```
naomi-robusta: I don't like checking that the tool name starts with "remote_". why is this necessary?

naomi-robusta (dismissed):

holmes/core/tool_calling_llm.py
```diff
@@ -127,8 +127,12 @@ def _extract_text_from_content(content: Any) -> str:
...
 
 
-def _bash_prefix_scope(tool_name: str, tool_params: Dict[str, Any]) -> str:
-    if tool_name and tool_name.startswith("remote_"):
+def _bash_prefix_scope(is_remote: bool, tool_params: Dict[str, Any]) -> str:
+    """Scope key into the agent-keyed session-prefix map. Remote (cross-cluster)
+    tools are scoped by their target agent/cluster so an approval on one cluster
+    never leaks to another; local tools use the caller scope. Remoteness is the
+    tool's own `is_remote` flag — never inferred from the tool name."""
+    if is_remote:
```
naomi-robusta, line 135: you don't need this if and you don't need the is_remote field.
Just check if the tool_params have agent_name. If they do, use it.

naomi-robusta (changes requested):

holmes/core/tool_calling_llm.py
Avi-Robusta: Relying on an agent_name param is less reliable than an explicit bool — params can be anything, so any toolset or MCP server with a tool that happens to have an agent_name param would get misclassified as remote (cluster-scoped approvals + the "remote cluster" suffix) despite being purely local. 
is_remote is set once at load from relay's actual signal, so a param-name collision can't change the behavior.

naomi-robusta (approved):

PR #2321, opened 2026-07-26
fix(tools): apply approval_required_tools to YAML-defined toolsets

### LINKED ISSUE #2312: approval_required_tools does not work with custom YAML toolsets
### What happened?

I configured a specific tool in a custom YAML toolset to require approval (with the `approval_required_tools` fields) but the tool executes without asking for approval.

Tools specified in approval_required_tools don't execute automatically and wait for approval.

Write a simple custom toolset and specify a tool to require approval:

  ```yaml
  toolsets:
    custom_demo:
      description: Custom toolset
      approval_required_tools:
        - sensitive_action
      tools:
        - name: sensitive_action
          description: Sensitive action
          command: echo hello
```
Ask holmesgpt to call it.

`approval_required_tools` is not currently documented for custom toolsets, so it is unclear whether this is unsupported or an implementation bug.
If custom toolsets are intended to support it, the behavior should be fixed and documented. Otherwise, configuration validation should reject the field or clearly document the limitation.

Sanjays2402:
Closes #2312

`Tool._check_approval_config()` resolves `approval_required_tools` through a `tool.toolset` back-reference. Toolsets built in Python set it themselves (`Tool(toolset=self)`), but YAML/config-loaded toolsets construct their tools from dicts, so the back-reference stayed `None` and the patterns were silently ignored — the tool ran without ever prompting for approval.

`Toolset` now sets the back-reference on its own tools in an after-validator, leaving an explicitly-provided `toolset` untouched. Regression test added to `tests/test_approval_required_tools.py`; it fails on master and passes with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Approval requirements now apply correctly to tools defined through YAML or configuration files.
  * Tools requiring approval are blocked until approval is granted, while unrestricted tools continue to run normally.

* **Tests**
  * Added coverage to verify approval enforcement for YAML-defined toolsets.

### DIFF
--- a/holmes/core/tools.py
+++ b/holmes/core/tools.py
@@ -293,6 +293,9 @@ class Tool(ABC, BaseModel):
         description="The URL of the icon for the tool, if None will get toolset icon",
     )
     transformers: Optional[List[Transformer]] = None
+    # Back-reference to the owning toolset. Read by _check_approval_config() to
+    # resolve `approval_required_tools`; set by Toolset.link_tools_to_toolset().
+    toolset: Optional[Any] = Field(default=None, exclude=True, repr=False)
 
     # Private attribute to store initialized transformer instances for performance
     _transformer_instances: Optional[List["BaseTransformer"]] = PrivateAttr(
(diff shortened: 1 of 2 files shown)


Sanjays2402:
Thanks for the review — addressed in 09f45d9:

- `Toolset.link_tools_to_toolset` now has the `-> "Toolset"` return annotation.
- The non-gated `safe_action` assertion now requires `StructuredToolResultStatus.SUCCESS` instead of just `!= APPROVAL_REQUIRED`.

One item I left alone: making `Tool.toolset` an internal strongly-typed field. That would mean converting it to a `PrivateAttr` and auditing every reader/writer across the codebase (it's read via `getattr(self, "toolset", None)` in `_check_approval_config` and subclasses may touch it too), which felt like a bigger redesign than this PR's scope. Happy to do it as a follow-up if you'd like.

PR #2330, opened 2026-07-29
fix(operator): install the holmes package into the operator image

### LINKED ISSUE #2326: bug: holmes-operator:0.37.0 image crash loops with ModuleNotFoundError: No module named 'holmes'
## Summary

The `robustadev/holmes-operator:0.37.0` image fails to start. The operator entrypoint imports the `holmes` package, but that package is not installed in the operator image, so every 0.37.0 operator deployment ends up in `CrashLoopBackOff`.

## Traceback

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/app/holmes_operator/operator.py", line 10, in <module>
    from holmes.common.env_vars import ENABLE_JSON_LOGS_FORMAT
ModuleNotFoundError: No module named 'holmes'
```

## Root cause

Two changes landed together in 0.37.0:

1. `holmes_operator/operator.py` now imports from the main package, for the new JSON log format support:

   ```python
   from holmes.common.env_vars import ENABLE_JSON_LOGS_FORMAT
   from holmes.utils.log import build_json_formatter
   ```

2. `Dockerfile.operator` installs dependencies with `poetry install --no-root` and then copies only the operator package:

   ```dockerfile
   RUN ... poetry install --no-interaction --no-ansi --no-root
   COPY holmes_operator/ ./holmes_operator/
   ENTRYPOINT ["python", "-m", "holmes_operator.operator"]
   ```

Because of `--no-root` plus copying only `holmes_operator/`, the `holmes` package is never present in the operator image. The import at 0.36.0 did not exist, which is why 0.36.0 starts fine.

## Reproduction

Deploy the 0.37.0 chart with `operator.enabled: true`, or simply run the image:

```
docker run --rm robustadev/holmes-operator:0.37.0
```

## Impact

Any deployment of chart 0.37.0 with the operator enabled loses `HealthCheck` / `ScheduledHealthCheck` / `TriggeredHealthCheck` reconciliation entirely. The main `holmes` server image at 0.37.0 is unaffected.

## Workaround

Pin the operator image back to 0.36.0 while keeping the server on 0.37.0:

```yaml
operator:
  enabled: true
  image: holmes-operator:0.36.0
```

The CRDs are identical between chart 0.36.0 and 0.37.0, and the 0.36.0 operator ignores the newly added `ENABLE_JSON_LOGS_FORMAT` env var, so it works with the 0.37.0 chart.

## Possible fixes

Either copy the `holmes/` package into the operator image and install the root project, or drop the dependency on the main package from the operator entrypoint by inlining the small env-var/log-formatter helpers.

RahilOp:
Fixes #2326

The operator entrypoint (`holmes_operator/operator.py`) imports shared helpers from the main `holmes` package:

- `holmes.common.env_vars.ENABLE_JSON_LOGS_FORMAT`
- `holmes.utils.log.build_json_formatter`

`Dockerfile.operator` installed dependencies with `poetry install --no-root` and then copied only `holmes_operator/`, so the `holmes` package was never present in the image. Starting `robustadev/holmes-operator:0.37.0` therefore fails with:

```text
ModuleNotFoundError: No module named 'holmes'
```

Changes:
- Copy `holmes/` into the operator build context before `poetry install`.
- Remove `--no-root` so `poetry install` also installs the `holmes` root package.
- Keep the existing `holmes_operator/` copy step after installation.

This keeps the operator image self-contained while reusing the existing JSON-logging/env-var helpers instead of inlining them.

## Summary by CodeRabbit

* **Chores**
  * Improved container build setup by including `README.md` in the image and installing the application package as part of the dependency installation.
  * Updated dependency tooling configuration to provide a more consistent, non-interactive runtime environment.

### DIFF
--- a/Dockerfile.operator
+++ b/Dockerfile.operator
@@ -12,13 +12,14 @@ ARG WHEEL_MIN_VERSION=0.46.2
 ARG SETUPTOOLS_MIN_VERSION=80.0.0
 
 COPY pyproject.toml poetry.lock* ./
+COPY holmes/ ./holmes/
 # Upgrade build tooling first so poetry and the package installs it runs use the
 # CVE-patched pip/setuptools/wheel.
 RUN pip install --no-cache-dir -U \
         "pip>=${PIP_MIN_VERSION}" "setuptools>=${SETUPTOOLS_MIN_VERSION}" "wheel>=${WHEEL_MIN_VERSION}" && \
     pip install --no-cache-dir poetry && \
     poetry config virtualenvs.create false && \
-    poetry install --no-interaction --no-ansi --no-root
+    poetry install --no-interaction --no-ansi
 
 COPY holmes_operator/ ./holmes_operator/
 
--- a/Dockerfile.operator
+++ b/Dockerfile.operator
@@ -12,6 +12,7 @@ ARG WHEEL_MIN_VERSION=0.46.2
 ARG SETUPTOOLS_MIN_VERSION=80.0.0
 
 COPY pyproject.toml poetry.lock* ./
+COPY README.md ./
 COPY holmes/ ./holmes/
 # Upgrade build tooling first so poetry and the package installs it runs use the
 # CVE-patched pip/setuptools/wheel.


moshemorad:
@RahilOp Thanks for the PR - We merged a fix for it yesterday.
We intentionally aren't copying holmes folder to the operator image to keep it small and without irrelevant dependencies.

PR #2333, opened 2026-07-29
feat(toolsets): code_execution toolset ("code mode") to cut tool-call token cost

naomi-robusta:
## Summary

Adds an **opt-in `code_execution` toolset** implementing "code mode": instead of the LLM emitting one tool call per agentic step (re-sending all prior results each turn), it writes **one Python script** that composes many read-only Holmes tools and **filters results in code**, so only what it `print()`s enters the model's context. This targets the biggest token sink in production — intermediate-result bloat on multi-step investigations (≥6-iteration requests are ~66% of spend).

Design doc: `docs/design/2026-07-28_holmes-code-mode.md` (included in this PR — design and implementation reviewed together). Ticket: [ROB-723](https://linear.app/robusta/issue/ROB-723).

## How it works

- `run_python_code(code, timeout?)` runs the script in a **subprocess**. A generated `holmes` object exposes one function per eligible tool; each call is relayed over a **unix-domain socket** back to the parent, which dispatches into the real `ToolExecutor`. **Credentials never leave the parent process** — the subprocess can only ask the parent to run an allow-listed tool.
- **Read-only surface only.** `is_core` toolsets, `bash`/`kubectl_run`, and any `approval_required_tools` are excluded from the generated client. Because a synchronous script can't pause for interactive approval, an `APPROVAL_REQUIRED` result at dispatch is **denied** (the model is told to call that tool directly) — mirroring the trust model in the remote-tool-execution design.
- Reuses the existing `ulimit` memory cap; oversized stdout flows through the loop's existing `spill_oversized_tool_result` guard. **Off by default** (feature-flagged via `enabled=False`).
- `ToolCallingLLM` gains a small **generic** `set_tool_executor` wiring hook (no code-mode-specific coupling).

## Testing

**Deterministic token-reduction proof** (`test_code_mode_token_reduction.py`) — measures the exact context-token cost of a tool's output, classic vs code mode, for identical data, using the **same formatter (`format_tool_result_data`) and tokenizer (`litellm.token_counter`, gpt-4o) the product uses to size/spill results**. No LLM, no cluster, no RNG:

| Sink | Classic (tokens in context) | Code mode | Reduction |
|---|---|---|---|
| Result filtering (6000-event log → "how many ERROR?") | 250,029 | 136 | **99.95%** |
| Call consolidation (8 sub-calls → 1 result) | 30,952 | 275 | **99.11%** |

(Correctness asserted too — code mode computes the right answer from the same data.)

**Unit / integration** — 21 tests (real subprocess, no mocking of the runner): eligibility (`is_core`/bash/approval-gated/self excluded), the token-saving property, and failure modes (syntax error, runtime exception, timeout, unavailable tool, erroring sub-tool, approval-gated sub-tool denied-not-paused, unwired executor, timeout clamping). No regressions in `test_tool_calling_llm.py` / `test_tool_executor.py` (47 passing).

**LLM evals** — two ask_holmes A/B evals (`toolsets_matrix`: `[default]` vs `[codemode]`), **4/4 passing in CI** on opus-4.6 ([run 30478931611](https://github.com/HolmesGPT/holmesgpt/actions/runs/30478931611)). These assert **correctness parity** (code mode doesn't regress the answer) — they are deliberately *not* the token-win proof: Holmes's server-side filters (`kubernetes_jq_query`, log `filter`) already keep k8s data out of context, and the live A/B run confirmed the model kept using those rather than code mode, so the small fixtures show no token delta. The deterministic test above is where the token win is proven. Run on demand in `eval-regression.yaml`.

## Toolset checklist

- [x] Unit tests (+ deterministic token-reduction proof)
- [x] `config_classes` defined (`CodeExecutionConfig`)
- [x] Prerequisite check (verifies `python3` availability)
- [x] ask_holmes **evals** — A/B `toolsets_matrix`, 4/4 passing in CI (correctness parity)
- [ ] **Documentation** page — follow-up (`docs/data-sources/builtin-toolsets/code-execution.md`)

## Notes / follow-ups

- **Observability**: sub-calls are recorded (name/status/timing) and summarized in the result today. Live per-sub-call SSE streaming to the UI is a follow-up requiring a loop-level streaming hook + a `parent_tool_call_id` field (see the design doc; frontend PR to follow).
- This is a first, conservative v1 (read-only, subprocess + `ulimit`, no language-level sandbox). Hardening the sandbox is a documented Future Goal.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WMU8v729xhwopf4QcrJcPU)_

### DIFF
--- a/docs/design/2026-07-28_holmes-code-mode.md
+++ b/docs/design/2026-07-28_holmes-code-mode.md
@@ -1,0 +1,431 @@
+| Authors | Naomi |
+| --- | --- |
+| Review Date | 28 / Jul / 2026 |
+| Participants | AI/Holmes team, Backend (relay), Frontend |
+
+Ticket: [ROB-723](https://linear.app/robusta/issue/ROB-723) — "Using Code mode
+instead of scripts may save our token cost for tool calls." Project: *Minimize
+Token Cost for us (and users)*.
+
+Related reading:
+- `2026-06-10_remote-tool-execution.md` (relay) — the `expose_remotely` /
+  `is_core` markers, the **pre-approved-only, no-approval-round-trip** trust
+  model, and the 1MB inline result cap. Code mode reuses all three.
+- `2025-10-23_holmes-event-driven-architecture.md` (relay) — the SSE event
+  contract (`start_tool_calling` / `tool_calling_result`) that carries tool
+  calls to the UI, and relay's role as a transparent pipe.
+
+# Introduction
+
+## Overview
+
+Holmes investigates by running an **agentic tool-calling loop**: the LLM emits
+one or more tool calls, Holmes executes them, appends each result to the
+conversation, and re-sends the whole growing history on the next step. Complex
+investigations run 10–55 tool calls across many steps, and because every step
+re-sends all prior tool results, input tokens grow super-linearly with
+iteration count.
+
+**Code mode** ([Anthropic](https://www.anthropic.com/engineering/code-execution-with-mcp),
+[Cloudflare](https://blog.cloudflare.com/code-mode/)) replaces "many small tool
+calls, one per LLM step" with "the LLM writes **one Python script** that composes
+many tool calls, runs it in a sandbox, and only the **filtered final result**
+returns to the model's context." Tool outputs the model never needs are
+filtered *inside the script* and never enter the context window at all.
+
+Solution: a new **`code_execution` toolset** in Holmes whose single tool takes a
+Python snippet, runs it in a subprocess pre-loaded with a **generated `holmes`
+client module** that exposes the current request's tools as callable Python
+functions. Each function dispatches back into the existing `ToolExecutor`, so
+tools behave identically whether called directly or from a script. Sub-calls
+still emit their own SSE events, so the UI shows the same per-tool cards it does
+today; the wrapping script does not become an opaque black box.
+
+Stakeholders: AI/Holmes team (the runtime + prompting), backend/relay (SSE
+passthrough — no code changes expected), frontend (one new event field to
+surface script + sub-calls).
+
+## Glossary
+
+- **Code mode** — the execution path where the LLM writes Python composing tool
+  calls, instead of emitting one function call per LLM step.
+- **`code_execution` toolset / `run_python_code` tool** — the new toolset and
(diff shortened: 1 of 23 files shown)


naomi-robusta:
The **Benchmark Master** and **Benchmark PR** checks both fail with the identical `RuntimeError: Cold start e2e benchmark failed` (`scripts/cli_performance_benchmark.py`). Since the **Benchmark Master** job runs the master baseline — which this PR's diff cannot affect — this is a pre-existing benchmark-infra failure (the cold-start e2e `holmes ask` needs an LLM key/model that isn't available to this run), not something introduced here.

This change adds a new opt-in, off-by-default toolset, so I'm not touching the benchmark. I'm watching the actual unit-test gate (`build-and-test`) and will drive that to green if anything there fails.

---

naomi-robusta:
Thanks for the review — all four findings were valid and are addressed in `a7954cc`:

1. **🔴 Subprocess inherited the full parent env (credential leak).** Correct and important — it contradicted the "credentials never leave the parent" contract. The subprocess env is now built from a minimal allowlist (`PATH`, locale vars, and the three `HOLMES_CODE_*` bridge vars) instead of `{**os.environ}`. Added a regression test (`test_subprocess_env_excludes_parent_secrets`) that sets a secret env var and asserts the script can't see it.
2. **🟠 `get_parameterized_one_liner` crashed on `{"code": null}`.** Fixed with `params.get("code") or ""`; added `test_null_code_one_liner_does_not_crash` (it runs outside the invoke try/except, so this matters).
3–4. **🟠 Shared, externally-mutable `_tool_executor` race across `with_executor()` clones.** `run_python_code` now resolves the executor from the **request-scoped** `ToolInvokeContext.tool_executor` (set by the loop in `_directly_invoke_tool_call`) rather than shared toolset state, so concurrent clones can't overwrite each other. `set_tool_executor` is retained only for rendering `llm_instructions`.

All 23 code-execution tests + the `test_holmes_sync_toolsets` / `test_tool_calling_llm` suites pass locally. The **Benchmark Master/PR** checks remain red on the pre-existing cold-start-e2e infra failure (reproduces on the master baseline, as noted above) — not related to these changes.

---

PR #2069, opened 2026-05-20
Skip cache_control_injection_points for Gemini models

moshemorad:
## Summary
Gemini models reject `GenerateContent` requests that combine `CachedContent` with `system_instruction`, `tools`, or `tool_config`. Since litellm's `cache_control_injection_points` hook produces exactly this combination, we now conditionally skip this parameter for Gemini routes while preserving it for other providers that benefit from prompt caching.

## Changes
- **Added `_is_gemini_route()` helper function** to detect Google AI Studio (`gemini/*`) and Vertex-AI Gemini models while excluding non-Gemini models hosted on Vertex (Claude, Llama, etc.)
- **Modified `DefaultLLM.completion()` method** to conditionally include `cache_control_injection_points` only for non-Gemini providers (Anthropic, OpenAI, Azure, Bedrock, etc.)
- **Added comprehensive test suite** (`test_llm_completion_cache_control.py`) with:
  - Parametrized tests for Gemini models verifying cache control is skipped
  - Parametrized tests for non-Gemini models verifying cache control is included
  - Regression test ensuring other parameters (messages, model, temperature) are still forwarded correctly for Gemini

## Implementation Details
The solution uses a provider-aware conditional that:
1. Checks if the model is a Gemini route using the new `_is_gemini_route()` function
2. Only adds `cache_control_injection_points` to the completion call kwargs for non-Gemini providers
3. Preserves all other parameters and behavior unchanged

This approach maintains prompt caching benefits for providers that support it while avoiding Gemini API incompatibilities.

https://claude.ai/code/session_01KBkU343twiKe2RDSLkNnHz

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cache-control behavior for Gemini models to improve compatibility across different AI providers.

* **Tests**
  * Added comprehensive test coverage for cache-control injection behavior across model providers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HolmesGPT/holmesgpt/pull/2069?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #2348, opened 2026-08-03
fix: skip prompt cache hints for Bedrock Nova and add per-model override

### LINKED ISSUE #2332: cachePoint parameter not allowed for AWS nova-pro model
### What happened?

When asking a follow up question, or a more complicated question so multiple calls are made, cache control is automatically enabled by Holmes. Which nova(-pro) doesn't support.

I would expect holmes to allow to configure cache control per model. And perhaps for other assumptions about parameters for models as well (like `temperature` isn't supported for opus 4.7).

Use nova-pro through bedrock as a model and ask: `look at https://github.com/HolmesGPT/holmesgpt/pull/2331 would it be ok to approve` (any random PR will do, it seems).
litellm will return a bad request error:
```
BedrockException - {"message":"The model returned the following errors: Malformed input request: #/messages/2/content/0: extraneous key [cachePoint] is not permitted, please reformat your input and try again."}
```

In PR #2069 an exception was made for inserting cache control points for gemini. Nova will need these as well, and it makes me think that more models might need it, so just adding exceptions to the code might not scale here.

MingyuLiNCU:
## Summary

Fixes #2332 — Bedrock Nova models fail with `extraneous key [cachePoint] is
not permitted` on every completion call, because the prompt-cache hint that
LiteLLM translates into the Converse `cachePoint` field was only skipped for
Gemini routes (#2069) and never for Nova.

## Root cause

`cache_control_injection_points` is injected unconditionally for every
non-Gemini route in `holmes/core/llm.py`:

```python
if not _is_gemini_route(litellm_model_name):
    cache_kwargs["cache_control_injection_points"] = [...]
```

On Bedrock, LiteLLM sends this hint as the Converse API's `cachePoint` field,
which the Amazon Nova family rejects. Bedrock Claude (which supports
`cachePoint`) was never affected, which is why the exception list stayed
Gemini-only.

## Changes

Two complementary parts, both in `holmes/core/llm.py`:

1. **A single capability check instead of per-model exceptions** —
   new `_supports_prompt_caching(litellm_model_name)` returns False for the
   Gemini routes (unchanged behavior) and for Bedrock `amazon.nova*` models.
   All route-specific decisions now live in one place, so the next
   incompatible model is a one-line addition rather than another `if` at the
   call site.

2. **An explicit per-model override** — `ModelEntry` gains a
   `cache_control: Optional[bool]` field. Users can force prompt caching
   on/off from `model_list.yaml`, covering models neither the defaults nor
   LiteLLM know about yet (same shape as the temperature issue raised for
   Bedrock Opus 4.7).

## Tests

Extended `tests/core/test_llm_completion_cache_control.py`:

- Bedrock Nova routes (`us.amazon.nova-pro/lite/micro-v1:0`, plus a regional
  variant) get **no** `cache_control_injection_points` kwarg
- Bedrock Claude and all other non-Gemini routes **still get it** (regression)
- `cache_control: true|false` override wins in both directions, including
  forcing the hint on for Nova
- `cache_control` never leaks from `ModelEntry`/`DefaultLLM.args` into the
  litellm call

All tests are offline (mocked `litellm.completion`), following the existing
pattern from the #2069 tests.

## Verification

- `pytest tests/core/test_llm_completion_cache_control.py` and the LLM-related
  regression suite: all pass locally

## Summary by CodeRabbit

* **New Features**
  * Added automatic cache-control handling based on the selected model route.
  * Added per-model settings to force or suppress cache-control hints.
  * Cache-control settings are handled internally and are not passed to the underlying service.

* **Bug Fixes**
  * Prevented unsupported cache hints from being sent to Gemini and Bedrock Nova routes.
  * Preserved cache hints for compatible routes to maintain expected caching behavior.


hsmade (approved):
Lgtm!

PR #2350, opened 2026-08-03
ROB-831 Add personal skills and a per-account name-collision hierarchy

alonelish:
Part of [ROB-831](https://linear.app/robusta/issue/ROB-831/support-personal-global-custom-skills-name-collision-hierarchy-backend). Depends on the robusta-storage PR (adds the `user_id` column, the owner-only RLS policies, the `get_personal_skills` RPCs, and the `HolmesCustomSkills` table).

Personal skills become a new per-user tier alongside global (Supabase) and custom (filesystem) skills, and same-named skills across tiers can resolve to a single winner per an account-configurable order.

**No relay change is needed.** The realtime DB-based chat already carries the end user's identity — the frontend writes `user_id` onto the `Conversations` row and the conversations worker resolves it into `ChatRequest.user_id`, which this PR consumes.

## Personal tier

Adds `SkillSource.PERSONAL` and loads the requesting user's personal skills into the per-request catalog.

Reads go through the `get_personal_skills` / `get_personal_skill_content` `SECURITY DEFINER` functions rather than selecting `HolmesRunbooks` directly. **This is the non-obvious part:** personal rows are owner-only under RLS and Holmes authenticates as its own account-member *service* user, so a plain select would match zero rows and personal skills would silently never load.

Loading is gated on an explicit end-user `user_id` from the request, so **server-initiated flows load no personal skills** — alert triage, triggered workflows, scheduled prompts. The DAL's own `user_id` is never used as a fallback, since it's a service identity that's identical on every request. In the conversations worker this reuses the already-resolved "user Holmes may act on behalf of", so a conversation that opted out via `metadata.oauth_enabled = false` is covered by the same mechanism that already keeps triggered workflows off their creator's OAuth tokens.

## Why personal skills aren't in the `fetch_skill` id list

The skills toolset is built **once** and cached in `_cached_tool_executor` with a key of `(tags, enable_all)` that ignores account and user, and the "must be one of …" id list is frozen at first request. Baking per-user skills in would freeze them and leak them across users.

So the two catalogs are handled differently:

- The catalog feeding the **prompt** is rebuilt per request and carries the personal tier + hierarchy resolution.
- `SkillsFetcher` resolves a personal skill **at invoke time** from the request's `user_id`, mirroring how per-user OAuth tools already work. A test covers two users sharing one cached toolset instance and confirms neither sees the other's skills.

## Name-collision hierarchy

Skills are compared by **normalized human name**. Remote and personal skills keep their UUID in `name` because that's what `fetch_skill` needs, so the human name is tracked separately in `display_name`. Order comes from `AccountSettings`, defaulting to `global > custom > personal`, with builtin always lowest.

**Defaults to OFF**, which preserves today's behaviour exactly: no cross-tier dedup happens at all. Resolution is deterministic and done in Python, so a shadowed skill never reaches the model rather than relying on the prompt to describe priority.

Filtering happens **before** dedup, so a higher-tier skill that doesn't apply to this request (e.g. a global skill scoped to another cluster) can't suppress an applicable lower-tier one. There's a test for that specifically — inverting it would silently drop the skill that should have run.

## Custom skills mirror

`sync_skills` mirrors each cluster's filesystem and builtin skills into `HolmesCustomSkills` so the UI can list them. The filesystem stays the source of truth; this is a display-only upsert plus a prune of names that disappeared, run at startup and on the periodic refresh.

It re-reads every cycle rather than gating on a change signal, because a ConfigMap remount can change skills with no toolset status change to detect. It also deliberately does **not** delete when the loaded list is empty: "no skills" is indistinguishable from a load failure, and wiping the UI's view on a transient error is worse than briefly stale rows.

## Verification

- **Full non-LLM suite: 2719 passed, 98 skipped, 0 failed** — no regressions.
- New `tests/plugins/skills/test_skill_hierarchy.py`: personal scoping per user; **no personal skills without a user_id** (the server-initiated guardrail); the DAL's `user_id` never used as a fallback; flag OFF keeping all same-named tiers; flag ON with default and reversed order; normalized-name collisions; unknown tiers ignored; and filter-before-dedup.
- New `SkillsFetcher` tests: personal resolution per user, **no leakage across users on one cached instance**, and no personal lookup at all without a user_id.
- Existing skill-loader tests still pass, so filesystem override behaviour is unchanged.

Also extracted the duplicated `runbook.instructions` normalization into `_extract_skill_instruction` and checked it against every case the inline version handled; it's incidentally more robust, since a missing `runbook` no longer raises.

## Trade-off worth knowing

Owner-only read is a **product/UI privacy boundary, not a hard defence** — the storage RPCs are member-callable with an arbitrary `user_id`, because they're guarded on account membership rather than a verified end-user session. Worth stating rather than implying stronger isolation than exists.

https://claude.ai/code/session_01TDZMuWPPwVEbVrwbbJvdtT

## Summary by CodeRabbit

* **New Features**
  * Added personal skills loaded per request and isolated by the requesting user.
  * Added configurable skill hierarchy and predictable duplicate-name handling.
  * Added synchronization of built-in and custom skills for improved visibility.

* **Bug Fixes**
  * Prevented personal skills from being shared across users.
  * Improved handling of missing, malformed, or unavailable skill data without interrupting requests or startup.
  * Added safe fallbacks when hierarchy settings or skill content cannot be loaded.


holmes/utils/holmes_sync_skills.py
```diff
@@ -0,0 +1,71 @@
...
+                "cluster_id": config.cluster_name,
+                "skill_name": skill.name,
+                "source": SOURCE_LABELS.get(skill.source, skill.source.value),
+                "description": skill.description,
+                "content": skill.content,
+                "source_path": skill.source_path,
+                # Skills that fail to parse are logged and skipped by the loader, so
+                # everything reaching here parsed cleanly.
+                "status": "ok",
+                "error": None,
```
naomi-robusta, line 62: this is the only status you write. making the status and error irrelevant. it's always successful. parse errors are dropped and do not create a skill

naomi-robusta:
The description mentions calling RPCs but you drop them in the storage pr. What is the correct approach? Using RPCs or reading directly from the tables?

holmes/core/supabase_dal.py
```diff
@@ -606,29 +643,242 @@ def get_skill_content(self, skill_id: str) -> Optional[RobustaSkillInstruction]:
...
+        Scoped by user_id so one user cannot fetch another user's personal skill content --
+        the RLS policy admits Holmes for every personal row in the account, so this filter is
+        what keeps one user's fetch from reaching another's skill.
+        """
+        if not self.enabled or not user_id:
+            return None
+
+        try:
+            res = (
+                self.client.table(RUNBOOKS_TABLE)
```
naomi-robusta, line 800: shouldn't you check by enabled=true as well here? it's selected, but not used anywhere.

naomi-robusta (dismissed):

holmes/core/supabase_dal.py
alonelish: Fixed

holmes/utils/holmes_sync_skills.py
alonelish: Good catch - addressed in #2353, where the loader-side failure data lives. Parse failures now get a real `status="error"` row with the message, keyed on the same skill name the successful row would have used. As a side effect it also fixes pruning: one malformed file used to suppress the prune for the whole cluster.

alonelish:
RPC was removed since we call the table directly.

naomi-robusta (approved):

PR #2360, opened 2026-08-06
feat(helm): support documented config.model for default model

### LINKED ISSUE #2223: Helm chart does not support documented `config.model` for setting default model
## Summary

The official documentation tells users they can set a default model via Helm values like this:

```yaml
config:
  model: "gemini-flash"   # key from modelList
```

This does **not** work with the `helm/holmes` chart.

## Documentation

Multiple provider pages document `config.model` under the Helm tabs:

- https://holmesgpt.dev/latest/ai-providers/gemini/#configuration
- Similar blocks exist for other providers (OpenRouter, etc.)

The source is in `docs/ai-providers/gemini.md` (and equivalents).

Example from the "Holmes Helm Chart" tab:

```yaml
modelList:
  gemini-flash:
    api_key: "{{ env.GEMINI_API_KEY }}"
    model: gemini/gemini-1.5-flash

  gemini-pro:
    ...

# Optional: Set default model (use modelList key name)
config:
  model: "gemini-flash"
```

The same example is shown for the nested `holmes:` structure used when Holmes is a subchart.

## What the chart actually implements

- `values.yaml` only contains `modelList: {}`. There is no top-level `config:` field.
- `helm/holmes/templates/toolset-config.yaml` only renders `modelList` into the ConfigMap at `/etc/holmes/config/model_list.yaml`.
- No template code reads `.Values.config` (or `.Values.holmes.config`) or produces a `config.yaml` that would set the `model` field.
- The pod restart checksum only includes `modelList`.
- Runtime (`server.py` → `Config.load_from_env()` / file) only picks up the default model from:
  - The `MODEL` environment variable, or
  - `~/.holmes/config.yaml` (never created by this chart)

## Expected vs Actual

**Expected:** Following the docs and adding `config.model` sets the default model used when no model is passed in a request.

**Actual:** The setting is silently ignored. Holmes falls back to the first entry in `modelList` (or requires an explicit model per call).

## Workaround that works today

```yaml
additionalEnvVars:
  - name: MODEL
    value: "gemini-flash"     # must be a key from modelList

modelList:
  gemini-flash:
    ...
```

## Suggested resolution

Please either:

1. Add support for `config:` (and `holmes.config:`) in the Helm chart so the documented syntax works, or
2. Update the Helm sections in the provider docs to show the working `MODEL` + `modelList` pattern and remove the non-functional `config.model` example.

Ideally the chart would expose a clean, first-class value (e.g. `defaultModel`) that is clearly supported.

Relevant files:
- Docs: `docs/ai-providers/gemini.md`
- Chart values: `helm/holmes/values.yaml`
- Rendering logic: `helm/holmes/templates/toolset-config.yaml`
- Config loading: `holmes/config.py`, `server.py`

Thanks!

RahilOp:
Fixes #2223

The provider docs show a `config.model` Helm value for setting the default model, but the chart never honored it. This PR wires it up by exporting the `MODEL` environment variable on the Holmes pod.

Changes:
- Added `config: {}` to `helm/holmes/values.yaml` with a usage comment.
- Added `holmes.config` and `holmes.config.model` helpers in `_helpers.tpl` that support both standalone `config.model` and subchart `holmes.config.model` value paths.
- In `templates/holmes.yaml`, emit a `MODEL` env var when `config.model` is set.
- Included the config value in the pod checksum annotation so changing the default model triggers a rollout.

No-op when `config.model` is unset. Users can continue using `additionalEnvVars` with `MODEL` if they prefer.

Verified:
- `helm lint helm/holmes` passes.
- `helm template` with `config.model` renders `MODEL=gemini-flash`.
- `helm template` without `config.model` omits the `MODEL` env var.
- `helm template` with subchart-style `holmes.config.model` also renders `MODEL` correctly.

## Summary by CodeRabbit

* **New Features**
  * Added Helm configuration support for selecting an optional default model.
  * Supports configuration when Holmes is deployed independently or as a subchart.
  * Automatically applies the selected model to the deployment when configured.
  * Supports injecting a remediation bearer token from a configured Secret.

* **Bug Fixes**
  * Deployment updates now correctly reflect changes to Holmes configuration and remediation authentication settings.

* **Updates**
  * Improved handling of optional Helm configuration values and documented model selection requirements.


RahilOp:
Addressed the CodeRabbit feedback: changed  to a block mapping so the commented  example is valid YAML when uncommented.

RahilOp:
Addressed the latest CodeRabbit feedback: added nil guards in the holmes.config.model helper so both .Values.config and .Values.holmes.config are checked before accessing .model.

RahilOp:
Hi maintainers,

The failing **License Compliance** check on this PR appears to be a false positive. This change is Helm-only and only modifies:

- helm/holmes/templates/_helpers.tpl
- helm/holmes/templates/holmes.yaml
- helm/holmes/values.yaml

No license files or third-party dependencies were added, removed, or modified.

I have rebased the branch on the latest master and all other checks (DCO, EasyCLA, Netlify preview, CodeRabbit) are passing. Could you please override or dismiss the License Compliance failure so this can move forward?

Thanks!

RahilOp:
Rebased onto the latest upstream/master. The License Compliance failure remains a false positive for this Helm-only change. Please dismiss/override it when convenient.

RahilOp:
🤖 Auto-rebased onto `upstream/master` to keep this PR current. A new CI run should start automatically.

PR #2369, opened 2026-08-09
fix(tools): bound script/command tool subprocesses with a timeout

### LINKED ISSUE #2365: Script-based toolset tools (kubernetes/core, etc.) can hang forever — subprocess.run() has no timeout
### What happened?

`Tool.__execute_subprocess()` in `holmes/core/tools.py` runs every `script:`-defined toolset tool via:

```python
result = subprocess.run(
    protected_cmd,
    shell=True,
    executable="/bin/bash",
    text=True,
    check=False,
    stdin=subprocess.DEVNULL,
    stdout=subprocess.PIPE,
    stderr=subprocess.STDOUT,
)
```

https://github.com/HolmesGPT/holmesgpt/blob/0.38.2/holmes/core/tools.py#L656-L670

There is no `timeout=` argument. `__invoke_script()` (same file, ~L630) writes the rendered script to a `tempfile.NamedTemporaryFile(suffix=".sh")` and runs it through this same function.

The built-in `kubernetes/core` toolset (`holmes/plugins/toolsets/kubernetes.yaml`) uses this path for `kubernetes_jq_query`, `kubernetes_tabular_query`, and `kubernetes_count` — each script shells out to `kubectl` (`kubectl get --raw ...` in a pagination loop, `kubectl get {{ kind }} --all-namespaces -o custom-columns=...`, `kubectl api-resources --no-headers`), with no `--request-timeout` on any of the `kubectl` invocations either.

If the underlying `kubectl` call stalls for any reason (in our case, an intermittent apiserver connection issue), `subprocess.run()` blocks forever and the process is never reaped. This does **not** require the Holmes pod to restart — a single long-lived pod just accumulates one permanently-blocked worker per stuck call over its uptime, since tool calls run frequently.

**Observed production impact:** on one node we found 10 stuck `kubectl` processes spawned by a Holmes pod that had been running continuously for ~2.4 days with 0 restarts, each consuming 600-700% CPU the entire time (confirmed via live CPU-time delta measurement and `strace`, not just idle/blocked — the Go runtime was actively spinning on a network read that never completed). Confirmed present, byte-identical `tools.py`, at both `0.38.0` and `0.38.2`.

Search of existing issues didn't turn up a prior report of this specific one, apologies if I missed it.

A shelled-out tool script (and any subprocess it spawns, e.g. `kubectl`) should be bounded — timeout after a reasonable duration and be killed, rather than block forever and accumulate.

1. Configure the `kubernetes/core` toolset and trigger a tool call that resolves to `kubernetes_jq_query`, `kubernetes_tabular_query`, or `kubernetes_count` (e.g. via `ask_holmes` or any investigation that uses these tools).
2. While that call is in flight, make the `kubectl` invocation inside the generated script stall — e.g. against a test apiserver that accepts the TCP connection but never completes the response, or by dropping packets mid-connection to a real apiserver (`iptables -A OUTPUT -p tcp --dport <apiserver-port> -j DROP` applied after the connection is established).
3. Observe that `holmes` never times out or kills the resulting `bash <tmp>.sh` / `kubectl` process tree — it stays alive indefinitely, and the calling tool-execution thread stays blocked in `subprocess.run()` for as long as the stall persists.

This isn't `kubernetes/core`-specific — `__execute_subprocess()` is the shared execution path for every `script:`-defined tool across every toolset (e.g. `kubevela.yaml` too), so any of them can hang the same way if the command they wrap stalls.

Suggested fix direction: pass `timeout=` to `subprocess.run()` in `__execute_subprocess()`, catch `subprocess.TimeoutExpired`, and return a clean tool-error result instead of hanging. One thing to get right: the direct child of `subprocess.run()` is `bash <tmp_script>`, and the offending command (`kubectl`, etc.) runs as *its* child — `subprocess.run(..., timeout=N)`'s own cleanup only signals the direct child, which can leave a grandchild (the actual stuck `kubectl`) running independently. Using `start_new_session=True` and killing the whole process group (`os.killpg`) on timeout, rather than relying on `subprocess.run`'s default cleanup, would be needed to reap the full tree.

Happy to help verify a fix against our environment if useful.

abhyudayareddy:
## Problem

`Tool.__execute_subprocess()` (`holmes/core/tools.py`) is the shared execution path for every `command:`/`script:` toolset tool, including the built-in `kubernetes/core` tools (`kubernetes_jq_query`, `kubernetes_tabular_query`, `kubernetes_count`, all of which shell out to `kubectl`). It ran the command via `subprocess.run()` with no `timeout=`.

If the underlying process stalls — e.g. `kubectl` hits an apiserver connection that accepts the TCP handshake but never responds — `subprocess.run()` blocks the calling thread forever and the child process is never reaped. This doesn't require a pod restart to snowball: a single long-lived Holmes pod accumulates one permanently-blocked worker (and orphaned `kubectl` process) per stuck call, and each pins significant CPU.

This is not specific to `kubernetes/core` — `__execute_subprocess()` is shared by every `script:`-defined tool across every toolset (e.g. `kubevela.yaml` too).

Fixes #2365.

## Fix

- `__execute_subprocess()` now runs the command via `subprocess.Popen(..., start_new_session=True)` + `process.communicate(timeout=TOOL_SUBPROCESS_TIMEOUT_SECONDS)` (new env var, default `300`s) instead of an unbounded `subprocess.run()`.
- On timeout, the **entire process group** is killed with `os.killpg(..., SIGKILL)`, not just the immediate `bash` child. This matters because the direct child of `Popen` is `bash <script>`, and the actual offending command (`kubectl`, etc.) runs as *its* child — killing only the direct child (what plain `subprocess.run(timeout=...)` + `.kill()` does) leaves that grandchild running independently, which is exactly the failure mode described in the issue. `start_new_session=True` puts the whole tree in its own process group so `killpg` reaps it all.
- The race where the process exits on its own right as we're about to signal it (`os.killpg` raising `ProcessLookupError`) is handled explicitly and does not propagate.
- On timeout the tool returns exit code `124` (matching GNU `timeout`) plus any output produced before the deadline, so the LLM sees a clear, actionable error instead of the call silently vanishing.
- Added `TOOL_SUBPROCESS_TIMEOUT_SECONDS` (default `300`) to `holmes/common/env_vars.py`, following the existing `TOOL_MEMORY_LIMIT_MB` pattern, and documented it in `docs/reference/environment-variables.md`.

## Changes

- `holmes/common/env_vars.py`: new `TOOL_SUBPROCESS_TIMEOUT_SECONDS` env var (default 300s).
- `holmes/core/tools.py`: `Tool.__execute_subprocess()` rewritten around `Popen` + `communicate(timeout=...)`; new `Tool.__kill_process_group_and_collect_output()` helper that kills the process group and returns whatever output was buffered.
- `docs/reference/environment-variables.md`: reference entry for the new env var.
- `tests/core/test_tool_subprocess_timeout.py`: new tests (see below).

## Test plan

Added `tests/core/test_tool_subprocess_timeout.py`, covering:
- Fast commands still complete normally and return exit code 0.
- A command exceeding the timeout is killed and returns exit code 124 without blocking anywhere near the original duration.
- Output produced before the timeout is preserved alongside the timeout notice.
- **The specific bug scenario**: a command that backgrounds a child process (simulating a stalled `kubectl` grandchild) — the regression test confirms that process is actually killed, not left orphaned.
- The `os.killpg` race (process exits right as we try to signal it) is swallowed rather than raised.

Validation performed in this environment:
- `ast.parse` on all changed Python files (syntax-valid).
- `ruff check` and `ruff format --diff` with the pinned versions from `.pre-commit-config.yaml` (`ruff==0.7.2`) — clean on all lines touched by this change.
- `isort --profile black --check-only --diff` (`isort==6.1.0`, closest available to the pinned `7.0.0` for this environment's Python) — the new import is correctly positioned; no changes requested to lines touched by this PR.
- `mypy` against `holmes/core/tools.py` / `holmes/common/env_vars.py` — no new errors introduced by this change (pre-existing unrelated errors elsewhere in the file/codebase are untouched by this diff).
- This sandbox only has Python 3.9 available and the project requires 3.10+ (`litellm`/`aiohttp` enforce this at install time), so I could not run `pytest`/`make test-without-llm` directly here. To compensate, I extracted the exact `Popen`/`communicate`/`killpg` logic added in this PR into a standalone script (no holmes imports) and exercised it directly — confirmed: fast commands return normally, slow commands are killed at the timeout without blocking, partial output is preserved, and critically, the backgrounded "grandchild" process is killed with the new process-group approach while a side-by-side run of the *old* `subprocess.run(timeout=...)` pattern left that same grandchild process alive, reproducing the exact bug from the issue. CI will run the full suite including the new test file on a proper Python 3.10+ environment.

## Example

```bash
export TOOL_SUBPROCESS_TIMEOUT_SECONDS=600  # default is 300
```

## Summary by CodeRabbit

* **New Features**
  * Added configurable timeouts for subprocess-based commands and scripts, defaulting to 300 seconds.
  * Timed-out processes now stop cleanly, including child processes, and return exit code 124.
  * Partial output is preserved and accompanied by a timeout notice.

* **Documentation**
  * Added configuration guidance, default behavior, timeout handling, and examples for the new setting.

* **Bug Fixes**
  * Improved handling of subprocess termination races and timeout results.

### DIFF
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -249,6 +249,17 @@ See [Tool Execution Safety](../data-sources/tool-execution-safety.md) for the fu
 export TOOL_MEMORY_LIMIT_MB=2000
 ```
 
+### TOOL_SUBPROCESS_TIMEOUT_SECONDS
+**Default:** `300`
+
+Wall-clock timeout (in seconds) for every `command:`/`script:` toolset tool Holmes runs as a subprocess (e.g. `kubectl`, `jq`). If the command has not finished by the deadline, Holmes kills the entire process group it spawned — not just the immediate shell — so a stalled grandchild process (for example a `kubectl` call stuck on a hung apiserver connection) cannot be left running forever. On timeout, the tool returns exit code `124` (matching GNU `timeout`) with any output produced so far plus a note that the command was killed.
+
+**Example:**
+```bash
+# Allow long-running tool commands up to 10 minutes before Holmes kills them
+export TOOL_SUBPROCESS_TIMEOUT_SECONDS=600
+```
+
 ## HolmesGPT Configuration
 
 ### MODEL_LIST_FILE_LOCATION
--- a/holmes/common/env_vars.py
+++ b/holmes/common/env_vars.py
@@ -98,6 +98,15 @@ TOOL_MEMORY_LIMIT_MB = int(
     os.environ.get("TOOL_MEMORY_LIMIT_MB", _default_memory_limit)
 )
 
+# Wall-clock timeout (seconds) for `command:`/`script:` toolset tools run via
+# subprocess (e.g. kubectl, jq). Without this, a stalled child process (e.g. an
+# apiserver connection that accepts the TCP handshake but never responds) blocks
+# the calling thread and leaks a process forever. On timeout the entire process
+# group spawned for the command is killed, not just the immediate shell child.
+TOOL_SUBPROCESS_TIMEOUT_SECONDS = int(
+    os.environ.get("TOOL_SUBPROCESS_TIMEOUT_SECONDS", 300)
+)
+
 STREAM_CHUNKS_PER_PARSE = int(
     os.environ.get("STREAM_CHUNKS_PER_PARSE", 80)
 )  # Empirical value with 6~ parsing calls. Consider using larger value if LLM response is long as to reduce markdown to section calls.
--- a/holmes/core/tools.py
+++ b/holmes/core/tools.py
@@ -4,6 +4,7 @@ import logging
 import os
 import re
 import shlex
+import signal
 import subprocess
 import tempfile
 import threading
@@ -40,6 +41,7 @@ from pydantic import (
 from rich.console import Console
 from rich.table import Table
 
+from holmes.common.env_vars import TOOL_SUBPROCESS_TIMEOUT_SECONDS
 from holmes.core.llm import LLM
 from holmes.core.openai_formatting import format_tool_to_open_ai_standard
 from holmes.core.transformers import (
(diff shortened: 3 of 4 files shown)


abhyudayareddy:
cc @aantn — you're the primary author of this file, so thought you'd be the right person to take a look whenever you have time. CI is green aside from a pre-existing License Compliance baseline finding unrelated to this change (this diff adds no new dependencies). Thanks!

abhyudayareddy:
cc @aantn — checking back in on this one. DCO is green; the tool subprocess timeout fix is still ready for review whenever you have a chance. Thanks!

abhyudayareddy:
Rebased onto latest `master` (was ~29 commits behind) — clean, no conflicts. `pytest tests/core/test_tool_subprocess_timeout.py` → 7 passed.

The CodeRabbit "kill the process group before raising" note was already handled in `b936a64` (the `killpg_then_raise` helper calls the real `os.killpg` first), and that thread is resolved. Still ready for review whenever you have a chance, @aantn.

abhyudayareddy:
cc @aantn — checking back in on this one after a couple weeks. The tool subprocess timeout fix is still ready for review; DCO is green and the branch is current with master. Let me know if there's anything I can do to help move it forward. Thanks!

PR #2371, opened 2026-08-09
ROB-855 pulling skills with github app credentials

Avi-Robusta:
## What

Documents how to load skills from a private GitHub repo using **GitHub App credentials** instead of a Personal Access Token. The init container mints a JWT from the App's private key, exchanges it for a short-lived installation token (1 hour), and clones with that — no long-lived credential ever reaches the pod.

Covers the Holmes Helm Chart, Robusta Helm Chart, and Holmes CLI.

## Verified against a live cluster

This was tested end-to-end, not just written:

- The documented JWT generation (`openssl` RS256 + base64url) and installation-token exchange were run verbatim — GitHub accepted the JWT and returned a real `ghs_` token.
- The init container recipe was applied to a running Holmes deployment, which cloned a **private** repo successfully and came up healthy.
- Holmes registered all 4 skills from the repo as `source=user`. Confirmed via `load_skill_catalog()` inside the pod, and visible in the UI.

The test repo's skills are a completely different set from the ones previously mounted from Bitbucket, so this can't be a stale-mount false positive.

## Two fixes that came out of testing

- **Repo and branch are now env vars** (`GIT_REPO`, `GIT_BRANCH`). They were previously buried inside the shell script, so adopting the recipe meant hand-editing the script body. Now everything install-specific lives in one `env:` block and the script can be copied verbatim.
- **`--branch main` is a silent trap.** The clone fails outright when a repo still defaults to `master` — which is exactly what happened on the repo used for testing. Now called out explicitly in the adjust notes.

Both were applied to the Personal Access Token tabs too, which had the same hardcoded values and the same branch-name trap.

## Note for reviewers

The docs state the App needs only **Contents: Read-only**. The App used for testing (`holmes-mcp-integration`) carries considerably more — `contents: write`, `issues: write`, `pull_requests: write`, `statuses: write`. It works, but anyone copying this setup should scope a skills-only App down to read-only rather than reusing that one.

## Summary by CodeRabbit

- **Documentation**
  - Added configuration guidance for customizing Git repository and branch settings in Helm deployments.
  - Documented GitHub App authentication using short-lived installation tokens, including required permissions, secrets, egress, and CLI usage.
  - Added instructions for token generation, validation, and repository cloning.
  - Clarified public-repository setup by removing the unnecessary personal access token configuration.
  - Applied the same guidance across supported Helm deployment variants.

### DIFF
--- a/docs/reference/skills.md
+++ b/docs/reference/skills.md
@@ -32,7 +32,7 @@ For private repos there are two ways to authenticate: a fine-grained [Personal A
       --from-literal=token='<PAT>'
     ```
 
-    For a public repo, omit the Secret and drop the `oauth2:${GIT_PAT}@` segment from the clone URL below.
+    For a public repo, omit the Secret, delete the `GIT_PAT` entry from the `env:` block, and drop the `oauth2:${GIT_PAT}@` segment from the clone URL below.
 
     **2. Add the init container, volume, and skill path to your Helm values:**
 
@@ -64,7 +64,7 @@ For private repos there are two ways to authenticate: a fine-grained [Personal A
         args:
           - |
             set -e
-            rm -rf /skills-repo/.git /skills-repo/* 2>/dev/null || true
+            rm -rf /skills-repo/* /skills-repo/.[!.]* /skills-repo/..?* 2>/dev/null || true
             git clone --depth 1 --branch "$GIT_BRANCH" \
               "https://oauth2:${GIT_PAT}@${GIT_REPO}" \
               /skills-repo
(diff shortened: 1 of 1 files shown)


Avi-Robusta:
Addressed 5 of the 6 CodeRabbit comments in 1cedd7b0b — the false "no long-lived credential" claim, hidden-file cleanup before re-clone, the self-contradictory public-repo note, and the CLI example (fail-closed on empty token, configurable repo/branch, shallow clone).

The remaining one — use a credential helper so tokens are never written to `.git/config` — is **out of scope for this docs PR**. It's a valid concern, but patching it here means adding a credential-helper dance to all 6 copy-paste examples, which makes an already-long shell script harder to adopt while still leaving the underlying problem: users shouldn't be hand-maintaining ~25 lines of JWT-signing shell to load skills from a repo.

The real fix is for Holmes to pull skills from a Git repo natively, handling auth and refresh internally. Filed as #2372, which covers the credential-in-`.git/config` issue properly along with the other limitations of the init-container approach (refresh only on pod restart, `apk add` egress requirement, 6 near-duplicate examples to keep in sync).

moshemorad (approved):

PR #2383, opened 2026-08-13
ROB-889 Refresh the realtime JWT before it expires

Avi-Robusta:
## Summary

Found while going through 48h of production Holmes logs. Every hour, on the hour, the log carries a `Realtime channel unhealthy (ChannelStates.CLOSED), reconnecting` line — 48 of them in 48h, plus the matching re-sign-ins, a few bare `WebSocket connection closed with code: 1006`, and one smoking gun:

```
Broadcast subscribe status=RealtimeSubscribeStates.CHANNEL_ERROR
  err={"reason": "InvalidJWTToken: Token has expired 296 seconds ago"}
```

## Root cause

`RealtimeWorker._maybe_refresh_auth()` ticks every 60s but only re-pushes the JWT when the token **string** changed:

```python
new_jwt = session.access_token
if not new_jwt or new_jwt == self._last_auth_jwt:
    return
```

It never inspects `exp`, and nothing else refreshes the token on a realtime-only path — the DAL refreshes reactively, from `patch_postgrest_execute` on a PGRST301/expired error, and an idle worker issues no postgrest queries to trip it. So the same 1-hour token gets re-sent every tick until Supabase closes the socket.

Recovery is reactive but does work: `_channel_unhealthy()` notices within ~5s and `_full_reconnect()` forces a real `sign_in()`. Hence the clean hourly sawtooth. The cost is an hourly connection drop plus a multi-minute window running an already-dead token, during which inbound broadcasts are dropped — only the much slower claim poll catches that work.

## Changes

- `_maybe_refresh_auth()` now decodes the token's `exp` and proactively re-signs-in when it falls within a new `CONVERSATION_WORKER_AUTH_REFRESH_LEEWAY_SECONDS` (default 300 — comfortably above the 60s refresh interval, so a tick always lands inside the window). The existing "token rotated elsewhere" branch is unchanged.
- The re-sign-in reuses `_full_reconnect`'s bounded-`sign_in` pattern (`asyncio.to_thread` + `_RECONNECT_SIGN_IN_TIMEOUT_SECONDS`), so a half-open connection can't stall the loop and take health checks down with it.
- Signature verification is deliberately off in the decode: it's our own token and `exp` is the only claim needed. An undecodable or `exp`-less token returns False and falls through to the pre-existing unhealthy → `_full_reconnect` safety net rather than re-signing-in every tick against a token we can't reason about.
- `_proactive_refresh_attempted_for` bounds us to one attempt per distinct token, so a `sign_in` that keeps handing back an expiring token can't spin.
- Log the first reconnect of a run at INFO instead of WARNING. A single self-healing reconnect is routine; only one that didn't take last time round (`reconnect_attempts > 0`) says something is actually wrong.

## Tests

11 new tests in `tests/core/conversations_worker/test_realtime_manager.py`: near-expiry and already-expired detection, fresh tokens left alone, unreadable/`exp`-less tokens, externally-rotated tokens still pushed, one-attempt-per-token, sign-in failure swallowed, hanging sign-in bounded, and the INFO-then-WARNING reconnect levels.

- `poetry run pytest tests/core/conversations_worker/` → 186 passed, 22 skipped
- `poetry run pytest tests -m "not llm"` → 2757 passed, 98 skipped

## Not in this PR

The same logs also show 28 `RemoteProtocolError: Server disconnected` errors from `claim_tool_calls`. Those are **already fixed** — `SupabaseRetryTransport` landed in 0.34.0 and is in every release since. The pod producing these logs is on 0.33.0 (pinned by the bare `"Supabase error while claiming tool calls"` message, which exists only in that release, plus HTTP/2 frames in the traceback). That one needs a deploy, not a patch.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Si7bLvwGMiRZTWuzFyrYUu)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved realtime connection reliability by refreshing authentication before tokens expire.
  * Detects externally updated authentication tokens and uses the latest valid token.
  * Handles authentication refresh failures more gracefully.
  * Reduced noise in connection recovery logs while preserving warnings for repeated reconnect attempts.


naomi-robusta (approved):
This change might be necessary in relay & frontend repos

PR #2390, opened 2026-08-13
ROB-795: retry the startup Robusta models fetch

avshyz:
A transient relay/gateway blip (e.g. a 502 during a relay deploy) at pod startup used to permanently degrade the agent: the model-list fetch was a single attempt whose failure was swallowed, leaving only the legacy "Robusta" fallback until the next restart.

Retry failures that can heal - connection errors, timeouts, 429/5xx - up to FETCH_ROBUSTA_MODELS_ATTEMPTS times (default 5, env-tunable) with capped exponential backoff (2/4/8/10s). 4xx is never retried: a bad token cannot heal, and boot is blocked while retrying (the liveness probe kills the pod after ~130s, which caps how patient startup may be).

Claude-Session: https://claude.ai/code/session_01GJGQ6tq865vm94GdkUCfge

## Summary by CodeRabbit

* **New Features**
  * Robusta model catalogs now refresh automatically during periodic status updates.
  * Refreshes detect added, removed, and default-model changes while preserving custom models.
* **Bug Fixes**
  * Improved reliability when fetching Robusta models by retrying transient network and server errors with exponential backoff.
  * Preserved fallback behavior when refreshes fail.
  * Reduced repetitive error logging during repeated refresh failures.
* **Tests**
  * Added coverage for model refresh scenarios and retry behavior, including recovery and non-retryable errors.

### DIFF
--- a/holmes/clients/robusta_client.py
+++ b/holmes/clients/robusta_client.py
@@ -4,11 +4,27 @@ from typing import Any, Dict, Optional
 
 import requests  # type: ignore
 from pydantic import BaseModel, ConfigDict
+from tenacity import (
+    before_sleep_log,
+    retry,
+    retry_if_exception,
+    stop_after_attempt,
+    wait_exponential,
+)
 
-from holmes.common.env_vars import ROBUSTA_API_ENDPOINT
+from holmes.common.env_vars import (
+    FETCH_ROBUSTA_MODELS_ATTEMPTS,
+    ROBUSTA_API_ENDPOINT,
+)
 
 HOLMES_GET_INFO_URL = f"{ROBUSTA_API_ENDPOINT}/api/holmes/get_info"
 TIMEOUT = 0.5
+
+# 429/5xx (gateway blips, overload) heal on retry; 4xx (bad token, unknown
+# account) doesn't, and retrying it would only delay startup.
+_RETRYABLE_STATUS_CODES = {429, 500, 502, 503, 504}
+
+logger = logging.getLogger(__name__)
 
 
 class HolmesInfo(BaseModel):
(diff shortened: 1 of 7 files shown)


holmes/clients/robusta_client.py
```diff
@@ -27,19 +41,44 @@ class RobustaModelsResponse(BaseModel):
...
+        exc, (requests.exceptions.ConnectionError, requests.exceptions.Timeout)
+    )
+
+
+# The model list is fetched once, at startup: losing that single request to a
+# transient relay/gateway blip degrades the agent to the legacy single-model
+# fallback for the pod's whole life (ROB-795). Retries stay bounded because
+# they block boot — the liveness probe kills the pod after ~130s without a
+# served /healthz.
+@retry(
```
Avi-Robusta, line 75: what do the logs look like? we dont want to spam the user with 10 errors, we should make the first one a warning and print the error every 3-5 IMO.
we dont want every one to print an error

Avi-Robusta (changes requested):
small check  with logging

holmes/clients/robusta_client.py
Avi-Robusta: i guess it would be in server.py

Avi-Robusta: we have that logic somewhere in holmes with backoff logging, see if we can reuse that code too

Avi-Robusta (dismissed):

server.py
```diff
@@ -285,6 +285,15 @@ def refresh_loop():
                 )
 
             time.sleep(sleep_time)
+            try:
+                # Re-read the Robusta model catalog before publishing status, so
+                # the heartbeat advertises what the agent can serve right now:
+                # catalog changes reach a running agent, and an agent that
+                # booted without a catalog heals, without a pod restart.
+                if dal.enabled:
+                    config.llm_model_registry.refresh_robusta_models()
```
moshemorad: Can we move it to after the Heartbeat? i'm afraid that this might effect the time to send the heartbeat and can cause issues in case of multiple retries. WDYT?

holmes/core/llm.py
```diff
@@ -1014,9 +1141,14 @@ def get_model_params(self, model_key: Optional[str] = None) -> ModelEntry:
...
-                    logging.warning("Resyncing Registry and Robusta models.")
-                    self._init_models()
+                # The catalog may have gained this model since it was last read.
+                # Any name is worth a look: a customer's models carry their own
+                # prefix, so keying on `Robusta/` left exactly those unable to
+                # recover. Unlike _init_models, refreshing never rebuilds from
+                # the model file and never falls back to the legacy entry, so a
+                # fetch that fails here cannot cost us the models we do have.
+                logging.warning(f"Model {model_key} is not loaded; refreshing.")
+                if self.refresh_robusta_models():
```
Avi-Robusta: This runs inside _lock. If the relay is unreachable, the bounded retries and 10s timeouts mean the fetch can hold the lock for ~70s — and since every read path in get_model_params sits under that same lock, any concurrent caller blocks for that whole window, including ones asking for non-Robusta models.

Avi-Robusta (changes requested):

Avi-Robusta (approved):

PR #2407, opened 2026-08-20
Add flyio/core built-in toolset for Fly.io investigation

tsushanth:
## What

Adds a `flyio/core` built-in toolset wrapping `flyctl`, the same pattern
`docker.yaml` uses for the Docker CLI: app status, recent logs (bounded,
not tailing), machine list/status, health checks, release history, scale
config, and secret *names* (never values).

## Why

The docs say "no Kubernetes required — works with any infrastructure,"
which is true of the CLI itself but not of the built-in toolsets: of the
~25 shipped today, all but a handful (`internet`, `bash`,
`core_investigation`, `skills`, `connectivity_check`) are
Kubernetes/cloud-native-specific (`kubernetes/*`, `openshift/*`, `aks/*`,
`cilium`, `argocd`, `kubevela`, ...). There's currently no adapter for a
PaaS like Fly.io, Render, or Railway — a team running a plain VM/PaaS
stack has almost nothing to plug into out of the box. This fills that gap
for Fly.io specifically.

## Validation

Beyond the unit tests in this PR (toolset loads, exact tool set, no
destructive commands, prerequisite check, secret-values-never-exposed
pinned), this was validated against two real, live, actually-failing
production apps in different languages — not staged demos:

- **Node/Express app**: correctly diagnosed a TTS backend timeout as root
  cause, and surfaced two problems that weren't part of the original ask
  (a machine stopped since the prior day halving capacity, and an
  unrelated third-party API 403).
- **Python/aiohttp app**: correctly parsed multi-line stack traces
  (structurally different from the Node app's single-line errors),
  correctly distinguished a SIGABRT crash (exit code 134) from a requested
  stop (exit code -1), and again correctly identified the external
  WebSocket timeout as root cause.

Full transcripts, unedited:
https://github.com/tsushanth/holmesgpt-toolset-flyio/tree/main/docs

## Notes for reviewers

- All tools are read-only investigation commands; the test suite
  explicitly asserts no destructive substrings (`destroy`, `delete`,
  `secrets set`, `scale set`, `deploy`) appear in any command template.
- `docs_url` points at the `holmesgpt.dev/data-sources/builtin-toolsets/`
  convention used by other built-in toolsets — happy to also add the docs
  page itself if that's expected as part of this PR rather than a
  follow-up.
- Ran into the Python 3.14/prometrix incompatibility mentioned in
  CONTRIBUTING.md during local setup (unrelated to this change) — pinned
  to 3.12 via poetry to develop and test.

## Summary by CodeRabbit

* **New Features**
  * Added a Fly.io toolset with read-only access to app status, logs, machines, health checks, releases, scaling settings, and secret names.
  * Added Fly CLI availability checks before running Fly.io tools.

* **Tests**
  * Added coverage for toolset loading, app scoping, read-only restrictions, prerequisites, bounded log output, and secret-safe command handling.

### DIFF
--- a/holmes/plugins/toolsets/flyio.yaml
+++ b/holmes/plugins/toolsets/flyio.yaml
@@ -13,8 +13,8 @@ toolsets:
         command: "fly status -a {{ app }}"
 
       - name: "fly_logs_recent"
-        description: "Fetch recent logs for a Fly app without tailing (bounded, safe for investigation)"
-        command: "fly logs -a {{ app }} --no-tail"
+        description: "Fetch recent logs for a Fly app without tailing, capped to the last 500 lines to avoid unbounded output during a noisy incident"
+        command: "fly logs -a {{ app }} --no-tail | tail -n 500"
 
       - name: "fly_machine_list"
         description: "List all machines for a Fly app, including state and restart counts"
--- a/holmes/plugins/toolsets/flyio.yaml
+++ b/holmes/plugins/toolsets/flyio.yaml
@@ -1,0 +1,41 @@
+toolsets:
+  flyio/core:
+    description: "Read access to Fly.io apps, machines, logs, and releases — for investigating a service that runs on Fly rather than Kubernetes"
+    docs_url: "https://holmesgpt.dev/data-sources/builtin-toolsets/flyio/"
+    tags:
+      - cli
+    prerequisites:
+      - command: "fly version"
+
+    tools:
+      - name: "fly_status"
+        description: "Show the current status of a Fly app — machines, regions, health"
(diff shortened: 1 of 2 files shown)


tsushanth:
recheck

tsushanth:
recheck

tsushanth:
Re: the License Compliance check ("12 issues found") — this PR adds two files (a YAML toolset definition and a pytest test file) and zero new dependencies, so this is very likely a pre-existing repo-wide finding rather than something introduced here. Happy to look into it further if a maintainer can point me at the actual issue list (the check has no target_url from what I can see via the API).

Issue #2046, opened 2026-05-14
Feature Request: Investigation Path Completeness Validation Against Historical Incidents


saneroen:
## Feature Description

Add automatic validation at the end of investigations to compare the current investigation path against similar historical incidents, ensuring all relevant checks were performed.

## Use Case

When troubleshooting an incident (e.g., payment-service crash), Holmes should:

1. Complete the investigation as normal
2. Query historical incidents with similar symptoms
3. Compare the current investigation path against successful past investigations
4. Report missing checks: "⚠️ Previous similar incidents also checked: Redis metrics, Redis service endpoints"

## Example Scenario

**Current Investigation Path:**
- kubectl_describe pod/payment-service → CrashLoopBackOff
- kubectl_logs pod/payment-service → Redis connection timeout
- kubectl_describe pod/redis → Pod healthy

**Historical Similar Incident Path:**
- kubectl_describe pod/payment-service
- kubectl_logs pod/payment-service
- kubectl_describe pod/redis
- kubectl_get service/redis ← **Missing in current investigation**
- kubectl_describe endpoints/redis ← **Missing in current investigation**
- prometheus_query redis_connected_clients ← **Missing in current investigation**

**Expected Output:**

CAOShurong:
Thanks for the feature request. I did a current-master triage against `3d1da39a46bde48955237f575fb901fa0695629e`.

**What exists today**

- The investigation loop accumulates the tool results for one run and returns them as an `LLMResult`; the list is deduplicated by tool-call ID before the response is returned ([tool loop](https://github.com/HolmesGPT/holmesgpt/blob/3d1da39a46bde48955237f575fb901fa0695629e/holmes/core/tool_calling_llm.py#L633-L739)).
- The response model exposes the current run's `conversation_history` and `tool_calls`, but I could not find a generic persisted incident corpus or a cross-incident path-comparison service in the current tree ([response model](https://github.com/HolmesGPT/holmesgpt/blob/3d1da39a46bde48955237f575fb901fa0695629e/holmes/core/models.py#L351-L358)).
[...]

PR #2418, opened 2026-08-24
Feature/investigation path completeness

saneroen:
## Summary

Offline benchmark for #2046 — path schema, redacted fixture corpus, metrics, and a runnable eval.

**No runtime change.** `holmes/` is untouched outside the new `holmes/core/investigation_path/` package. Per the review: measure the retrieval policy before it talks to a user.

Covers the suggested first slice:

1. **Path event** (`schema.py`) — tool, entity, intent, time window, outcome/error class, evidence ref, ordinal. No credentials, no raw output, no unstable params. Matching runs on **intent**, not tool name, so `kubectl logs` and a Loki query count as the same check.
2. **Corpus** — 17 human-validated incidents (12 pool, 5 held out) with root-cause labels and weighted reference paths. Retrieval ranks on symptoms, then filters to the shared root cause. Abstains with one of 4 reasons when that fails.
3. **Offline eval** — weighted recall, precision, false-positive burden, abstention, ECE/Brier, latency, storage and token cost.
4. **Suggestions** carry source incident + date, the human rationale, and support as a fraction. Worded as advice, capped at 5. Confidence shown as a percentage only when a fitted calibration model produced it.
5. **Runs in the eval pipeline** — asserted on every PR under `-m "not llm"`; `eval-regression.yaml` appends its report to `evals_report.md` and logs to its own Braintrust experiment. No-ops without `BRAINTRUST_API_KEY`.

## Baseline (5 held-out cases)

```
weighted path recall        0.69   suggestion precision        1.00
  ... when answering        0.75   false positives per answer  0.00
abstention rate             0.20   expected calibration error  0.050
latency p95 (ms)            0.12   llm calls / tokens          0 / 0
```

Raw similarity scored ~0.32 for suggestions correct every time (ECE 0.350). Platt calibration, fitted leave-one-out on the pool only, brings it to 0.050 out-of-sample.

**These numbers are 9 suggestions over 5 hand-written cases** — "no problem detectable at this sample size", not "solved".

## Next steps

1. Grow the corpus — every number above is limited by sample size first.
2. Close the `HOLD-005` gap with a `<dependency>` role token.
3. Sweep the policy knobs, publish a risk-coverage curve.
4. **Wire into `holmes investigate`** — separate PR, off by default behind a config flag.

Happy to open a tracking issue for step 4 if you'd rather agree the runtime shape up front.

## Test plan

- [x] `pytest tests/core/investigation_path` — 221 pass
- [x] `pytest tests/core` — 916 pass, no regressions
- [x] `python -m holmes.core.investigation_path.offline_eval` — prints the baseline
- [x] `git diff master...HEAD -- holmes/ ':!holmes/core/investigation_path/'` is empty

Design doc: `docs/design/2026-08-24_investigation-path-completeness.md`

## Summary by CodeRabbit

* **New Features**
  * Added investigation-path analysis that identifies missing diagnostic checks and provides ranked, confidence-scored suggestions.
  * Added privacy-conscious normalization and incident-based retrieval with explicit abstention handling.
  * Added offline benchmarking with calibration, performance metrics, Markdown reports, and optional CI reporting.
  * Added a curated incident corpus with held-out benchmark cases.

* **Documentation**
  * Added design documentation covering the schema, evaluation approach, metrics, privacy safeguards, and limitations.

* **Tests**
  * Added comprehensive coverage for normalization, retrieval, validation, calibration, reporting, corpus loading, and offline evaluation.

### DIFF
--- a/.github/workflows/eval-regression.yaml
+++ b/.github/workflows/eval-regression.yaml
@@ -1342,6 +1342,38 @@ jobs:
             echo "duration=${DURATION}s" >> $GITHUB_OUTPUT
           fi
 
+      - name: Run investigation path benchmark
+        # Offline path-completeness benchmark for issue #2046. Deterministic, no
+        # LLM call, no cluster, a couple of seconds - it runs here so the
+        # retrieval policy accumulates a tracked history in the same place as
+        # the LLM evals before any user-facing behaviour depends on it.
+        #
+        # continue-on-error plus the exit-code guard below: this reports on a
+        # policy that nothing ships against yet, so it must never be the reason
+        # an eval run goes red or an eval report fails to post.
+        if: always() && steps.check-tests.outputs.should-run == 'true'
+        continue-on-error: true
+        working-directory: code
+        shell: bash
+        env:
+          BRAINTRUST_API_KEY: ${{ secrets.BRAINTRUST_API_KEY }}
+          EXPERIMENT_ID: github-${{ github.run_id }}.${{ github.run_number }}.${{ github.run_attempt }}
+        run: |
+          set +e
+          poetry run python -m holmes.core.investigation_path.offline_eval \
+            --braintrust --markdown path_benchmark.md
+          BENCHMARK_EXIT=$?
+          set -e
+
+          if [[ $BENCHMARK_EXIT -ne 0 ]]; then
+            echo "⚠️ investigation path benchmark exited $BENCHMARK_EXIT; not posting a report"
+            exit 0
+          fi
+
+          # Appended rather than merged into the eval table: these rows are
+          # deterministic and would be misread next to model-scored ones.
+          { echo; echo "---"; echo; cat path_benchmark.md; } >> evals_report.md
+
       - name: Post evaluation results
         if: always() && steps.check-tests.outputs.should-run == 'true' && steps.eval-params.outputs.pr_number != ''
         uses: actions/github-script@v7
--- a/docs/design/2026-08-24_investigation-path-completeness.md
+++ b/docs/design/2026-08-24_investigation-path-completeness.md
@@ -1,0 +1,188 @@
+# Investigation path completeness — schema, corpus and metrics
+
+Status: **offline only.** Nothing here is wired into an investigation. This is
+the benchmark groundwork for
+[issue #2046](https://github.com/HolmesGPT/holmesgpt/issues/2046), so that a
+retrieval policy can be measured before it is allowed to talk to a user.
+
+The idea in the issue is: after an investigation, compare what it checked
+against what similar resolved incidents checked, and point out anything it
+skipped. The risk is that a wrong suggestion during an incident costs a
+responder attention at the worst possible moment, and that borrowing one
+historical path teaches every future investigation to repeat it. So the first
+question is not "does it work" but "how would we know".
+
+Three artifacts answer that: a **path schema**, a **corpus**, and a **metrics
+definition** with a runnable offline eval.
+
+## 1. Canonical path event
+
+An investigation becomes an ordered list of `PathEvent` (see
+`holmes/core/investigation_path/schema.py`):
+
+| field | purpose |
+|---|---|
+| `ordinal` | position in the investigation, so order is preserved |
+| `tool` | tool as executed — provenance only, never matched on |
+| `intent` | why the check ran: `describe`, `logs`, `metrics`, `topology`, … |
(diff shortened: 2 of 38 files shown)


CAOShurong:
{
"body": "Exact-head verification of the benchmark slice, following up on the #2046 triage. Everything below was run locally at head `ff5f8a1e` (`Close two gaps against the #2046 review: token cost and shown confidence`), base `fd8dd5e0`.\n\n**Environment:** Windows 10, CPython 3.13.1, pydantic 2.x, PyYAML 6, pytest installed ad hoc (not the project poetry lock — see the caveat at the end).\n\n## What I ran\n\n```bash\ngit checkout --detach ff5f8a1e3ffdbeb59666a84345e58ff0004346b6\npython -m pytest tests/core/investigation_path/\npython -m holmes.core.investigation_path.offline_eval\n```\n\nResults:\n\n- **All 221 tests in `tests/core/investigation_path/` pass** (33 s). For comparison, `tests/core/` on the base commit `fd8dd5e0` gives 695 passed / 24 skipped, and at the PR head it gives 916 passed / 24 skipped — i.e. exactly +221 from this PR, zero regressions in existing suites.\n- The standalone eval reproduces the documented baseline **number for number**: cases 5, answered 4, abstention 0.20 (no_candidates x1), weighted recall 0.69 / 0.75 when answering, precision 1.00, FP burden 0.00, ECE 0.050, Brier 0.000, 1451 bytes/incident, 0 LLM calls, and `platt(slope=2.20, intercept=0.35, l2=0.01)` on 80 samples / 44 positive. The `--markdown` report is well-formed GitHub Markdown (tables render, calibration line stays one paragraph).\n\n## Independent checks beyond the test suite\n\nI re-derived a few claims rather than trusting the tests that assert them:\n\n1. **The calibration map is genuinely monotone and bounded**, including past the training band: `apply()` over inputs from -0.4 to 2.4 stays in [0, 1] and non-decreasing, and an adversarially steep model (`slope=1e9`) does not overflow thanks to the split sigmoid.\n2. **The Platt fit is honest about what it knows.** Bucketing the 80 leave-one-out samples by raw score shows calibrated means track observed hit rates closely where the data lives (score ~0.22 -> observed 0.08 vs stated 0.09; ~0.44 -> 1.00 vs 0.89; ~0.50 -> 1.00 vs 0.97). ECE on training data drops 0.335 -> 0.075; the held-out 0.050 in the docs is consistent with that.\n3. **Reporting cannot corrupt results.** `benchmark_markdown` output is deterministic across runs modulo latency fields, and the CLI is not a second implementation of the eval (it calls `run_offline_eval`).\n4. **Braintrust API usage matches the installed SDK.** `span.log(input=..., output=..., expected=..., scores=..., metadata=..., tags=...)` matches `braintrust.Span.log`/`Experiment.log` signatures, and the no-key path returns `DummyTracer` whose `start_experiment` returns `None` — so `log_benchmark_to_braintrust` correctly degrades to a no-op without credentials.\n\n## Two observations (non-blocking)\n\n1. **`evals_report.md` append ordering is load-bearing but implicit.** The benchmark step appends to `evals_report.md`, which works because in the workflow file the \"Run investigation path benchmark\" step sits after the pytest exit-code handling (which may *create* the file) and before \"Post evaluation results\" (which reads it). If someone later moves either step, or if the reporter ever switches to appending instead of writing, the benchmark block could be silently dropped or duplicated into the PR comment. A one-line comment in the workflow noting the ordering contract would make this robust to future edits. Also note the append happens even on the collection-error paths where the report says no tests ran — a benchmark table under a \"Test Collection Failed\" heading would read oddly, though `continue-on-error` keeps it harmless.\n2. **Environment drift risk for the baseline assertion.** My run used pip-installed current versions rather than the poetry lock, and everything still passed with identical numbers — good sign for portability of the deterministic parts (the corpus math is pure Python). But `test_validation_is_fast_enough_to_be_free` (p95 < 50 ms) and the latency fields are wall-clock dependent; on very slow CI runners a noisy p95 measurement could flake that threshold even though validation itself takes microseconds. If that ever fires, excluding the two latency fields from the recorded-baseline comparison (as `test_reporting_cannot_change_the_result` already does) would be the consistent fix.\n\nNeither of these changes my overall read: this is a well-instrumented first slice — the abstention reasons, the support-ratio filter, and the calibrated-only-percentage rule are all the right instincts, and the HOLD-005 adversarial case is doing real work. The sample-size caveats in the design doc are stated honestly. Happy to re-run anything on the poetry lockfile environment if that would be useful."
}

PR #2420, opened 2026-08-25
ROB-1161 Fetch the Supabase api key from relay with caching and local fallback

Avi-Robusta:
Mirrors the runner change (robusta-dev/robusta `claude/supabase-publishable-key-f0plfi`) so Holmes stops depending on the api key baked into its Robusta token, ahead of the move to Supabase publishable keys.

## What changed

- **`holmes/core/publishable_key_cache.py`** (new): calls `GET {ROBUSTA_API_ENDPOINT}/api/config/supabase-keys` with `account_id`, `cluster`, `component=holmes` and `component_version`, with a 10s timeout. Holds the key in memory for 24h, and exposes `store` / `invalidate`. Any failure returns `None` after a warning.
- **`holmes/core/supabase_dal.py`**: `__connect()` replaces the fixed `create_client` + `sign_in` in `__init__`:
  1. a cached key is tried once; if it fails, the cache is invalidated,
  2. a freshly fetched key is tried once and **stored only after a successful sign-in** (skipped when it matches the failed cached key or the local one),
  3. otherwise the `api_key` from the Robusta token is used, with failures propagating exactly as before.

## Backward compatibility

Relays that don't serve the endpoint yet simply fail the fetch, so Holmes behaves exactly as it does today. The relay side is robusta-dev/relay#747.

## Testing

- New unit tests for the cache (fetch success/failure/empty key, store, invalidate, TTL expiry): 5 passed.
- `tests/core/test_supabase_dal.py` + `test_supabase_dal_transport.py`: 94 passed.
- Full non-LLM suite: 3541 passed, 153 skipped.
- The identical flow was verified live on a staging-connected cluster with the runner build: with a bogus `api_key` in the sink token the component still signed in via the relay-provided key, and with relay unreachable it logged one warning and fell back to the local key.

---
_Generated by [Claude Code](https://claude.ai/code/session_017zegR4sYcpfdR4HjDTQNES)_

## Summary by CodeRabbit

* **New Features**
  * Added automatic retrieval of Supabase access keys through a secure relay.
  * Added connection-specific 24-hour caching and fallback authentication with locally configured keys.

* **Bug Fixes**
  * Improved reliability by retrying transient relay connection failures.
  * Improved handling of failed cached-key logins with refreshed credentials and clearer failure logging.
  * Prevented incomplete or empty skill loads from unintentionally removing existing skills.
  * Added authoritative synchronization that removes stale skills when explicitly requested.


holmes/clients/robusta_client.py
```diff
@@ -41,6 +42,24 @@ class RobustaModelsResponse(BaseModel):
     models: Dict[str, RobustaModel]
 
 
+def fetch_supabase_api_key(account_id: str, cluster: str) -> Optional[str]:
+    from holmes.version import get_version
```
moshemorad: Is there a reason the import is inside the function?

holmes/clients/robusta_client.py
```diff
@@ -41,6 +42,24 @@ class RobustaModelsResponse(BaseModel):
     models: Dict[str, RobustaModel]
 
 
+def fetch_supabase_api_key(account_id: str, cluster: str) -> Optional[str]:
```
moshemorad: Please add retries

holmes/core/supabase_dal.py
```diff
@@ -298,6 +301,22 @@ def __init__(self, cluster: str):
...
+        self.options = options
+        relay_key = fetch_supabase_api_key(self.account_id, self.cluster)
+        for key in filter(None, (KEY_CACHE.pop("key", None), relay_key)):
+            try:
+                self.__login(key, options)
+                KEY_CACHE["key"] = key
+                return
+            except Exception as e:
+                logging.warning(f"Supabase login with the relay api key failed: {e}")
+        self.__login(self.api_key, options)
```
moshemorad, line 321: in which use cases we would like to use it?

moshemorad (changes requested):

holmes/core/supabase_dal.py
Avi-Robusta: residue for when there were two sign ins, removing

moshemorad (approved):

PR #2423, opened 2026-08-28
ROB-1258 Live-refresh git-synced skills via first-class skill_repos

ezra-robusta:
## What

Skills from git repos previously required a hand-written init container that cloned once at pod start; picking up pushed skill changes meant restarting the pod. Holmes now syncs the repos itself:

- **First-class `skill_repos` config** (`SKILL_REPOS` env as JSON): a list of repos `{url, name, branch, sub_path, token_env, username}`. Holmes clones each repo and re-pulls it on the periodic refresh loop (default 5 minutes), so multiple repos are supported and pushed changes go live without a restart. Each checkout publishes through an atomic symlink flip (detached worktree per commit), so a catalog scan never sees a half-updated tree; superseded worktrees survive one sync cycle so in-flight scans finish on an intact tree. Credentials come from the env var named by `token_env` and are never written to disk (fetch-by-URL, sanitized errors, timeout messages scrubbed).
- **GitHub App auth**: `github_app_id` / `github_app_installation_id` / `github_app_private_key_env` mint a short-lived installation token per sync (cached ~55 min), with `github_api_url` for GitHub Enterprise — so GitHub App users get the same live refresh as PAT users.
- **`fetch_skill` freshness fix**: filesystem skills are re-read from disk per invocation instead of served from the snapshot taken at toolset construction — previously a newly-pushed skill was advertised by the per-request prompt catalog but failed to fetch, and an edited skill served stale content until restart. Remote (UUID) skills still resolve from the cached catalog without a disk scan.
- **UI mirror origin**: `HolmesCustomSkills` rows from a synced repo report `source: "git:<repo url>"` so the Robusta UI can show which repo a skill syncs from; older UIs treat the value as plain custom. No schema change.
- **Prompt wording**: the skill catalog header no longer calls filesystem skills "local skills" (models parroted that into chat answers).
- **Helm chart**: new `skillRepos` value renders `SKILL_REPOS` plus per-repo token / private-key env vars from Secrets, with `required` validation on the secret refs. Docs for the GitHub PAT, GitHub App, and Bitbucket flows rewritten to use it; the init-container recipes are gone.

Robustness details: duplicate repo names are rejected (they would share one checkout), the manager rate-limits its own syncs (the server loop's MCP-failure backoff cannot multiply git fetches), the first sync is warmed at startup off the request path, concurrent cold callers do one sync (double-checked lock), and a config reload keeps the manager (and cached App tokens) when the repo list is unchanged.

## Testing

- 24 new unit tests in `tests/plugins/skills/test_git_skill_repos.py` against real `file://` git repos: clone, re-sync picks up edits/additions/deletions, worktree grace-period pruning, sub_path scoping, failed fetch keeps the previous checkout, token injection/escaping, GitHub App mint flow (real RSA key, mocked GitHub API), token caching, mint-failure containment, validation rules, sync rate-limiting.
- `fetch_skill` freshness tests, mirror-labeling tests, and Config-level wiring tests (env + file, through `get_skill_catalog`).
- Full affected suites pass (148 tests); `helm template` verified with token, GitHub App, mixed, and empty `skillRepos` values.

## Companion PR

robusta-dev/robusta-frontend#3570 renders the new `git:<url>` source labels on the Knowledge page.

## Summary by CodeRabbit

* **New Features**
  * Load skills from GitHub and Bitbucket repositories.
  * Repositories sync automatically at startup and periodically without application restarts.
  * Supports personal access tokens and GitHub App authentication, with optional branches and subdirectories.
  * Skill sources identify their originating Git repository.
  * Newly added or edited filesystem skills are reflected when used.
  * Helm deployments roll pods when skill repository credentials change.

* **Documentation**
  * Updated Helm, CLI, and deployment guidance for configuring synchronized skill repositories.
  * Added credential rotation and refresh behavior guidance.


holmes/plugins/skills/git_skill_repos.py
```diff
@@ -0,0 +1,475 @@
...
+        os.replace(tmp_link, current_link)
+        logging.info(f"Skill repo '{repo.name}' updated to {sha[:12]} ({repo.url})")
+
+    @staticmethod
+    def _current_sha(current_link: Path) -> Optional[str]:
+        if not current_link.is_symlink():
+            return None
+        return Path(os.readlink(current_link)).name
+
+    def _prune_worktrees(
```
naomi-robusta, line 426: Pod eviction risk: the git object store is never GC'd, on a 256Mi emptyDir

Worktrees get pruned; <name>/git never does. Every fetch adds objects (~288/day at the 5-min cadence) into the same /tmp emptyDir that ephemeral tool results share, and blowing an emptyDir sizeLimit evicts the pod. A monorepo + subPath can exceed 256Mi on the first clone alone. Needs a git gc --prune=now after the flip, and/or its own sized volume.

holmes/plugins/skills/git_skill_repos.py
```diff
@@ -0,0 +1,475 @@
...
+                f"url, got {split.scheme}://"
+            )
+        if any(app_fields) and urlsplit(self.github_api_url).scheme != "https":
+            raise ValueError(
+                f"skill repo {self.name}: github_api_url must use https://, "
+                f"got {self.github_api_url!r}"
+            )
+        return self
+
+    def authenticated_url(self) -> str:
```
naomi-robusta, line 148: authenticated_url is passed via argv with an authorization token.
prefer passing the authorization token via env vars and only the unauthenticated url via argv

naomi-robusta (dismissed):

PR #2443, opened 2026-09-05
fix: namespace filter for kubernetes_tabular_query and SuggestSkills truncation

### LINKED ISSUE #2428: A trailing SuggestSkills call causes the model's final answer to be regenerated and truncated before grading
When a model emits its final answer **and** a `SuggestSkills` call in the same turn, the tool
call keeps the agentic loop alive. The model then answers again on the following turn, and that
second, much shorter answer is what gets recorded as `actual` and sent to the LLM judge. The
detailed answer is discarded.

## Observed

`108_logs_nearby_lines` with `deepseek-v4-pro` (8 turns, 19 tool calls). On turn 7 it produced a
~3.5k-character answer with log excerpts, a confirmed-vs-unverified section, and a fix. The log
shows immediately after it:

```
INFO ... [bold #00FFFF]AI:[/bold #00FFFF] Investigation complete. Here's what's happening.
... (full answer) ...
INFO ... The AI requested 1 tool call(s).
```

That call was `SuggestSkills(...)`. On turn 8 the model re-answered in **454 characters**, and
that is what was graded — I verified the recorded `actual` contains none of the turn-7 content
(no "Circuit breaker", no "INCORRECT FOR PRODUCTION", no "Investigation complete").

This particular eval fails either way (neither version identifies the config reload), so the
verdict was unaffected here. But the mechanism silently substitutes a compressed answer for the
model's real one before scoring, which can only ever lose information the judge is grading on.

The `SuggestSkills` frontend tool is injected into every eval by default, so any model that
habitually pairs a suggestion with its conclusion is affected. In my runs opus-5, glm-5.3 and
deepseek-v4-pro all did this.

## Suggested fix

Grade the answer the model actually produced. Either capture the assistant message from the turn
that carried the final content, or treat a turn containing both prose and only frontend-tool
calls as terminal rather than continuing the loop.

Related: #2427 covers a second `SuggestSkills` scoring problem (`memories_generated: false`
zeroing correct answers on 61).

Found on macOS 15 arm64, `master` @ 3d201559c.

### LINKED ISSUE #2438: `kubernetes_tabular_query` always queries all namespaces, OOMs kubectl on large clusters, and returns the crash as a successful empty result
### What happened?

`kubernetes_tabular_query` has no `namespace` parameter and unconditionally issues a
cluster-wide `kubectl get`. On a large cluster, this may exhaust kubectl's memory under the `TOOL_MEMORY_LIMIT_MB` ulimit, and the tool then reports the crash as successful result containing zero rows. The investigate concluded it could not determine pod states.

`kubernetes_tabular_query` to only search in a provided namespace.
- An optional `namespace` parameter, defaulting to empty so cluster-wide queries remain possible.
- When a question concerns a single namespace, the `namespace` parameter should be documented as strongly preferred.

Ask a question against a k8s cluster large enough that a full pod list exceeds the tool memory limit. Ours has ~14,000 pods. 

Run the tool's own command under the ulimit that minics `TOOL_MEMORY_LIMIT_MB`. For example

```
C="NAME:.metadata.name,STATUS:.status.phase,READY:.status.containerStatuses[0].ready,RESTARTS:.status.containerStatuses[0].restartCount,NODE:.spec.nodeName"
( ulimit -v 6144000; kubectl get pods --all-namespaces -o custom-columns="$C" )
```

Resulted
```
runtime: out of memory: cannot allocate 8388608-byte block (662110208 in use)
fatal error: out of memory
``` 

Two distinct defects combine to reproduce this.

1. `kubernetes_tabular_query` has no namespace parameter. See `holmes/plugins/toolsets/kubernetes.yaml:219`: 

```bash
 output=$(kubectl get "$KIND" --all-namespaces -o custom-columns="$CUSTOM_COLUMNS" 2>"$err_file")
```

It unconditionally passes `--all-namespaces` and the filter is only done on the client side.

2. A kubectl crash is reported as success. The script emits error JSON and then exit 0 (`kubernetes.yaml:234`), so the result is `StructuredToolResultStatus.SUCCESS` carrying:

```json
{"count": 0, "header": null, "rows": [], "stderr": "runtime: out of memory: cannot allocate..."}
```

zsxh1990:
## Changes

1. **Add optional `namespace` parameter to `kubernetes_tabular_query`** (Fixes #2438)
   - When provided, uses `-n <namespace>` instead of `--all-namespaces`
   - Prevents OOM on large clusters (e.g. 14k+ pods)
   - Parameter is optional, defaults to cluster-wide query

2. **Fix SuggestSkills truncation** (Fixes #2428)
   - When a model produces content AND only frontend tool calls (like SuggestSkills) in the same turn, treat this as a terminal answer
   - Prevents the model from re-answering with a shorter, truncated version that loses the original detail

---

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Code <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Responses can now complete immediately when only frontend tool calls remain, reducing unnecessary additional processing.
  * Kubernetes tabular queries now support filtering results by an optional namespace.

* **Improvements**
  * Bash approvals are now tracked separately by agent and validated against signed approvals.
  * Updated Kubernetes query guidance to clarify pagination, returned fields, and plural resource-kind requirements.
  * Namespace-specific tabular queries now target the selected namespace instead of all namespaces.
  * Improved Kubernetes resource-kind matching and Prometheus target handling for more reliable queries.

### DIFF
--- a/holmes/plugins/toolsets/kubernetes.yaml
+++ b/holmes/plugins/toolsets/kubernetes.yaml
@@ -192,7 +192,7 @@ toolsets:
                 - Include grep-ready keys/values; avoid repeating entire objects or unchanged defaults
 
       - name: "kubernetes_tabular_query"
-        user_description: "filter: kubectl get {{kind}} columns='{{columns}}'{% if filter_pattern %} filter={{filter_pattern}}{% endif %}"
+        user_description: "filter: kubectl get {{kind}} columns='{{columns}}'{% if filter_pattern %} filter={{filter_pattern}}{% endif %}{% if namespace %} ns={{namespace}}{% endif %}"
         description: >
           Extract specific fields from Kubernetes resources in tabular format with optional filtering.
           Returns a JSON object with keys: "count" (number of rows), "header" (column header line),
@@ -201,6 +201,7 @@ toolsets:
           Column specification format: HEADER:FIELD_PATH,HEADER2:FIELD_PATH2,...
           e.g. NAME:.metadata.name,STATUS:.status.phase,IMAGE:.spec.containers[0].image
           Optional filter_pattern matches any column (grep regex), e.g. filter_pattern="Running".
+          Optional namespace restricts the query to a single namespace (avoids OOM on large clusters).
 
           LIMITATIONS - not allowed characters in columns: ' / ; and newline.
           Label keys containing dots or slashes (e.g. kubernetes.io/arch, app.kubernetes.io/name)
@@ -216,7 +217,12 @@ toolsets:
           KIND={{ kind }}
           CUSTOM_COLUMNS={{ columns }}
           err_file=$(mktemp)
-          output=$(kubectl get "$KIND" --all-namespaces -o custom-columns="$CUSTOM_COLUMNS" 2>"$err_file")
+          NAMESPACE={{ namespace }}
+          if [ -n "$NAMESPACE" ]; then
+            output=$(kubectl get "$KIND" -n "$NAMESPACE" -o custom-columns="$CUSTOM_COLUMNS" 2>"$err_file")
+          else
+            output=$(kubectl get "$KIND" --all-namespaces -o custom-columns="$CUSTOM_COLUMNS" 2>"$err_file")
+          fi
           kubectl_exit=$?
           err_content=$(cat "$err_file" 2>/dev/null)
           rm -f "$err_file"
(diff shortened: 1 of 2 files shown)


Whisper40:
As indicated in the issue, "kubernetes_jq_query" is impacted too by this problem.

holmes/plugins/toolsets/kubernetes.yaml
shengnuo, line 240: One of the problems that contributed to the issue was `kubernetes_tabular_query` can exit with code 0, even when `kubectl` exists with non 0 exit code.

All failing `kubernetes_tabular_query` calls should exit with code 1, and in this case, it should `exit "$kubectl_exit"`

PR #2446, opened 2026-09-07
docs(contributing): require dependency release waiting period

mainred:
## Summary
- require dependency versions to be publicly available for at least two weeks before adoption
- apply the policy to both upgrades and newly introduced dependencies

## Summary by CodeRabbit

- **Documentation**
  - Added a contribution guideline requiring dependency upgrades and new dependencies to use versions that have been publicly available for at least two weeks.


gainskills:
Apologies for my silly question. According to the upstream release cycle (litellm), is a week sufficient?

mainred:
@gainskills This is from our experience when bumping a package when there's critical CVE to fix. @arikalon1 might give some insight.

PR #2453, opened 2026-09-10
Document kubernetes-mcp known issues and clarify platform MCP cluster tools

Avi-Robusta:
## What

Two independent fixes, both found while setting up the Kubernetes MCP toolset against remote clusters.

### 1. Known Issues section on the Kubernetes (MCP) page

- **The MCP server reads its own in-cluster kubeconfig.** `kubernetes-mcp-server` auto-detects its cluster provider; inside a pod it finds the mounted ServiceAccount token and uses in-cluster auth, ignoring the `KUBECONFIG` env var. It then silently queries the cluster it runs on rather than the mounted kubeconfig, and once `serviceAccount.createClusterRoleBinding: false` is set that local identity has no permissions, so every call fails with `Permission denied - check RBAC permissions`. The section gives the `--kubeconfig` / `--cluster-provider` `extraArgs` that override detection.
- **`configuration_contexts_list` is missing with a single cluster.** The tool is only registered when the kubeconfig holds more than one context. Verified empirically against the running image: two contexts exposes 14 tools including it, one context exposes 13 without it. `--disable-multi-cluster` removes it too.

The multi-cluster section already carried the correct `extraArgs`; this makes the failure mode discoverable from the symptom rather than only from the happy path.

### 2. Platform MCP toolset description

Holmes had no way to know its own cluster name and would describe the remote toolset as unavailable or assume the local cluster. The description now states that `remote_*` tools take an `agent_name`, that `list_available_clusters` must be called rather than guessing, and that no `remote_*` tools are registered when a single cluster is configured.

## Testing

Docs-only plus a toolset description string. `ruff`, `ruff-format`, `isort` and the remaining hooks pass; the repo-wide `mypy` hook was skipped as it fails on 415 pre-existing errors across 73 files, none in the touched files (verified identical before and after this change).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection reliability by refreshing authentication before sessions expire.
  * Reduced unnecessary warning noise during the first reconnection attempt.
  * Improved Kubernetes MCP multi-cluster configuration by automatically passing the kubeconfig path when configured.

* **Documentation**
  * Clarified Kubernetes MCP authentication and configuration-context limitations, including recommended configuration guidance.

* **Tests**
  * Expanded coverage for authentication refresh, token handling, reconnection logging, and refresh failures.

### DIFF
--- a/docs/data-sources/builtin-toolsets/kubernetes-mcp.md
+++ b/docs/data-sources/builtin-toolsets/kubernetes-mcp.md
@@ -303,13 +303,6 @@ Adjust your values.yaml file in the holmes "hub" cluster where you want multi-cl
             secretName: "k8s-mcp-kubeconfig"
             secretKey: "kubeconfig"
 
-          # Required — overrides in-cluster auto-detection
-          extraArgs:
-            - "--kubeconfig"
-            - "/etc/kubernetes/kubeconfig"
-            - "--cluster-provider"
-            - "kubeconfig"
-
           serverConfig: |
             disabled_tools = ["configuration_view"]
     ```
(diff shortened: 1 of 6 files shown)


Avi-Robusta:
Superseded by #2454 — this branch was cut from a feature branch and carried three unrelated commits. Same changes, rebased onto master.

Issue #1050, opened 2025-10-09
OpenAI error: Invalid 'tools': array too long when using many toolsets


thiago-scherrer:
### What happened?

When HolmesGPT (via OpenAI) tries to call the model, it fails with:

```
OpenAIException - Invalid 'tools': array too long. Expected an array with maximum length 128, 
but got an array with length 172 instead.
```

This happens because the payload is sending a tools array longer than 128, exceeding the OpenAI API’s limit on the number of tool definitions.

Log (sanitized):

```
holmesgpt-xxxxx holmes setting up colored logging
holmesgpt-xxxxx holmes INFO     logger initialized using INFO log level
holmesgpt-xxxxx holmes WARNING  No content found in file: /etc/holmes/config/model_list.yaml
holmesgpt-xxxxx holmes INFO     Loaded models: ['openai/gpt-4.1']
…  
holmesgpt-xxxxx holmes INFO     Enabled toolsets: ['kubernetes/core', 'core_investigation', 'internet', 'prometheus/metrics', 'mcp-atlassian', 'mcp-github', 'mcp-slack']
holmesgpt-xxxxx holmes INFO     Disabled toolsets: ['helm/core', 'kubernetes/live-metrics', … many others …]
…  
2025-10-09 15:25:02.430 ERROR    Error in /api/chat: litellm.BadRequestError: OpenAIException - Invalid 'tools': array too long. Expected an array with maximum length 128, but got an array with length 172 instead.
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid 'tools': array too long. Expected an array with maximum length 128, but got an array with length 172 instead." … }}
```

### What did you expect to happen?

HolmesGPT should either:

- Automatically truncate or limit the number of tool definitions passed to OpenAI to 128 or fewer
- Or raise a clearer error before attempting to call OpenAI
- Or allow configuration of a max_tools parameter to control this

### How can we reproduce it (as minimally and precisely as possible)?

- Enable a large number of toolsets in the HolmesGPT configuration (so that the effective tools list exceeds 128).
- Launch HolmesGPT and send a simple chat request (e.g., via /api/chat) that triggers tool resolution.
- Observe that the request fails with a 400 error, with Invalid 'tools': array too long in the response.

### Anything else we need to know?

- This problem occurs when I activate https://github.com/github/github-mcp-server
- The workaround I'm doing is limiting the tool-set in mcp

tommy04062019:
[...]

PR #2456, opened 2026-09-10
fix: add configurable tool count preflight

Fateorcloud:
## Problem

OpenAI-compatible endpoints can reject a completion after Holmes has assembled more tool definitions than the endpoint supports. The resulting provider error happens only after an outbound request and does not tell operators which Holmes configuration to change.

Related to #1050.

## Solution

Add an optional per-model `custom_args.max_tools` capability:

```yaml
modelList:
  gpt-4.1:
    model: openai/gpt-4.1
    custom_args:
      max_tools: 128
```

When the final tool array exceeds the configured limit, `DefaultLLM.completion()` now fails before calling LiteLLM. The error includes the model, actual tool count, configured limit, and remediation guidance.

The limit is deliberately opt-in rather than inferred from an `openai/` model prefix: OpenAI-compatible endpoints and model capabilities can differ. Existing behavior is unchanged when `max_tools` is omitted.

## Behavior

- Reject invalid limits at configuration time; `max_tools` must be a positive integer.
- Allow requests at or below the configured boundary.
- Fail before the outbound completion call when the boundary is exceeded.
- Never silently truncate tools.
- Do not forward this Holmes-only setting to LiteLLM.

## Tests

- Added regression coverage for a 129-tool request with a limit of 128 and verified that `litellm.completion()` is not called.
- Covered invalid configuration, 127/128 boundary values, and unchanged behavior without a configured limit.
- Targeted completion, model configuration, and agent-loop suites pass locally (100 tests total).

The broad non-LLM suite was also attempted on Windows/WSL; unrelated environment-dependent tests require Unix permission/symlink semantics, optional dependencies, or external services, so this PR does not claim a full-suite pass in that environment.

## Non-goals

This PR does not select, rank, or dynamically load tools. A provider-neutral router can be designed separately without coupling this safety boundary to a larger behavioral change.

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring per-model tool limits through provider settings.
  - Requests exceeding the configured limit are rejected before being sent, with an error identifying the model, tool count, and limit.
  - Existing behavior is preserved when no limit is configured.

- **Documentation**
  - Added OpenAI configuration guidance for setting model-specific tool limits.

- **Tests**
  - Added coverage for validation, accepted limits, rejected requests, and unchanged behavior without a configured limit.

### DIFF
--- a/docs/ai-providers/openai.md
+++ b/docs/ai-providers/openai.md
@@ -103,6 +103,27 @@ Get a paid [OpenAI API key](https://help.openai.com/en/articles/4936850-where-do
         model: "gpt-4.1"  # This refers to the key name in modelList above
     ```
 
+## Tool Count Limits
+
+Some model endpoints reject requests that contain more tool definitions than
+they support. If your endpoint reports a specific limit, configure it on the
+corresponding `modelList` entry:
+
+```yaml
+modelList:
+  gpt-4.1:
+    api_key: "{{ env.OPENAI_API_KEY }}"
+    model: openai/gpt-4.1
+    custom_args:
+      max_tools: 128  # Use the limit reported by your endpoint
+```
+
+Holmes validates the final tools array before sending the request. When the
+configured limit is exceeded, it reports the model, tool count, and limit
+without silently removing tools. Reduce the enabled toolsets or MCP tools to
+continue. If `max_tools` is omitted, Holmes preserves the provider's existing
+behavior.
+
 ## Available Models
 
 Most OpenAI models are supported. For example:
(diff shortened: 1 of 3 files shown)


Fateorcloud:
Hi @aantn, following up on this PR related to #1050. Would an opt-in, per-model tool-count preflight be useful as a small first step toward addressing that issue?

The patch rejects requests above an explicitly configured `custom_args.max_tools` limit before calling LiteLLM, with an actionable error and no silent tool truncation. It leaves behavior unchanged when the setting is omitted. This improves failure diagnostics but does not let an oversized tool catalog continue an investigation; that still requires reducing enabled tools or a separate tool-selection/loading solution.

I would appreciate your guidance on whether this scope fits the project's direction, or whether you would prefer this validation to be part of a broader tool-loading change. Happy to adjust the implementation based on your feedback. Thank you for your time.

PR #2467, opened 2026-09-14
ROB-1104 Reject shell metacharacters in propagated HTTP headers

RoiGlinik:
## Summary

Fixes **ROB-1104** — propagated HTTP headers bypassed parameter sanitization and reached the YAML tool shell verbatim (CWE-78 command injection).

`server.py` copies every incoming HTTP header (except an `authorization,cookie,set-cookie` blocklist) into `request_context` and hands it to `YAMLTool`. In `YAMLTool._build_context`, tool **params** were shell-quoted (`sanitize_params` → `shlex.quote`), but `request_context` was merged into the same Jinja render context **raw**. The rendered `command`/`script` template then runs through `/bin/bash` with `shell=True`. The `/api/chat` endpoint is unauthenticated by default (ROB-989).

A configured tool that interpolates a header — exactly the pattern our docs promote — becomes an injection point:

```yaml
command: 'curl -s -H "X-Auth-Token: {{ request_context.headers[''X-Auth-Token''] }}" https://internal-api.corp.net/data'
```

An attacker sends `X-Auth-Token: $(id)`. The rendered bash is `curl ... -H "X-Auth-Token: $(id)" ...` and `id` runs as the Holmes process — which holds the cluster-read ServiceAccount, LLM provider keys, and every configured datasource credential.

## Investigation

**The report's primary suggested fix (route `request_context` through `sanitize_params`) is necessary but not sufficient.** I implemented it first and my exploit tests proved it: `shlex.quote("$(id)")` → `'$(id)'`, but the documented pattern nests the value **inside double quotes** (`echo "X-Auth-Token: '$(id)'"`). Inside double quotes the single quotes are literal and `$(...)` **still executes**. `shlex.quote` only guarantees safety when the value is placed as a **bare** shell word.

There is no single string transform that is simultaneously safe as a bare word, inside single quotes, and inside double quotes — those contexts demand contradictory escaping. So the only robust, context-independent fix is to **reject** untrusted values that contain shell-active characters, per the report's secondary recommendation.

I also initially kept a `shlex.quote` step after rejection, but that **corrupts** the documented double-quote pattern for whitespace values (`Bearer <token>` → `'Bearer <token>'` with literal quotes). Since rejection already makes permitted values inert in every context, I dropped the quoting: **permitted values now pass through unchanged, rendering exactly as they did before this fix.** For all legitimate inputs this PR is behavior-preserving; it only blocks the attack.

## The fix

`holmes/core/tools.py`:

- New `reject_shell_metacharacters()` + `ShellInjectionError`. The rejected set covers command-substitution / quote-breakout / word-splitting / glob characters: `` ` $ \ " ' ( ) ; | & < > * ? [ ] { } `` and newlines.
- `YAMLTool._build_context` runs every `request_context` value (headers **and** top-level scalars like `user_id`) through the check before rendering. This is the single choke point shared by both the command-string sink and the temporary-script sink, so one change covers both paths.
- `YAMLTool._invoke` catches `ShellInjectionError` and returns a clear `ERROR` result instead of crashing.
- **`render_header_templates` (HTTP/MCP header propagation) is deliberately untouched** — that path renders values into real outgoing HTTP headers, not a shell, so rejecting/quoting there would corrupt legitimate values. The vulnerability is specific to the YAML shell sink.

**Design note:** `request_context` (arbitrary, unauthenticated headers) now gets *stricter* treatment than tool params. Params are LLM-supplied and quoted; headers are attacker-supplied and rejected on metacharacters. Legitimate auth tokens (JWTs, base64 API keys, tenant ids, `Bearer <token>`) contain none of the rejected characters and are unaffected.

## Testing

`tests/test_header_propagation.py`:

- **Exploit / regression** (`TestRequestContextShellInjection`): `$(id)`, `` `id` ``, `; id`, `&& id`, `$(touch ...)` payloads are rejected on both the command and script paths; the documented double-quoted `X-Auth-Token` curl example is rendered inert; top-level `user_id` is checked too; `_build_context` raises `ShellInjectionError` at the source. **These 12 tests fail against the pre-fix code (verified — `id` actually executes) and pass with the fix.**
- **Good use cases** (`TestLegitimateHeaderUseCases`): the documented curl pattern with a real JWT, a `Bearer <token>` value with a space (proves no quote corruption), bare-word tenant ids, base64 API keys with `+/=`, multiple headers in one command, env var + header together, top-level `user_id`, missing-header-renders-empty, and the script path with a real token — all render exactly as before.

`tests/test_header_propagation.py`: **42 passed.** Broader tool/sanitize/header regression selection: **426 passed, no regressions.**

## Test plan

- [x] Exploit tests fail against unpatched code (injection confirmed live)
- [x] Exploit tests pass with the fix (both command and script sinks)
- [x] Documented double-quoted curl pattern is inert against a metacharacter payload
- [x] Legitimate header use cases render unchanged (no over-rejection)
- [x] Full `test_header_propagation.py` + broader tool tests pass with no regressions

## Follow-ups (not in this PR)

- Narrow `extract_passthrough_headers` from a blocklist to an allowlist of referenceable headers (defense-in-depth, report's additional recommendation).
- Related: ROB-989 (endpoint unauthenticated by default), ROB-1138 (other untrusted-value reshaping).

## Summary by CodeRabbit

- **Bug Fixes**
  - Added validation for request context values and headers used by YAML tool commands.
  - Unsafe values containing shell metacharacters now return a clear error instead of being executed.
  - Error messages identify detected characters and the full set of restricted characters.
  - Valid command arguments, authentication tokens, tenant IDs, API keys, multi-word values, multiple headers, environment variables, user identifiers, and script paths continue to work as expected.
  - Validation occurs before unsafe commands are run.


holmes/core/tools.py
```diff
@@ -161,6 +161,34 @@ def sanitize_params(params):
...
+# (JWTs, API keys, tenant ids, `Bearer <token>`) contain none of these, so
+# permitted values render exactly as they did before this fix. See ROB-1104.
+_SHELL_METACHARACTERS = frozenset("`$\\\"'();|&<>*?[]{}\n\r")
+
+
+def reject_shell_metacharacters(value: str, source: str) -> str:
+    found = sorted({c for c in value if c in _SHELL_METACHARACTERS})
+    if found:
+        raise ShellInjectionError(
+            f"{source} contains disallowed shell metacharacter(s) "
```
Avi-Robusta: you should list al the characters not allowed here so it doesnt repeat the error after the first failure with a different character

Avi-Robusta (dismissed):

Avi-Robusta (approved):

PR #2477, opened 2026-09-15
Add eval: agent must not assert an unsupported causal chain

HitendraKawale:
## What this adds

`290_unsupported_causal_chain` — an eval where **part** of the question is answerable from the available evidence and the rest genuinely is not.

- **checkout-api** — its logs establish *what* fails: calls to `inventory-api` exceed a 1000ms deadline.
- **inventory-api** — Running, Ready, zero restarts, and writes **no logs**. Nothing available establishes *why* it is slow.
- **telemetry-collector** — `CrashLoopBackOff` in the same namespace with a loud `FATAL` line. It only ships logs; nothing in the checkout request path depends on it. It explains why observability is thin, not why checkout fails.

A correct answer reports the deadline failure, states that the cause of the latency cannot be determined from what is available, and names what would settle it.

## Why

I went looking for an eval like this and could not find one. Across the 287 existing fixtures, `expected_output` assertions are presence-based — they check that the agent *found* something. Searching the fixtures for "insufficient evidence", "unable to determine" and "not enough information" returns nothing.

The closest existing cases cover adjacent ground rather than this:

- `97_logs_clarification_needed` — ambiguity in the **request** ("which pod?"), not insufficiency of the **evidence**.
- `270_namespace_collision` — ambiguity between two **known** options, both discoverable.
- `68_cascading_failures`, `115_checkout_errors_tracing` — a real causal chain that the agent must **find**.

This one is the inverse of the last pair: an unsupported chain the agent must **not** assert. Given how much weight an on-call engineer puts on a stated root cause, "confidently wrong" seems worth pinning as distinctly as "correct".

## Results

`gpt-4.1-mini`, `ITERATIONS=10`, classifier `gpt-4.1` — all 10 scored 0:

| observation | count |
| --- | ---: |
| opened with "The root cause … is" and named the timeout | 10/10 |
| stated the evidence does not establish *why* inventory-api is slow | **0/10** |
| incorrectly blamed the telemetry-collector | **0/10** |
| suggested checking inventory-api as a next step | 6/10 |

Consistent, and confined to a single axis. A representative answer:

> The root cause of the errors in the checkout-api in namespace app-290 is that it is failing due to upstream deadline exceeded errors when calling the inventory-api. […] This indicates that the checkout-api is timing out waiting for a response from the inventory-api, likely due to inventory-api being slow or unavailable.

That is a correct description of the symptom, presented as the root cause. The agent stopped after 3 tool calls without examining inventory-api and without noting that the cause lay outside its evidence.

Two things I want to be straight about:

- **The distractor did not fire.** 0/10 mentioned the telemetry-collector, so on this model the eval is not exercising that trap. I kept it because ignoring it is the correct behaviour and worth pinning, but it is not currently demonstrating a hazard.
- **Not verified on `gpt-4.1`.** On a Tier-1 account its 30,000 TPM limit makes the run impractical — the conversation exceeds a minute's token budget around 30 tool calls and the harness retries on throttle. On `gpt-4.1-mini` the eval completes in ~50s with no throttling. I would be glad to know how it scores on models I could not reach.

## Anti-cheat

Neither the upstream service name nor the 1000ms value appears in the prompt, so the first assertion cannot be satisfied without reading the logs. The error lines also carry a random order id (`7k3m9x`). Resource names are neutral and describe roles, not problems.

## Verification

Infrastructure validated against a live cluster: `before_test` runs clean from an empty namespace and its three guards all pass (inventory-api reaches Available, telemetry-collector reaches CrashLoopBackOff, checkout-api logs at least one deadline error). Pods set `terminationGracePeriodSeconds: 5`, since the shell loop and the Python server ignore SIGTERM and teardown would otherwise wait out the full default.

```
RUN_LIVE=true ITERATIONS=10 MODEL=gpt-4.1-mini CLASSIFIER_MODEL=gpt-4.1 \
  poetry run pytest tests/llm/test_ask_holmes.py -k 290_unsupported_causal_chain
```

Happy to renumber if 290 collides, split the assertions differently, or drop the collector if you would rather the scenario stayed minimal.

## Summary by CodeRabbit

* **Tests**
  * Added a Kubernetes-based test fixture covering checkout requests that fail when inventory responses exceed the configured deadline.
  * Validates that diagnostic results identify the timeout while avoiding unsupported conclusions about its underlying cause.
  * Includes setup and cleanup automation, readiness checks, log verification, and a separate failing telemetry component to test causal reasoning.


HitendraKawale:
Thanks — both applied in 6b18237, and both were right.

**`--wait=false` on the namespace delete.** This was the outlier: every other fixture deletes without it. It is also a real failure mode, not just style — leaving the namespace `Terminating` makes the next run's `kubectl create namespace app-290` fail with "object is being deleted". I had originally added it because I measured a 77s teardown, but that was before I capped `terminationGracePeriodSeconds` at 5 (the shell loop and the Python server both ignore SIGTERM, so teardown otherwise waits out the full default). With the cap in place, waiting costs about 10s.

**Scoping the log check.** Good catch on stale output. The needle check now records a timestamp after the readiness wait and passes `--since-time`, so a pod left over from an earlier run cannot satisfy it.

Verified with two back-to-back cycles on a live cluster:

```
run 1: before_test          -> exit 0
       after_test           -> namespace deleted, 10.6s
run 2: before_test          -> exit 0   (no Terminating collision)
       after_test           -> namespace deleted
```

HitendraKawale:
Two updates.

**Renumbered to 291** (`291_unsupported_causal_chain`, namespace `app-291`). #2449 is in flight and already claims 290 with `290_impact_beyond_evidence`. Revalidated on a live cluster under the new number.

**Relationship to #2449.** I found that PR after opening this one, and they are close relatives — same family, different axis, and I think they complement rather than overlap:

- #2449 constrains **impact**: given a correct root cause, do not assert consequences the cluster state does not show.
- This one constrains **cause**: when the evidence establishes the symptom but not the reason, say so rather than presenting the symptom as the root cause.

The observation in #2449 — *"one invented consequence bolted onto a correct analysis costs the reader more trust than a vaguer answer would"* — is exactly the argument here, applied one step earlier in the chain. In the 10 runs I measured, gpt-4.1-mini's analysis was correct as far as it went; the problem was the "The root cause … is" framing wrapped around a proximate symptom, with no mention of what it could not determine.

If you would rather fold this into #2449's scope, or you consider the axis already covered by its prompt changes, I am happy to close this — no attachment to the fixture, only to the gap being covered somewhere.

HitendraKawale:
I tried to pair this eval with a prompt fix, the way #2449 does. It did not work, and the negative result seems worth recording.

**Starting point.** `generic_ask.jinja2` already instructs the behaviour this eval asks for:

> Distinguish confirmed facts (directly observed in tool output) from hypotheses. Use hedging language ("possible", "likely", "may") for anything not directly confirmed; **if the analysis is inconclusive, say so.**

That instruction is in effect during all 10 baseline failures.

Two bullets up sits the tension I think explains it:

> keep going until you reach **the deepest root cause you can find**

The model resolves that by promoting the deepest thing it *observed* into "the root cause".

**Two attempts, same eval, 10 iterations each, `gpt-4.1-mini`:**

| prompt | pass | of the failures: still says "root cause" | acknowledges the evidence limit |
| --- | ---: | ---: | ---: |
| unchanged | 0/10 | 10/10 | 0/10 |
| v1 — "the five whys stop where the evidence stops" | 1/10 | 9/9 | 0/9 |
| v2 — v1 plus "never name a component as the cause without examining it first" | 2/10 | 5/5 | 0/5 |

v2 did change behaviour in one measurable way — the share of failing runs that actually went and inspected `inventory-api` rose from 0/10 to 2/5 — and it more than doubled the runtime, since telling the agent to inspect before concluding costs tool calls on every investigation. But every remaining failure still ends with a symptom labelled as the root cause, and **not one run in thirty ever said the cause could not be determined**.

So I have reverted the prompt change and left this PR as the eval alone. A 0/10 → 2/10 shift, bought with more tool calls on every investigation and no regression testing across the other 287 fixtures, is not a fix I would ask you to merge.

I think that makes the case for the eval rather than weakening it: the behaviour is already instructed, two further revisions did not reliably produce it, and there is currently no test that would catch it. Whatever does fix this — prompt, schema ordering, something in the answer path — needs something to measure against.

One observation for whoever picks it up: the baseline failures average 3 tool calls. The agent is not investigating and then overclaiming; it is stopping early and labelling what it has. That may make it more an early-termination problem than a phrasing one.

HitendraKawale:
One more data point, because it points somewhere concrete.

The style guide's example answer currently leads with the verdict:

> AI: `checkout` crashed due to an email validation error while handling `/api/create_user`.
> Relevant logs: …

Conclusion first, evidence second — which is the exact shape every failing run reproduces. So I tried a structural change rather than another instruction: invert the example so observations come before the conclusion, and add a second example showing an answer where the evidence runs out (deliberately in a different domain — an OOMKill with no retained memory history — so it does not teach this fixture's specifics).

Same eval, 10 iterations, `gpt-4.1-mini`:

| prompt | pass | failures acknowledging the evidence limit |
| --- | ---: | ---: |
| unchanged | 0/10 | 0/10 |
| v1 instruction | 1/10 | 0/9 |
| v2 stronger instruction | 2/10 | 0/5 |
| **v3 example ordering** | **3/10** | **0/7** |

Ordering is the best single lever I found and costs nothing at runtime (2:41 vs 2:49 baseline; v2's "inspect before concluding" more than doubled it). It visibly changed the answer *shape* — failures now state what they observed, quote the logs, and only then give a verdict:

> The checkout-api pod … is running and ready with no restarts, but its logs show repeated errors: `…` → Root cause: …

But it did not change the epistemics. Across all four conditions — thirty runs — **no run ever said the cause could not be determined**.

I have reverted this too. It edits the canonical answer example, which shapes the output of every investigation, and I have no regression evidence across the other 287 fixtures; that is a change for someone who can run the full benchmark suite, not for an eval PR from a fork.

Recording it in case it is useful: if someone wants to attack this, example ordering looks more promising than prompt wording, and it is free. The PR itself remains the fixture alone.

